PR #5, opened 2018-07-05
Get fingerprint from key

chocol4te:
Sorry, I didn't do a clean merge before doing this PR, but it is only for displaying the fingerprint code: really only the contents of `src/packet/types/key.rs` needs to be copied and pasted into a new commit.

### DIFF
--- a/src/armor.rs
+++ b/src/armor.rs
@@ -1,14 +1,14 @@
-use nom::{self, digit, line_ending, not_line_ending};
-use nom::types::CompleteStr;
-use crc24;
 use base64;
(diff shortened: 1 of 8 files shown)


dignifiedquire:
Thank you! A lot has changed, but it should be much easier to implement this now. I have added placeholder methods here: https://github.com/dignifiedquire/pgp/commit/731edd783c3e2fbce0767f0f690fe4276a712a32 where you can fill in the details. The key now stores expiration time and creation time, and the type structure is much simpler and hopefully easier to navigate.

PR #7, opened 2018-07-06
Add fingerprint() and key_id() method

chocol4te:
I have added test outlines, but am unsure of how you would like the tests to be structured. These functions should apply to both a whole parsed key, but also apply to subkeys.

I tested only with DSA and RSA on V3 and V4, and *should* be correct for other algorithms.

### DIFF
--- a/src/composed/key.rs
+++ b/src/composed/key.rs
@@ -569,8 +569,6 @@ mod tests {
     #[test]
     fn test_fingerprint_rsa() {
         // V4
-        w
-
 
         // V3
     }
--- a/src/composed/key.rs
+++ b/src/composed/key.rs
@@ -42,6 +42,11 @@ impl PublicKey {
     pub fn fingerprint(&self) -> Vec<u8> {
         self.primary_key.fingerprint()
     }
+
+    /// Returns the Key ID of the associated primary key.
+    pub fn key_id(&self) -> Vec<u8> {
+        self.primary_key.key_id()
+    }
 }
 
 /// Represents a Public PGP SubKey.
@@ -90,6 +95,11 @@ impl PrivateKey {
     /// Returns the fingerprint of the associated primary key.
     pub fn fingerprint(&self) -> Vec<u8> {
         self.primary_key.fingerprint()
+    }
+
+    /// Returns the Key ID of the associated primary key.
+    pub fn key_id(&self) -> Vec<u8> {
+        self.primary_key.key_id()
     }
 }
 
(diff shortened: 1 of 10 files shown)


src/packet/types/key.rs
```diff
@@ -257,11 +261,137 @@ macro_rules! key {
...
+                        }
+
+                        h.update(&packet).unwrap();
+
+                        h.finish().unwrap().deref().to_vec()
+                    }
+                }
+            }
+
+            pub fn key_id(&self) -> Vec<u8> {
```
dignifiedquire: given that this is not always supported I think returning `Option<Vec>` would be better

src/composed/key.rs
```diff
@@ -555,4 +565,39 @@ mod tests {
 
         assert_eq!(ua.signatures, vec![sig3]);
     }
+
+    #[test]
+    fn test_fingerprint_rsa() {
+        // V4
```
dignifiedquire: The testcases in `test/opengpg-interop/testcases/keys` have json files that contain the right fingerprints.  I plan to parse and test them automatically, but for now you can just hardcode these from there. A single key for each is good enough for now

src/packet/types/key.rs
chocol4te: What would result in returning None? I can't find much on the specification :/

src/composed/key.rs
chocol4te: Sure thing, writing them now :)

src/packet/types/key.rs
dignifiedquire: if it’s a v3 key with non rsa algorithm, though, instead of an empty vec

src/composed/key.rs
```diff
@@ -301,7 +314,10 @@ mod tests {
         let input = ::std::str::from_utf8(raw).unwrap();
         let key = PublicKey::from_string(input).expect("failed to parse key");
 
-        // assert_eq!(key.primary_key.fingerprint(), "56c65c513a0d1b9cff532d784c073ae0c8445c0c");
+        assert_eq!(
+            hex::encode(key.primary_key.fingerprint()),
+            "56c65c513a0d1b9cff532d784c073ae0c8445c0c"
```
dignifiedquire: there was already hex_literal used in another test, can you replace that with this as well, no need to have to crated for hex decoding and this one looks nicer

chocol4te: Will do :)

chocol4te:
I think I just have one last bug to fix for ECDH keys and this should be good to go.

PR #11, opened 2018-07-08
CI changes

chocol4te:
- Switch to Nightly
- Added cargo clippy
- Added rustfmt check - only passes if rustfmt was ran in commit

Switching to Nightly shouldn't be an issue as a nightly build passing means a stable build will pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ cache:
 
 install:
   - rustup component add rustfmt-preview
-  - cargo install clippy
+  - cargo install clippy --force
 
 script:
   - cargo fmt --all -- --check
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,12 +1,21 @@
 sudo: false
+
 language: rust
-rust: stable
```
dignifiedquire: lets keep stable, to make sure that we don’t start using nightly only features.

checkout https://dev.to/cad97/great-rust-ci-1fk6 for a nice setup description

dignifiedquire:
Update ci quite a lot since this was made, so closing for now

PR #12, opened 2018-11-08
Implement basic message parsing and decryption using RSA

dignifiedquire:
- no openssl, pure rust 
- basic message parsing works
- lots of things still missing


hpk42:
nice :)

PR #23, opened 2019-02-28
Some Makefile voodoo

flub:
This attempts to make the usual `make; make install` work.  In particular
it means people can invoke make install as sudo without having to worry
needing rust/cargo installed for the root user.

It needs to have a stab at knowing what the input files are sadly, and
also needs to create a single output; hence the, somewhat traditional,
build-stamp file.

It also uses some $(CURDIR) fun to so that `make -C pgp-ffi` will work
fine as well.

Lastly it fixes the librpgp.h typo and -lrpgp in the pkg-config file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ Cargo.lock
 .profile
 ios_simulator_app
 *.h
+pgp-ffi/build-stamp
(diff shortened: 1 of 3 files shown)


r10s:
yeah, great :)

PR #36, opened 2019-03-31
Make rpgp.pc relocatable

flub:
This uses the built-in ${pcfiledir} variable to make the paths
relative to where the rpgp.pc file is.  It also uses the convention of
${exec_prefix} for libdir, which is perhaps a bit frivolous.

The motivation for this is that there are now binary packages
available for download which include rpgp.pc but do not have any
install step and thus do not modify the prefix.


hpk42:
seems this is abandoned, if not please re-open ;)

PR #45, opened 2019-09-05
Add failing RSA 4096 secret key test

link2xt:


link2xt:
I am not sure why keys are in the `autocrypt/` dir, should it be aadded somewhere else?

dignifiedquire:
> I am not sure why keys are in the autocrypt/ dir, should it be aadded somewhere else?

yes the autocrypt folder is just cloned from the actual autocrypt test cases, please add in the `tests` folder.

Can you give me some information about this key, how was it generated, does gnupg verify & validate it?

link2xt:
I have generated the key for my email with Enigmail first. I selected 4096 RSA key. I have compiled latest desktop DC and tried to import the key via Autocrypt setup message, but it failed.

Then I figured out that the problem is in rpgp, cloned it, exported my secret key with Enigmail and added it as a testcase. It failed.

To reproduce the case without publishing my secret key, I have started "gpg --full-generate-key", selected RSA, 4096, password "1" and no expiration date (my own key has expiration date), and added it as a testcase. I have then exported the key with `gpg -a --export-secret-keys Test`.

Here is the key info:
```
$ gpg --import-options show-only --import  < tests/autocrypt/test@example.com.sec.asc
sec   rsa4096 2019-09-05 [SC]
      5EEC0096120A360E657899A771A0F7F78E02BD56
uid                      Test <test@example.com>
ssb   rsa4096 2019-09-05 [E]
```

`cargo test rsa4096` fails with ```
thread 'key_autocrypt_rsa4096_sec' panicked at 'failed to unlock key: Incomplete(Size(6017))', src/libcore/result.rs:999:5
```

So far I have traced it down to this line:
https://github.com/rpgp/rpgp/blob/d2034a092edae7237c043019b38f73ff68fa4d53/src/types/params/plain_secret.rs#L41

dignifiedquire:
@link2xt that error is expected, you are trying to decode a private key that has a password, you will need to pass it (the autocrypt keys do not have password protection)

See this testcase for an example: https://github.com/rpgp/rpgp/blob/master/tests/key_test.rs#L504

link2xt:
Replaced with an issue #46

PR #47, opened 2019-09-14
Handle incomplete packet body properly

link2xt:
Incomplete packet body should be handled the same as incomplete packet header.

If packet is incomplete and no more data is available, that is an error.

### DIFF
--- a/src/packet/many.rs
+++ b/src/packet/many.rs
@@ -109,6 +109,7 @@ impl<R: Read> Iterator for PacketParser<R> {
 
             if let Some((length, p)) = res_body {
                 info!("got packet: {:#?} {}", p, length);
+                assert!(length > 0);
                 b.consume(length);
                 return Some(p);
             }
--- a/src/packet/many.rs
+++ b/src/packet/many.rs
@@ -70,11 +70,9 @@ impl<R: Read> Iterator for PacketParser<R> {
                 second_round = true;
             }
 
-            let res = match {
-                match single::parser(b.buf()) {
-                    Ok(v) => Ok(v),
-                    Err(err) => Err(err.into()),
-                }
+            let res_header = match single::parser(b.buf()) {
+                Ok(v) => Ok(v),
+                Err(err) => Err(err.into()),
             }
             .and_then(|(rest, (ver, tag, _packet_length, body))| match body {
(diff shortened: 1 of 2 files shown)


link2xt:
The cause of the problem is that when nom returns Incomplete error in
single::body_parser, it was packed into 
Error::InvalidPacketContent(Box::new(...)) and never looked at.

It should be handled the same way as Incomplete error during the packet header parser. Parser should attempt to read more data, instead of trying to skip over the packet. Trying to skip over the packet does not work because its length is 0.

Marked as WIP because I have not looked into cases `ParseResult::Fixed` and `ParseResult::Partial` yet, they should be handled similarly probably.

@dignifiedquire

dignifiedquire:
this looks like a good approach, please make sure to add a testcase for this

link2xt:
@dignifiedquire there is already a testcase commit, it panics without fixes due to assertion I added or gets stuck in the infinite loop if you run it with the existing codebase.

link2xt:
Once CI passes I'll do some refactoring and mark it as non-WIP

link2xt:
@dignifiedquire Can you take a look at `parse_dumps!` change in the last commit?

I just modified it to the values that I got in assertion failures, and the last commit that modifies them has no comment on how they were obtained: https://github.com/rpgp/rpgp/commit/0f0046d8f88eac686b9498f2b820fdc3eb83957a#diff-29f8d647e4640bb61795c93756a9a91aL118-R127

As I understand, these numbers indicate how many packets are successfully parsed and shouldn't decrease?

link2xt:
The problem is that in `ParseResult::Fixed(body)` case body parser returns `Incomplete` error sometimes. We probably want to pass it through indeed, because we know the length to skip.

link2xt:
@dignifiedquire PR is ready for review.

dignifiedquire:
@link2xt looks good to me, can you take a look and fix the ci failures please?

link2xt:
> looks good to me, can you take a look and fix the ci failures please?

`composed::message::types::tests::test_x25519_encryption` test is flaky for some reason, I can't reproduce it even by cleaning and re-running the tests many times on my machine.

dignifiedquire:
Thanks, merged it manually

PR #55, opened 2019-10-11
fix security issue where a remote attack can send an invalid-length c…

hpk42:
…yphertext and crash decryption


src/crypto/sym.rs
```diff
@@ -214,6 +214,7 @@ impl SymmetricKeyAlgorithm {
     ) -> Result<(&'a [u8], &'a [u8])> {
         let bs = self.block_size();
 
+        ensure!(bs + 2 < ciphertext.len(), "invalid ciphertext");
```
dignifiedquire: this should be the other  way around, bs+2>=ciphertext.len()

PR #60, opened 2019-11-01
fix(line-reader): improve correctness of LineReader::seek

link2xt:
Currently `LineBreak::seek` processes only `io::SeekFrom::Current(n)` and immediately converts `n` to its absolute value.

I test that forward and backward seek works as expected (currently it doesn't, hence draft PR).

### DIFF
--- a/src/base64_reader.rs
+++ b/src/base64_reader.rs
@@ -35,7 +35,7 @@ impl<R: Read + Seek> Read for Base64Reader<R> {
         for i in 0..n {
             if !is_base64_token(into[i]) {
                 // the party is over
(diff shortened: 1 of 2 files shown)


link2xt:
@dignifiedquire Could you check if the test I wrote is correct?

I made `LineReader` tests pass, but base64 and depending functions are broken now. Somehow the rest of the line (`-----hello`) is lost in some buffer.

link2xt:
Fixed it, there was an additional bug in Base64Reader::read. It moved the cursor forward, not backward, when it encountered non-base64 character.

PR #63, opened 2019-11-04
CI: Add cargo audit job

link2xt:


link2xt:
I have tested this CI configuration on my own fork.

link2xt:
We may want to tweak this to also run daily/weekly instead of on each commit/PR, but this change is an improvement anyway.

PR #65, opened 2019-11-05
Use strum::IntoEnumIterator to reduce boilerplate

link2xt:


dignifiedquire:
not convinced the added dep is really worth it atm

PR #73, opened 2019-11-13
refactor!: remove email module

link2xt:
It is not complete and is not used in Delta Chat


dignifiedquire:
> is not used in Delta Chat

That is not the only reason things exist in rpgp. But I agree that it should be removed from this repo

PR #78, opened 2019-11-23
refactor: Upgrade to nom 5

link2xt:

PR #79, opened 2019-11-27
nom 5 with custom errors

link2xt:
This PR is built on top of #78, which already kinda works.

nom 5 removed `ErrorKind::Custom` and requires that user of the library implements their own error type instead. Since `named_args!` macro [does not support custom errors yet](https://github.com/Geal/nom/issues/455) and `named!` macro does not offer much, I am replacing calls to both with ordinary functions, so I can specify error type explicitly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,7 @@ rand = "0.6"
 smallvec = "0.6.9"
 cast5 = "0.6.0"
 rsa = "^0.1.4"
-nom = "^4.2"
+nom = "5"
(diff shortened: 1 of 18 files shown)


link2xt:
This is currently stuck beause of upstream bug https://github.com/Geal/nom/issues/1078
Apparently nobody used nom 5 with generic errors to parse bits yet.

dignifiedquire:
closing for now, feel free to reopen if you make any progress

PR #83, opened 2019-12-20
Fix typo in feature name

Sergey-Kitov:
ringbuf feature name has typo, so "default-features=false" doesn't have any effect, when rpgp is used


dignifiedquire:
Merged and fixed manually into #86

PR #84, opened 2019-12-21
wrap armored data correctly

r10s:
RFC5322 requires emails to be wrapped using CRLF (\r\n) and not only LF (\n)

fixes https://github.com/deltachat/deltachat-core-rust/issues/1093 and similar issues.


dignifiedquire:
This is not quite correct, it should likely be an option, as this is used for writing to files primarily

dignifiedquire:
Clearer alternative in https://github.com/deltachat/lettre/pull/1

r10s:
@dignifiedquire thanks a lot!

PR #88, opened 2020-02-29
Fix ECDH serialization for encrypted session keys with leading zeros

link2xt:
<del>Ported test by @r10s from https://github.com/deltachat/deltachat-core-rust/pull/1316</del>

Description of bug and bugfix is in commit messages. Do not squash.


link2xt:
Existing `test_x25519_encryption` also fails when run 1000 times.

dignifiedquire:
@link2xt can you fix ci please?

link2xt:
> @link2xt can you fix ci please?

Fixed clippy, using `write_u8` now.

PR #92, opened 2020-04-02
fix: handle x25519 keys that are shorter than 32 bytes on deserialization

dignifiedquire:


r10s:
great, thanks a lot for fixing!

PR #101, opened 2020-06-13
fix: fail to verify unsigned messages

link2xt:
Encrypted but unsigned messages should fail to verify, otherwise simply
dropping invalid signature makes verify() check pass.


dignifiedquire:
I am not certain this is the right way to do it. An unsigned message is very much valid, it just isn't signed. This change would mean that if you have a combination of signed and unsigned messages it would failt to verify, even if all signed messages verify properly.

link2xt:
> This change would mean that if you have a combination of signed and unsigned messages it would fail to verify, even if all signed messages verify properly.

`Message` type can only store one message. If at some higher level there can be multiple messages (e.g. MIME parts), then user may need to look if each part is signed and decide whether it is acceptable or not that there is no signature at all.

In this case there is a need for some sort of `is_signed()` method. So user can check `!msg.is_signed() || msg.verify().is_ok()`. But returning `Ok()` from `verify` is error-prone, user will keep using it incorrectly.

link2xt:
@dignifiedquire rPGP itself does not seem to have any structures that contain `Message`s and verify them recursively? So adding `Message.is_signed() -> bool` should be sufficient?

link2xt:
I have worked around it in https://github.com/deltachat/deltachat-core-rust/pull/1616 by decompressing the message first and then matching on `Signed` variant, but still think there is a need for a method that checks if the message is signed with a valid key. And can't see how `verify` in its current state is useful.

rk0n:
> > This change would mean that if you have a combination of signed and unsigned messages it would fail to verify, even if all signed messages verify properly.
> 
> `Message` type can only store one message. If at some higher level there can be multiple messages (e.g. MIME parts), then user may need to look if each part is signed and decide whether it is acceptable or not that there is no signature at all.
> 
> In this case there is a need for some sort of `is_signed()` method. So user can check `!msg.is_signed() || msg.verify().is_ok()`. But returning `Ok()` from `verify` is error-prone, user will keep using it incorrectly.

I would love to see this merge request in the project. The Result of this function was really missleading for me. 

It could also help to change the comment on the function and include that for other encrypted message type it will only return everything is ok:
```
/// Verify this message.
/// For signed messages this verifies the signature and for compressed messages
/// they are decompressed and checked for signatures to verify.
```

Another possibility would be to introduce error messages for the other message types. Then you know that something is going on and you don't get the incorrect result that everything is ok although nothing has been verified.

link2xt:
@dignifiedquire How about deprecating this method and maybe replacing it with some fine-grained functions instead? I don't see any use case for it currently. In deltachat-core-rust we don't use it because of this problem.

dignifiedquire:
@link2xt I believe this is fixed on `master` with the new unsupported error, can you verify and close if so?

dignifiedquire (approved):

PR #106, opened 2020-08-23
feat: make sign/verify utilize trait std::io::Read

### LINKED ISSUE #99: Signing based on `Read` trait instead of slices
It would be advantageous to provide a way to sign multiple existing slices.
The easiest way would be to introduce API consuming a `R: Read` type which the user can than implement as desired. Otherwise, especially if parsing chunked external data, it would be necessary to copy the data.

drahnr:
Closes #99

### DIFF
--- a/src/composed/message/types.rs
+++ b/src/composed/message/types.rs
@@ -356,7 +356,12 @@ impl Message {
                     hashed_subpackets,
                     unhashed_subpackets,
                 );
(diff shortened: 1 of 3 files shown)


drahnr:
@dignifiedquire could you review this? I think the length specification can be just implicitly defined by the reader.

dignifiedquire (approved):

dignifiedquire:
thanks

PR #108, opened 2020-10-17
Make some ciphers optional

link2xt:


hpk42:
any particular reason for this change?

link2xt:
I want to remove support for DES and Cast5 and blowfish from Delta Chat. Maybe also twofish. These packages add unnecessary dependencies and 3DES is insecure.

wiktor-k:
One thing if you don't mind... 

Why not leave enum unchanged even if the algo is not supported and only return "unsupported" errors when trying to use the algo that has been unselected?

This way the errors could be more descriptive (like unsupported algorithm cast5) instead of errors during parsing (unsupported algorithm constant 5).

dignifiedquire:
We might want to do this at some point, but not right now, so closing for now.

PR #117, opened 2021-01-22
Make signature field of StandaloneSignature public

rvolgers:
This is consistent with many other structs defined in rpgp, and it's
hard to do basic things such inspecting details of the signature
otherwise.


dignifiedquire (approved):
makes sense, thank you

PR #118, opened 2021-01-22
Fix corruption of non-ascii characters in text literals

rvolgers:
First commit adds a test revealing the bug.

Second commit fixes the problem by making `normalize_lines..rs` work on bytes instead of chars.

This simplifies the code and it should be faster as well.


dignifiedquire (approved):
nice catch, looks good, thanks

PR #119, opened 2021-01-22
Simplify code by making Hasher implement std::io::Write

rvolgers:
This makes it possible to skip intermediate buffers and simplifies the code.

The upstream `Digest` in rustcrypto also supports this, so there is precedent for doing it this way.

No functional change.


dignifiedquire (approved):
thanks

PR #123, opened 2021-02-03
chore: update dependencies

link2xt:
stream-cipher and block-cipher were moved into the cipher crate, so all depending crates had to be updated.


dignifiedquire (approved):
thanks

PR #124, opened 2021-02-08
Add DSA verify implementation

rvolgers:
This handles no secret data and I'm reasonably confident in the
implementation, but I'm not sure what the standard is for accepting new
cryptographic implementations.

This is actually already covered by the test suite, the DSA signatures
were just being ignored previously.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,7 +52,7 @@ thiserror = "1.0.9"
 try_from = "^0.3"
 twofish = "^0.4"
 zeroize = { version = "1.1.0", features = ["zeroize_derive"] }
-hmac-drbg = "0.3.0"
+hmac = "0.10.1"
 
 [dependencies.buf_redux]
 version = "0.8.1"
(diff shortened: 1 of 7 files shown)


rvolgers:
Sigh. Please hold. I may have spotted a concern.

rvolgers:
Okay, pretty sure that was a legitimate bug, and that it's fixed now. But this means this should probably have more tests.

rvolgers:
Okay this has grown way beyond what I had planned, it's now a complete DSA implementation except for key generation.

Probably this would be better in an external crate. But I will leave this open in case there is interest.

dignifiedquire:
A wild DSA implementation appeared 😄 
Thanks a lot, it will take me a little bit more time to review this, given it is an entirely new algorithm, but I definitely want to get it in. 

In the meantime, could you take a look at the CI failures please?

rvolgers:
See related discussion in https://github.com/RustCrypto/signatures/pull/223

Most of this willl probably *eventually* go into RustCrypto. I'm not quite sure when though. If you want to merge this as is and later replace it with the implementation there that'd be fine too. But maybe keep this open for a little longer as I expect to add key generation as well, making it a pretty complete implementation.

dignifiedquire:
closing in favor of #276

Issue #87, opened 2020-02-18
Can't verify cross-signed ids using Signature::verify_certificate


Nemo157:
`pgp::packet::Signature::verify_certificate` assumes that the issuing key is also the key that had the id signed to it. This isn't the case when you have signed someone else's identity, in that case you're signing their `(PublicKey, UserId)` pair using your key.

It might be a while till I can throw together a small testcase, but I encountered this while verifying my signatures from https://nemo157.com/gpg, and [adding an extra parameter to take both keys](https://github.com/Nemo157/rpgp/commit/96476589e2c250c2c04f2e406aa4184491f2574b) appears to verify at least the cross-signing `CertGeneric` signatures fine.

adrianheine:
I ran into this issue since I cross-signed my new key using my old (DSA) key. As you wrote, `Signature::verify_certificate` currently uses the subject key even if the issuer is a different key, and thus tries to verify a DSA signature using a completely different RSA key. I think an intermediate correct behavior would be to skip those signatures and warn (as it already does). In the long run, a) not being able to verify a signature because of a missing key should be an output value and b) it should be possible to supply multiple keys.

dignifiedquire:
@hko-s I failed to come up with a good solution for this in the past, any thoughts on how to handle this best?

hko-s:
> @hko-s I failed to come up with a good solution for this in the past, any thoughts on how to handle this best?

I'd imagine a function on `Signature` that takes two `&impl PublicKeyTrait` and one `UserId` would be the most reasonable?

Maybe roughly `Signature::verify_user_id_binding(signer, signee, user_id)` (not sure about the order of the two keys, maybe one order is more in line with the other `verify` functions?)

hko-s:
(Ah, but I guess that's only part of what's needed - I'm tempted to hand-wave, and say that the big-picture operation should be addressed in a higher layer :))

hko-s:
The above comment is regarding "Third-party signatures on User IDs" (signature types 0x10-0x13, see https://www.rfc-editor.org/rfc/rfc4880.html#section-5.2.1).

There are also third-party signatures directly on the primary key (signature type 0x1F). Frankly, I'm not sure which software interprets those, and if they are at all commonly made.

link2xt:
As of version 0.19.0 there is a `Signature.verify_third_party_certification`: https://docs.rs/pgp/0.19.0/pgp/packet/struct.Signature.html#method.verify_third_party_certification

[...]

PR #130, opened 2021-04-08
Skip validating cross-signatures

adrianheine:
This relates to #87. An alternative short-term approach would be to return an error and mask it for example in `SignedUser::verify`.


mbelop:
Please consider this change for inclusion. I've also encountered RSA keys containing signature objects from other RSA or EdDSA keys and the current behavior results in attempts to validate these signatures using incorrect key params. Returning `Ok(())` instead of the error is reasonable in this case since the key validation shouldn't fail because of this.

dignifiedquire:
should be fixed by #262

PR #142, opened 2021-09-07
Add a method to construct a PublicKey from parameters.

jeamland:
Currently the only way to create a PublicKey is to have a pre-existing packet or to generate a new SecretKey and get it from there. This change adds a method that allows PublicKey objects to be synthesised from data from other sources, for example external key material.

### DIFF
--- a/src/packet/public_key_macro.rs
+++ b/src/packet/public_key_macro.rs
@@ -12,6 +12,45 @@ macro_rules! impl_public_key {
         }
 
         impl $name {
+            /// Create a new `PublicKeyKey` packet from underlying parameters.
+            pub fn new(
+                packet_version: $crate::types::Version,
+                version: $crate::types::KeyVersion,
+                created_at: chrono::DateTime<chrono::Utc>,
+                expiration: Option<u16>,
+                public_params: $crate::types::PublicParams,
+            ) -> Self {
+                let algorithm = match public_params {
+                    $crate::types::PublicParams::RSA { .. } => {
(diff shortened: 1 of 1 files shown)


src/packet/public_key_macro.rs
```diff
@@ -44,6 +44,42 @@ macro_rules! impl_public_key {
...
+                        $crate::crypto::public_key::PublicKeyAlgorithm::Elgamal
+                    }
+                    $crate::types::PublicParams::EdDSA { .. } => {
+                        $crate::crypto::public_key::PublicKeyAlgorithm::EdDSA
+                    }
+                };
+
+                $name {
+                    packet_version: $crate::types::Version::Old,
+                    version: $crate::types::KeyVersion::V4,
```
dignifiedquire, line 75: Either the function should be named differently, or the `packet_version` and `version` should be passed in as actual parameters. Currently it is not clear what will be constructed.

jeamland:
Hi, and thanks for the feedback!

I've turned it into a more generic `::new` constructor that takes the packet and key version parameters. I'm more than happy to tweak things further if you feel it's needed.

dignifiedquire:
Thanks, I think at this point it would be nice to reuse this in the `from_sclice` method as well, as the `new` method should also do the basic verfication for RSA only on older KeyVersions.

jeamland:
Hi again, I've reworked ::new to incorporate pretty much all the guts from ::from_slice. Let me know what you think.

dignifiedquire (approved):
nice, thank you

PR #166, opened 2022-04-25
Avoid some allocations via to_string

sebastinas:
&String can be compared with &str and &str satisfies the trait bounds
for std::fmt::builders::DebugStruct::field.


dignifiedquire (approved):
thanks

PR #185, opened 2022-07-16
Add comma in `README` for clearer message

Rudxain:
I think it makes it less ambiguous

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ And then run tests with `RUST_LOG=pgp=info`.
 
 Some key differences:
 
-- rPGP has a more libre license than Sequoia, that allows a broader usage
+- rPGP has a more libre license than Sequoia, which allows a broader usage
 
 - rPGP is a library with a well-defined, relatively small feature-set
   where Sequoia also tries to be a replacement for the GPG command line tool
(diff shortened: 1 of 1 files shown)


r10s (commented):
lgtm - but that does not mean much in this case, my english is not good enough to decide on that comma :) cc @rosano .

just a general note: meanwhile also Sequoia has a more libre license, LGPL , https://sequoia-pgp.org/blog/2021/10/18/202110-sequoia-pgp-is-now-lgpl-2.0/ - ianal, however, i think MIT/Apache is still more permissive so that sentence is still correct.

Rudxain:
Interesting, thank you for the info

README.md
```diff
@@ -59,7 +59,7 @@ And then run tests with `RUST_LOG=pgp=info`.
 
 Some key differences:
 
-- rPGP has a more libre license than Sequoia that allows a broader usage
+- rPGP has a more libre license than Sequoia, that allows a broader usage
```
rosano, line 62: I think this is more gramatic

```suggestion
- rPGP has a more libre license than Sequoia, which allows a broader usage
```

r10s (approved):

dignifiedquire:
thanks

PR #187, opened 2022-08-19
Remove unused clear_on_drop dependency

sebastinas:
`clear_on_drop` seems to be completely replaced by `#[zeroize(drop)]`.


dignifiedquire:
unfortunately dalek hasn‘t been updated yet last time I checked, so we need this to force the correct features for the dependency there

sebastinas:
`x25519-dalek` switched to `zeroize` in version 0.6.0, and `ed25519-dalek` switched in version 1.0.0.

PR #189, opened 2022-08-30
Replace find_map with a find

sebastinas:
The argument is not transformed, so find is sufficient.

This change fixes a clippy error causing the CI to fail.


dignifiedquire (approved):
thanks

PR #190, opened 2022-08-30
Remove unused circular dependency and also rewrite IV handling to not require lazy_static

sebastinas:
The old code was taking a copy of IV. Instead, we can just use `GenericArray::from` to convert the `u8` array into a `GenericArray`. This approach has the same cost (for the copy) without the need for `lazy_static`.


dignifiedquire (approved):
thanks

PR #192, opened 2022-09-19
feat: allow subkeys to be created with the authentication flag enabled

jeffreybolle:
The method `can_authenticate(bool)` was missing from the `SubkeyParams` builder but was supported by everything further down.  I have tested this by generating a key that has an authenticate subkey and using this to successfully log into a Linux server using SSH.


dignifiedquire (approved):
thanks

PR #204, opened 2022-12-17
ECDSA support

lumag:
This PR provides support for ECDSA signatures. Only P256 and P384 are supported for now.


src/crypto/ecdsa.rs
```diff
@@ -0,0 +1,137 @@
...
+use signature::{
+    hazmat::{PrehashSigner, PrehashVerifier},
+    Signature as SigSignature,
+};
+
+use crate::crypto::{ECCCurve, HashAlgorithm};
+use crate::errors::Result;
+use crate::types::{ECDSASecretKey, Mpi, PlainSecretParams, PublicParams};
+
+/// Generate an EdDSA KeyPair.
```
dignifiedquire, line 12: typo eddsa

src/crypto/ecdsa.rs
```diff
@@ -0,0 +1,137 @@
...
+            let r = sig[0].as_bytes();
+            let s = sig[1].as_bytes();
+
+            ensure!(r.len() <= flen, "invalid R (len)");
+            ensure!(s.len() <= flen, "invalid S (len)");
+            ensure_eq!(p.len(), 2 * flen + 1, "invalid P (len)");
+            ensure_eq!(p[0], 0x04, "invalid P (prefix)");
+
+            let pk = p256::ecdsa::VerifyingKey::from_sec1_bytes(p)?;
+            let mut sig_bytes = vec![0u8; 2 * flen];
```
dignifiedquire, line 68: `flen` could be `const` and this an array

src/crypto/ecdsa.rs
```diff
@@ -0,0 +1,137 @@
...
+            let r = sig[0].as_bytes();
+            let s = sig[1].as_bytes();
+
+            ensure!(r.len() <= flen, "invalid R (len)");
+            ensure!(s.len() <= flen, "invalid S (len)");
+            ensure_eq!(p.len(), 2 * flen + 1, "invalid P (len)");
+            ensure_eq!(p[0], 0x04, "invalid P (prefix)");
+
+            let pk = p384::ecdsa::VerifyingKey::from_sec1_bytes(p)?;
+            let mut sig_bytes = vec![0u8; 2 * flen];
```
dignifiedquire, line 93: same as above

src/crypto/ecdsa.rs
```diff
@@ -0,0 +1,137 @@
...
+            let secret = p384::ecdsa::SigningKey::from_bytes(&d)?;
+            let signature = secret.sign_prehash(digest)?;
+            signature.to_vec()
+        }
+        _ => unsupported_err!("curve {:?} for ECDSA", curve),
+    };
+
+    let len = bytes.len() / 2;
+    let r = bytes[..len].to_vec();
+    let s = bytes[len..].to_vec();
```
dignifiedquire, line 134: instead of converting to a vec twice, as the signature length is fixed for each variant could do sth like 
```rust
let r = vec![0u8; len];
r.copy_from_slice(&signature[..len]);
```

src/crypto/ecdsa.rs
```diff
@@ -0,0 +1,137 @@
...
+    }
+}
+
+/// Sign using ECDSA
+pub fn sign(
+    curve: &ECCCurve,
+    secret_key: &ECDSASecretKey,
+    _hash: HashAlgorithm,
+    digest: &[u8],
+) -> Result<Vec<Vec<u8>>> {
```
dignifiedquire, line 115: could be a tuple `Result<(Vec<u8>, Vec<u8>)>` instead of the additional vec allocation

src/types/params/plain_secret.rs
```diff
@@ -182,9 +182,31 @@ impl<'a> PlainSecretParamsRef<'a> {
...
+                        Ok(SecretKeyRepr::ECDSA(ECDSASecretKey {
+                            oid: curve.oid(),
+                            x: d.into(),
+                        }))
+                    }
+                    ECCCurve::P384 => {
+                        ensure!(d.len() <= 48, "invalid secret");
+
+                        let mut secret = [0u8; 48];
+                        secret[48 - d.len()..].copy_from_slice(d.as_bytes());
```
dignifiedquire, line 199: padding should be done for P256 as well, no?

tests/key_test.rs
```diff
@@ -739,22 +739,22 @@ openpgp_key!(
 openpgp_key!(
     key_openpgp_samplekeys_ecc_sample_1_pub,
     "samplekeys/ecc-sample-1-pub.asc",
-    false
+    true
```
dignifiedquire, line 742: 🎉

dignifiedquire:
Thanks, some small nits and could you rebase onto current `master` please?

src/crypto/ecdsa.rs
lumag: I followed other PK systems here. `crate::crypto::rsa::sign()` and `crate::crypto::eddsa::sign()` return Vec of Vec here. See `src/packet/secret_key_macro.rs`.

dignifiedquire: fair, I should probably clean this up at some point 😅

src/crypto/ecdsa.rs
lumag: Not quite. `signature` isn't an array here, it is a thin wrapper around GenericArray,  but we don't have access to internals. I changed this to `signature.split_bytes()`, then each component is converted using `to_vec()`.

dignifiedquire:
sorry, I broke `master` with the last merge, should be fixed again

lumag:
@dignifiedquire 
Hmm. I tried enabling verification on `key_openpgp_samplekeys_e2e_p256_1_prt` and it fails

dignifiedquire:
> @dignifiedquire Hmm. I tried enabling verification on `key_openpgp_samplekeys_e2e_p256_1_prt` and it fails

Merged it for now, will take a look at those tests in a bit.

PR #140, opened 2021-08-31
Upgrade Nom

dignifiedquire:
- [x] ~~Nom 5~~
- [x] Nom 6
- [ ] Nom 7

PR #214, opened 2023-03-16
Update nom

Skgland:
Upgrade from nom 4 to nom 7.

Done from scratch rather than using #140 as the rebase was ugly.

# TODO
- [x] cleanup some functions
- [x] cleanup commits / commit messages

### DIFF
--- a/src/types/mpi.rs
+++ b/src/types/mpi.rs
@@ -1,11 +1,12 @@
 use std::{fmt, io};
 
 use byteorder::{BigEndian, WriteBytesExt};
-use nom::{self, be_u16, Err, InputIter, InputTake};
+use nom::number::streaming::be_u16;
+use nom::{self, Err, InputIter, InputTake};
 use num_bigint::BigUint;
 use zeroize::Zeroize;
 
-use crate::errors;
+use crate::errors::{self, Error, IResult};
 use crate::ser::Serialize;
 use crate::util::{bit_size, strip_leading_zeros, strip_leading_zeros_vec};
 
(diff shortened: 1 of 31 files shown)


dignifiedquire:
Very cool, looking forward to reviewing this. Please let me know if you need any help.

Skgland:
@dignifiedquire should now be ready for review

src/types/mpi.rs
```diff
@@ -28,23 +29,36 @@ const MAX_EXTERN_MPI_BITS: u32 = 16384;
...
         // same as take!
         let cnt = len_actual as usize;
         match number.slice_index(cnt) {
-            None => nom::need_more(number, nom::Needed::Size(cnt)),
-            Some(index) => {
+            Err(needed) => {
+                // if input.at_eof() {
+                //     Err(nom::Err::Failure(nom::error::Error::new(input, nom::error::ErrorKind::Eof)))
+                // } else {
+                // FIXME what about eof?
```
dignifiedquire, line 57: what about this?

src/types/mpi.rs
```diff
@@ -28,23 +29,36 @@ const MAX_EXTERN_MPI_BITS: u32 = 16384;
...
-            nom::ErrorKind::Custom(errors::MPI_TOO_LONG)
-        )))
+        /*
+            Err(Err::Error(
+                error_position!(
+                input,
+                nom::error::ErrorKind::Custom(errors::MPI_TOO_LONG)
+            )))
+            FIXME should this give more information in the error?
+            FIXME is this the right Error Variant?
```
dignifiedquire, line 46: what about these?

dignifiedquire:
Looking good, just a clippy issue & the `FIXME`s I believe

src/types/mpi.rs
Skgland: Right, I wanted to check what todo about that.
Turns out for slices at_eof was always false in nom 4.
So this is correct as is.
I have removed the FIXME in 44da300

src/types/mpi.rs
Skgland: I think originally this would end up as `pgp::errors::Error::ParsingError(nom::error::ErrorKind::Custom(MPI_TOO_LONG))` with MPI_TOO_LONG being just the constant `1000u32`.

`nom::error::ErrorKind` no longer has a `Custom` Variant, 
should I add new variant to `pgp::errors::Error`, keep it as is without the FIXME comment or something else?

Skgland:
I will fix rustfmt together with the fixme once I have an anser regarding that

src/types/mpi.rs
dignifiedquire: great

src/types/mpi.rs
dignifiedquire: Invalid input is good enough for now.

Skgland: Ok the I will remove the FIXME

dignifiedquire:
All manual tests also pass, including no regression in benchmarks.

dignifiedquire (approved):

Skgland:
- last of my FIXMEs removed
- rustfmt
- clippy
should be good now

dignifiedquire:
Thanks again!

PR #221, opened 2023-03-19
Move from String to BString

### LINKED ISSUE #195: Allows creating non utf-8 `String` without `unsafe`
https://github.com/rpgp/rpgp/blob/66c4f76ee7420c51d4f04e1f44400d4ec2d34ffb/src/util.rs#L189-L191

[Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5b55d6ec60d2b01cf276a700f9ae4a27)

This should probably use one of the [`from_utf8`](https://doc.rust-lang.org/stable/std/str/fn.from_utf8.html) family of [functions](https://doc.rust-lang.org/stable/std/?search=from_utf8) ?

```rust
pub fn read_string(raw: &[u8]) -> String {
    // this would (violently) prevent invalid utf8 to propagate to other parts of the code
    std::str::from_utf8(raw).unwrap().to_owned()
}
```

or

```rust
pub fn read_string(raw: &[u8]) -> String {
    // Marks this place as unsafe but behave the same as the original cast
    unsafe {
        std::str::from_utf8_unchecked(raw).to_owned()
    }
}
```

VincentWo:
Non ascii characters weren't correctly converted from `[u8]` to `str` to deal with non utf-8 data. Using the [bstr](https://docs.rs/bstr/latest/bstr/)  crate allows to still accept arbitrary bytes and also handles utf-8.

Closes #195


VincentWo:
Hi, this should already fix all the issues concerning unicode, I couldn't find the correct place to add tests, but the added functionality is also not really complicated.

src/packet/signature/ser.rs
```diff
@@ -370,4 +373,9 @@ mod tests {
     fn packet_signature_roundtrip_openpgp_sig_2_keys_2() {
         test_roundtrip("sig-2-keys-2.sig");
     }
+
+    #[test]
+    fn packet_signature_roundtrip_openpgp_with_unicode() {
+        test_roundtrip("unicode.sig");
```
dignifiedquire, line 379: can you add a comment as to what exactly is in that file 😅

dignifiedquire:
If you rebase on `master` CI should be happy again

VincentWo:
Added a comment and rebased

dignifiedquire (approved):

dignifiedquire:
thanks a lot!

PR #231, opened 2023-03-29
Do not treat all signatures with mismatching key ID as valid

link2xt:
This looks like a typo introduced in
https://github.com/rpgp/rpgp/pull/215

This leads to skip of validation for signatures with mismatching ID, but instead of returning an error, the signatures are treated as valid.

This PR was later squashed into commit
https://github.com/rpgp/rpgp/commit/a463cca1cb2bbefe4bb3148babee3685a06a14a6

### DIFF
--- a/src/packet/signature/types.rs
+++ b/src/packet/signature/types.rs
@@ -88,8 +88,8 @@ impl Signature {
                     &key.key_id(),
                     issuer
                 );
-                // We can't validate this against this key, as there is a missmatch.
-                return Ok(());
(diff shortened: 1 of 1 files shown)


dignifiedquire:
The tests currently rely on this "feature", so this requires a considerable change before we can merge this

dignifiedquire:
This will need a security advisory though..

dignifiedquire:
Sigh, you are entirely right, there is actually a bug hiding in the keyid/serialization code for ECDSA..

PR #234, opened 2023-04-17
Packet: make members of OnePassSignature public

Xynnn007:
We need to parse some specific members of an OnePassSignature, thus they need to be public


dignifiedquire (approved):
seems reasonable

dignifiedquire (approved):
seems reasonable

PR #249, opened 2023-08-08
Add secp256k1 curve to ECDSA

LiveDuo:
I recently had a use case that needed the `Secp256k1` curve. Thought why not give it a try and implement it for the crate.

~~I left a few `unwrap` operations in [ecdsa.rs#L129](https://github.com/LiveDuo/rpgp/blob/master/src/crypto/ecdsa.rs#L129) and [plain_secret.rs#L196](https://github.com/LiveDuo/rpgp/blob/master/src/types/params/plain_secret.rs#L196). At first I thought of using `SignatureError` from [errors.rs#L72](https://github.com/LiveDuo/rpgp/blob/master/src/errors.rs#L72) but then realised that wraps `Signature` error from `ed25519_dalek` which does not seem appropriate. So I guessed it's better to ask.~~


LiveDuo:
@dignifiedquire

LiveDuo:
@dignifiedquire Can we run the CI again?

LiveDuo:
@dignifiedquire The latest CI run had two issues.

1. An issue with `clippy::perf` in [src/crypto/ecdh.rs](https://github.com/LiveDuo/rpgp/commit/7bdd1580711db27a3030eddf100a16e9bfa3df65#diff-962d2eaeb82df8d56721a98d7d86e270a5021cfa2b361b25146ecbd68d47cf51L143).
2. An issue with `test_parse_dumps_3` and `test_parse_dumps_8` in [tests/key_test.rs].(https://github.com/LiveDuo/rpgp/commit/7bdd1580711db27a3030eddf100a16e9bfa3df65#diff-d17677f7f412539024c880640c9f03510a6e24dd46469a673c8fe65292f95512L123-L130). The test counts of `0003.pgp` are decreased (both `expected_count` and `total_count`) and the test count of `0008.pgp` is increased (only `expected_count` that does not make sense to me). 

Both issues seam unrelated to the PR but I could be missing things. Regardless, I made a [commit](https://github.com/LiveDuo/rpgp/commit/7bdd1580711db27a3030eddf100a16e9bfa3df65) that seem to fix them.

Let me know if you want me to update the PR and include the commit that fixes the tests.

Cargo.toml
```diff
@@ -44,6 +44,7 @@ num-derive = "0.4.0"
 num-traits = "0.2.6"
 p256 = { version = "^0.13", features = ["ecdsa"] }
 p384 = { version = "^0.13", features = ["ecdsa"] }
+libsecp256k1 = "0.7.1"
```
tarcieri, line 47: `k256` provides better performance than `libsecp256k1` and is sourced from @RustCrypto like `p256` and `p384`

LiveDuo: Didn't noticed in the first place. Just replaced it with `k256`.

src/crypto/ecdh.rs
```diff
@@ -140,7 +140,7 @@ pub fn decrypt(priv_key: &ECDHSecretKey, mpis: &[Mpi], fingerprint: &[u8]) -> Re
         None => 0,
     };
 
-    let mut encrypted_session_key_vec: Vec<u8> = Vec::new();
+    let mut encrypted_session_key_vec: Vec<u8> = vec![0; encrypted_key_len];
     encrypted_session_key_vec.resize(encrypted_key_len, 0);
```
dignifiedquire, line 144: no need for this line, if you initialize them above already

LiveDuo: `clippy::perf` suggested it but could change it back.

tests/key_test.rs
```diff
@@ -121,12 +121,12 @@ parse_dumps!(
     (test_parse_dumps_0, 0, 18_141, 20_998),
     (test_parse_dumps_1, 1, 18_021, 21_000),
     (test_parse_dumps_2, 2, 18_058, 20_999),
-    (test_parse_dumps_3, 3, 18_098, 20_999),
+    (test_parse_dumps_3, 3, 18_097, 20_998),
```
dignifiedquire, line 124: These numbers going down is suspicious, this likely means a secpk key that is not being able to verified/work with the impl. Needs checking out

LiveDuo: Does it mean it's one key that has the problem?

LiveDuo: Using `dbg!` in [src/packet/single.rs#L192](https://github.com/rpgp/rpgp/blob/master/src/packet/single.rs#L192) shows there's only one _new parsing error_ in the PR.

The error is `InvalidPacketContent(EllipticCurve(Error))` with a `PublicKey` tag and it's corresponding body is `040000000013052b8104000a02074615813d2d31329d51e5a8e7c8eaa65637e9ad5de00ea3fa65b2f1325aced0a527b6d28e1818e1f3d797d37f3c7c8256ff5f711756255a25f422224540966ed708`.

If there's a simple way to decode that hex, we might see exactly what message it is possibly why it fails.

LiveDuo:
From https://github.com/rpgp/rpgp/pull/249#discussion_r1299233154:

> Using dbg! in [src/packet/single.rs#L192](https://github.com/rpgp/rpgp/blob/master/src/packet/single.rs#L192) shows there's only one new parsing error in the PR.
> 
> The error is InvalidPacketContent(EllipticCurve(Error)) with a PublicKey tag and it's corresponding body is 040000000013052b8104000a02074615813d2d31329d51e5a8e7c8eaa65637e9ad5de00ea3fa65b2f1325aced0a527b6d28e1818e1f3d797d37f3c7c8256ff5f711756255a25f422224540966ed708.
> 
> If there's a simple way to decode that hex, we might see exactly what message it is possibly why it fails.

src/crypto/ecdh.rs
LiveDuo: Reverting the change causes Github actions to fail ie. [rpgp/actions/runs/5912769030](https://github.com/LiveDuo/rpgp/actions/runs/5912769030/job/16036499778).

dignifiedquire:
> if there's a simple way to decode that hex, we might see exactly what message it is possibly why it fails.

Unfortunately I don't know of any, what I usually do is sth like 

```
> gpg --list-packets --verbose problematic_key.asc
```

which gives you a detailed list of the packets in the key, and then inspect which piece it is that is failing and where.

LiveDuo:
> Unfortunately I don't know of any, what I usually do is sth like
> 
> ```
> > gpg --list-packets --verbose problematic_key.asc
> ```
> 
> which gives you a detailed list of the packets in the key, and then inspect which piece it is that is failing and where.

@dignifiedquire Thanks for the suggestion.

I'm wondering if there's a way to convert that hex to an armored string or import it directly into `pgp`. In other cases I would use [from_armor_single](https://docs.rs/pgp/latest/pgp/composed/trait.Deserializable.html#method.from_armor_single) but that's what's we want fixed. Is there another way to export the key from the [sks-dump/0003.pgp](https://github.com/rpgp/tests/blob/master/sks-dump/0003.pgp) to import it into `gpg` and inspect the packets?

dignifiedquire:
Ah yes, you should be able to import the dump directly into `gpg`, it understand the binary format. Alternatively you can write the single key in binary into a file like `key.gpg` and that should be importable the same way `.asc` files are into gpg

LiveDuo:
@dignifiedquire

Thanks for the tip. I think I come closer to the bottom of the issue.

At first I ran:
```sh
gpg --list-packets --verbose ./rpgp/tests/tests/sks-dump/0003.pgp | grep "version 4, algo 19" -C 10
```
and located the problematic packet.

Here's the output of `list-packet` for that key:
```sh
# off=10419589 ctb=98 tag=6 hlen=2 plen=79
:public key packet:
	version 4, algo 19, created 0, expires 0
	pkey[0]: 052B8104000A secp256k1 (1.3.132.0.10)
	pkey[1]: 4615813D2D31329D51E5A8E7C8EAA65637E9AD5DE00EA3FA65B2F1325ACED0A527B6D28E1818E1F3D797D37F3C7C8256FF5F711756255A25F422224540966ED708
	keyid: F50C012FB632C4B9
```

If you look closely the public key (`pkey[1]`) has a format that does not start with a `02` or a `04` (for compressed and uncompressed keys respectively). All other `secp256k1` public keys in the tests start with these known prefixes.

I haven't seen a `secp256k1` public key in the past that has this format but it may very well be a format that I'm not aware of.

Also note that `gpg` does not process this key because it contains no user id. This is the output:
```
rpgp-decode % gpg --import key_bin.gpg                  
gpg: key F50C012FB632C4B9: new key but contains no user ID - skipped
gpg: Total number processed: 1
gpg:           w/o user IDs: 1
```


Any idea what this format might be?

dignifiedquire:
thanks a lot, merged in #275

PR #253, opened 2023-10-16
chore: Fix deprecation warning reported by clippy

wiktor-k:


wiktor-k:
If you don't mind I'll be fixing and adjusting stuff here and there (when time permits). Thanks for taking the time for review! :wave:

dignifiedquire:
of course, thanks a lot

wiktor-k:
It seems the `mips-unknown-linux-gnu` target test fails (`error: component 'rust-std' for target 'mips-unknown-linux-gnu' is unavailable for download for channel 'stable'`). I see this target has only Tier 3 support so it seems plausible that the stable target is not yet available: https://doc.rust-lang.org/nightly/rustc/platform-support.html

Do you have clients using MIPS? (maybe some mobile phones etc.) If so maybe marking the job step with [`continue-on-error`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) would be a good idea (or some else mechanism no to block the PR when that fails).

dignifiedquire:
> Do you have clients using MIPS?

no known users, just trying to support as wide a range as possible

> If so maybe marking the job step with [`continue-on-error`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) would be a good idea (or some else mechanism no to block the PR when that fails).

sigh, I guess sth like that needs to be done, though honestly maybe just pinning to `1.72` would also be fine, not much has happened that I would worry latest stable releases breaking the code

wiktor-k:
> sigh, I guess sth like that needs to be done, though honestly maybe just pinning to 1.72 would also be fine, not much has happened that I would worry latest stable releases breaking the code

Okay, let's see if it fixes the job :)

wiktor-k:
```
error: component 'rust-std' for target 'mips-unknown-linux-gnu' is unavailable for download for channel '1.72'
```

:disappointed:

dignifiedquire:
going to merge anyway, as it makes at least more CI happy 😅

PR #258, opened 2023-12-08
perf: avoid `concat` in sha1 checksum impl

dadleyy:
When decrypting large files, the `data` used in the `sym::SymmetricKeyAlgorithm::decrypt_protected` method would be copied via the `concat` used when taking the checksum.

This change updates the function to avoid the copy, by iterating over the relevant parts of the checksum as immutable references. This should be "transparent" to the overall api of the lib so I thought it might make a nice change.

### Use Case

I am using `gpg` to create large filesystem updates (basically, `.ext4.gpg`) that can be several hundred MB in size, destined for embedded machines with only several hundred MB of RAM. On those machines, I am using this library in my application to decrypt the updates and apply them. Currently, I am running out of memory during the process.

I did some testing to verify that `concat` does indeed result in allocations here: [gist.github.com/dadleyy/a7f8f01](https://gist.github.com/dadleyy/a7f8f01d9e79dcc51641e72becc22467), using the `Instruments` application on my macOS machine.

### DIFF
--- a/src/crypto/checksum.rs
+++ b/src/crypto/checksum.rs
@@ -82,6 +82,14 @@ impl Hasher for SimpleChecksum {
 
 /// SHA1 checksum, first 20 octets.
 #[inline]
-pub fn calculate_sha1(data: &[u8]) -> Vec<u8> {
-    Sha1::digest(data)[..20].to_vec()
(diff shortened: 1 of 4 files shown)


dignifiedquire (approved):
thanks, makes sense 👍

PR #260, opened 2023-12-17
fix: check SHA-1 checksum before parsing

### LINKED ISSUE #46: SHA-1 checksum of the Secret-Key packet is not checked
Now trying to open a key encrypted with a password fails with an error similar to `Incomplete(Size(6017))`. It is not clear from this error whether the key file is indeed truncated or if the password provided is wrong.

See #45

link2xt:
Fixes #46


dignifiedquire:
@link2xt can you rebase this so we can get it merged?

dignifiedquire (approved):

PR #262, opened 2024-01-15
fix: Don't return unhashed subpackets in `subpackets()`

hko-s:
It's unsafe for consumers to rely on data from an unhashed subpacket as though it were self-signed by the certificate owner. Applications that need to interpret data from an unhashed subpacket should opt in to this using explicit code.

### DIFF
--- a/src/packet/signature/config.rs
+++ b/src/packet/signature/config.rs
@@ -308,11 +308,23 @@ impl SignatureConfig {
         }
     }
 
-    /// Returns an iterator over all subpackets of this signature.
+    /// Returns an iterator of all subpackets in the signature: all subpackets in the hashed area
+    /// followed by all subpackets in the unhashed area.
+    #[deprecated(
+        note = "Usually only hashed_subpackets should be used. unhashed_subpackets are only safe and useful to access in rare circumstances. When they are needed, unhashed_subpackets should be explicitly called."
+    )]
     pub fn subpackets(&self) -> impl Iterator<Item = &Subpacket> {
-        self.hashed_subpackets
-            .iter()
-            .chain(self.unhashed_subpackets.iter())
+        self.hashed_subpackets().chain(self.unhashed_subpackets())
+    }
+
+    /// Returns an iterator over the hashed subpackets of this signature.
+    pub fn hashed_subpackets(&self) -> impl Iterator<Item = &Subpacket> {
+        self.hashed_subpackets.iter()
+    }
+
+    /// Returns an iterator over the unhashed subpackets of this signature.
+    pub fn unhashed_subpackets(&self) -> impl Iterator<Item = &Subpacket> {
+        self.unhashed_subpackets.iter()
     }
 
     /// Returns if the signature is a certification or not.
@@ -327,26 +339,43 @@ impl SignatureConfig {
         )
     }
 
+    /// Signature Creation Time.
+    ///
+    /// The time the signature was made.
+    /// MUST be present in the hashed area.
+    ///
+    /// https://datatracker.ietf.org/doc/html/rfc4880#section-5.2.3.4
+    ///
+    /// Returns the first Signature Creation Time subpacket, only from the hashed area.
     pub fn created(&self) -> Option<&DateTime<Utc>> {
         if self.created.is_some() {
             return self.created.as_ref();
         }
 
-        self.subpackets().find_map(|p| match p.data {
+        self.hashed_subpackets().find_map(|p| match p.data {
             SubpacketData::SignatureCreationTime(ref d) => Some(d),
             _ => None,
         })
     }
 
     pub fn issuer(&self) -> Option<&KeyId> {
-        if self.issuer.is_some() {
-            return self.issuer.as_ref();
+        // legacy v2/v3 signatures have an explicit "issuer" field
+        if let Some(issuer) = self.issuer.as_ref() {
+            return Some(issuer);
         }
 
-        self.subpackets().find_map(|p| match p.data {
-            SubpacketData::Issuer(ref id) => Some(id),
-            _ => None,
-        })
+        // v4+ signatures use subpackets
+        //
+        // We consider data from both the hashed and unhashed area here, because the issuer Key ID
+        // only acts as a hint. The signature will be cryptographically checked using the purported
+        // issuer's key material. An attacker cannot successfully claim an issuer Key ID that they
+        // can't produce a cryptographically valid signature for.
+        self.hashed_subpackets()
+            .chain(self.unhashed_subpackets())
+            .find_map(|p| match p.data {
+                SubpacketData::Issuer(ref id) => Some(id),
+                _ => None,
+            })
     }
 }
 
--- a/src/packet/signature/config.rs
+++ b/src/packet/signature/config.rs
@@ -358,10 +358,17 @@ impl SignatureConfig {
         })
     }
 
-    pub fn issuer(&self) -> Option<&KeyId> {
+    /// Issuer.
+    ///
+    /// The OpenPGP Key ID of the key issuing the signature.
+    ///
+    /// https://datatracker.ietf.org/doc/html/rfc4880#section-5.2.3.5
+    ///
+    /// Returns Issuer subpacket data from both the hashed and unhashed area.
+    pub fn issuer(&self) -> Vec<&KeyId> {
         // legacy v2/v3 signatures have an explicit "issuer" field
         if let Some(issuer) = self.issuer.as_ref() {
-            return Some(issuer);
+            return vec![issuer];
         }
 
         // v4+ signatures use subpackets
@@ -372,10 +379,31 @@ impl SignatureConfig {
         // can't produce a cryptographically valid signature for.
         self.hashed_subpackets()
             .chain(self.unhashed_subpackets())
-            .find_map(|p| match p.data {
+            .filter_map(|sp| match sp.data {
                 SubpacketData::Issuer(ref id) => Some(id),
                 _ => None,
             })
+            .collect()
+    }
+
+    /// Issuer Fingerprint.
+    ///
+    /// The OpenPGP Key fingerprint of the key issuing the signature.
+    ///
+    /// This subpacket type was introduced after RFC 4880, in the RFC 4880-bis lifecycle.
+    /// It sees some use in the wild for v4 signatures, in both the hashed and unhashed areas.
+    ///
+    /// https://datatracker.ietf.org/doc/html/draft-ietf-openpgp-rfc4880bis-10#name-issuer-fingerprint
+    ///
+    /// Returns Issuer Fingerprint subpacket data from both the hashed and unhashed area.
+    pub fn issuer_fingerprint(&self) -> Vec<&[u8]> {
+        self.hashed_subpackets()
+            .chain(self.unhashed_subpackets())
+            .filter_map(|sp| match &sp.data {
+                SubpacketData::IssuerFingerprint(_, fp) => Some(fp.as_slice()),
+                _ => None,
+            })
+            .collect()
     }
 }
 
--- a/src/packet/signature/types.rs
+++ b/src/packet/signature/types.rs
@@ -80,20 +80,36 @@ impl Signature {
         self.config.typ()
     }
 
+    /// Does `key` match any issuer or issuer_fingerprint subpacket in `sig`?
+    /// If yes, we consider `key` a candidate to verify `sig` against.
+    ///
+    /// We also consider `key` a match for `sig` by default, if `sig` contains no issuer-related
+    /// subpackets.
+    fn match_identity(sig: &Signature, key: &impl PublicKeyTrait) -> bool {
+        let issuers = sig.issuer();
+        let issuer_fps = sig.issuer_fingerprint();
+
+        // If there is no subpacket that signals the issuer, we consider `sig` and `key` a
+        // potential match, and will check the cryptographic validity.
+        if issuers.is_empty() && issuer_fps.is_empty() {
+            return true;
+        }
+
+        // Does any issuer or issuer fingerprint subpacket matche the identity of `sig`?
+        issuers.iter().any(|&key_id| key_id == &key.key_id())
+            || issuer_fps.iter().any(|&fp| fp == key.fingerprint())
+    }
+
     /// Verify this signature.
     pub fn verify<R>(&self, key: &impl PublicKeyTrait, data: R) -> Result<()>
     where
         R: Read,
     {
-        if let Some(issuer) = self.issuer() {
-            if &key.key_id() != issuer {
-                bail!(
-                    "validating signature with a non matching Key ID {:?} != {:?}",
-                    &key.key_id(),
-                    issuer
-                );
-            }
-        }
+        ensure!(
+            Self::match_identity(self, key),
+            "verify: No matching issuer or issuer_fingerprint for Key ID: {:?}",
+            &key.key_id(),
+        );
 
         let mut hasher = self.config.hash_alg.new_hasher()?;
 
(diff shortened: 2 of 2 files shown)


src/packet/signature/config.rs
```diff
@@ -299,13 +299,18 @@ impl SignatureConfig {
         }
     }
 
-    /// Returns an iterator over all subpackets of this signature.
+    /// Returns an iterator over the hashed subpackets of this signature.
     pub fn subpackets(&self) -> impl Iterator<Item = &Subpacket> {
         self.hashed_subpackets
             .iter()
             .chain(self.unhashed_subpackets.iter())
```
wiktor-k, line 306: Wait... are you still returning unhashed packets here? :thinking: 

(The idea behind the change looks good, though).

hko-s: Ouch, thanks for checking. Not sure how I managed to commit this confused version :sweat_smile: 
I pushed an amended fix.

src/packet/signature/config.rs
```diff
@@ -299,11 +299,14 @@ impl SignatureConfig {
         }
     }
 
-    /// Returns an iterator over all subpackets of this signature.
+    /// Returns an iterator over the hashed subpackets of this signature.
     pub fn subpackets(&self) -> impl Iterator<Item = &Subpacket> {
```
dignifiedquire, line 303: I think if we do this, this method should be renamed to `hashed_subpackets`

src/packet/signature/types.rs
```diff
@@ -271,7 +271,7 @@ impl Signature {
         self.config.is_certificate()
     }
 
-    /// Returns an iterator over all subpackets of this signature.
+    /// Returns an iterator over the hashed subpackets of this signature.
     fn subpackets(&self) -> impl Iterator<Item = &Subpacket> {
```
dignifiedquire, line 275: same naming concern

src/packet/signature/config.rs
wiktor-k: I'd say rename, and add a `subpackets` function marked with `#[deprecated(note = "Use hashed_subpackets")]` that delegates to hashed_subpackets.

hko-s:
I pushed an additional commit that renames both occurrences of `subpacket()`.

This diff makes explicit that this change affects many convenience getters for specific subpackets. In most cases, I am relatively convinced that the semantics of implicitly retrieving subpacket data from the unhashed area was problematic, before. For a few subpackets, I'll grant that this change *might* be surprising to applications/users.

(However, compared to problems such as accidentally relying on maliciously added subpackets from an unhashed area, I'd think those are relatively unlikely edge cases)

src/packet/signature/config.rs
hko-s: > I'd say rename, and add a `subpackets` function marked with `#[deprecated(note = "Use hashed_subpackets")]` that delegates to hashed_subpackets.

Good point. I did this in 31aa70a0ed197fb196886175be22e0b2b40b6f34 (and am happy to clean up the commit history/comments, if/when we arrive at a consensus for an appraoch)

src/packet/signature/config.rs
```diff
@@ -299,11 +299,19 @@ impl SignatureConfig {
...
     }
 
-    /// Returns an iterator over all subpackets of this signature.
-    pub fn subpackets(&self) -> impl Iterator<Item = &Subpacket> {
-        self.hashed_subpackets
-            .iter()
-            .chain(self.unhashed_subpackets.iter())
+    #[deprecated(note = "Use hashed_subpackets")]
+    pub fn subpackets(&self) -> impl Iterator<Item=&Subpacket> {
+        self.hashed_subpackets()
```
dignifiedquire, line 304: the deprecation is a good idea, but we should keep the functionality as is in this case, otherwise this is still a breaking change

hko-s: My perspective is that the previous implementation was a bug, and that this PR shifts the semantics to what they should always have been. With the deprecation warning, as it currently stands, users of the API get notified and have a chance to reflect if they actually intended to read from the unhashed area.

But of course this is arguable.

And the question of how to deal with adjusting the subpacket getter semantics is even more tricky: Note that the semantics of subpacket getters in `types.rs` are changed by this PR.

However, while this is formally a "breaking change", I wouldn't consider this change "breaking".
Unfortunately, the RFC doesn't offer clear guidance about which subpacket types are reasonable to store in the unhashed area.

hko-s:
Taking a step back, I have a few more thoughts on subpackets - and about potential breaking changes to them:

- First off, a it's possible to access the raw subpacket data via the SignatureConfig struct, my concerns in this issue are mostly about convenience and safety of the API. Users of rpgp are not limited in implementing whatever semantics they want. So that's good :)
- I remain relatively convinced that it's a bad tradeoff to offer convenience wrappers that don't tell users if a subpacket came from the hashed or unhashed subpacket area. A lot of PGP's complexity revolves around interpretation of data that is in signature subpackets, and how to correctly interpret this data. I wonder if there are plausible cases where an application really wants to implicitly access subpackets in the unhashed area - on the other hand it seems easy to start coming up with attacks against applications that don't realize they might be consuming unhashed metadata from subpackets.
- An unpleasant papercut is the type of the data in `SignatureExpirationTime` and `KeyExpirationTime` . Those types should be durations, not timestamps (there is code that compensates for this within rpgp, in at least one place: `pgp::composed::signed_key::shared::SignedKeyDetails`). Fixing this would seem nice to me.

Of these, I'd be quite eager to see the expiration time type issue fixed. While this is a breaking change, I imagine it's not particularly controversial. Let me know if I should open a separate ticket for that.

If we don't arrive at an approach to this issue that we are all excited about, we could decide to let it sit. (I'm worried about, but not blocked by, the content of this issue.)

src/packet/signature/config.rs
hko-s: Right, that's fair. I still worry about how applications might use these, but maybe that's not actionable in a reasonable way.

At this point, i am happy to drop this issue, for now. In my use of rpgp, I shouldn't use any of these accessors. The abstraction they provide is not right for my use case.

Maybe in the long run, the right thing to do is to point users at a higher level abstraction library, and to deprecate these subpacket access helper functions in rpgp?

dignifiedquire: > Maybe in the long run, the right thing to do is to point users at a higher level abstraction library, and to deprecate these subpacket access helper functions in rpgp?

Probably for most use cases, I think it's good for lower level tools to have it though.

hko-s: Exposing low level access to subpackets from rpgp is reasonable, right. But then the access should expose the OpenPGP data without interpreting it.

Most importantly, it should not conflate the two subpacket areas - they are separated in the packet format precisely because their semantics are very different, even though their data format is of course identical between the two subpacket areas.

If we want to go in that direction (which I'd find worthwhile), I could propose a change to the API to expose uninterpreted access to the subpackets.

dignifiedquire: yes that would be nice

dignifiedquire:
the latest version looks great from the direction

hko-s:
> the latest version looks great from the direction

Great!

I really need to sort out the git history :sweat_smile:, but in terms of functionality and API, I'm arriving roughly where I'd like to be.

src/packet/signature/types.rs
```diff
@@ -219,22 +222,23 @@ impl Signature {
...
         );
 
-        let key_id = signing_key.key_id();
-        if let Some(issuer) = self.issuer() {
-            if &key_id != issuer {
+        // don't check issuer/issuer_fp subpacket match for backsigs. the subpackets are inconsistent in the test suite ("Binding signature subpackets"). (why?)
+        // just always verify is the signature itself is valid, regardless of issuer(_fp) subpackets.
+        if !backsig {
+            let key_id = signer.key_id();
+            if !Self::match_identity(self, signer) {
```
hko-s, line 238: I tend to think that this `match_identity` check should be entirely removed. A key binding, just by existing, always implicitly claims to be a self signature.

So I think key binding signatures should just be cryptographically checked, without taking issuer subpackets into account.

hko-s:
I've readjusted the commit history, and think this PR is now ready for serious consideration.

src/packet/signature/config.rs
```diff
@@ -332,21 +342,36 @@ impl SignatureConfig {
...
+            .chain(self.unhashed_subpackets())
+            .filter_map(|sp| match sp.data {
+                SubpacketData::Issuer(ref id) => Some(id),
+                _ => None,
+            })
+            .collect()
+    }
+
+    pub fn issuer_fingerprint(&self) -> Vec<&[u8]> {
```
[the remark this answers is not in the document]
hko-s: good point. i've added some API doc for the subpaket getters.
extending the API documentation more generally sounds like a good future project.

i could do some of that.

src/packet/signature/types.rs
```diff
@@ -80,19 +80,25 @@ impl Signature {
...
     {
-        if let Some(issuer) = self.issuer() {
-            if &key.key_id() != issuer {
-                bail!(
-                    "validating signature with a non matching Key ID {:?} != {:?}",
-                    &key.key_id(),
-                    issuer
-                );
-            }
+        if !Self::match_identity(self, key) {
```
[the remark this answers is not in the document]
hko-s: good pointer, thanks!

src/packet/signature/types.rs
```diff
@@ -219,22 +222,23 @@ impl Signature {
...
-            "verifying key binding: {:#?} - {:#?} - {:#?}",
-            self, signing_key, key
+            "verifying key binding: {:#?} - {:#?} - {:#?} (backsig: {})",
+            self, signer, signee, backsig
         );
 
-        let key_id = signing_key.key_id();
-        if let Some(issuer) = self.issuer() {
-            if &key_id != issuer {
+        // don't check issuer/issuer_fp subpacket match for backsigs. the subpackets are inconsistent in the test suite ("Binding signature subpackets"). (why?)
```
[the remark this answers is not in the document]
hko-s: Specifically, some of the embedded "back signatures" in https://tests.sequoia-pgp.org/#Binding_signature_subpackets seem to use the primary key's fingerprint as the "issuer fingerprint".

This is odd, because that embedded signature is issued by the signing subkey, which the embedded signature binds to the primary key. However, based on those test cases, I assume that such signatures exist in the wild.

My suggestion is: I would entirely drop the `match_identity` check in `verify_key_binding_internal`. I think it's pragmatic to just always check the cryptographic signature, and not to bother with interpreting the issuer subpackets. Key binding signatures are always "self signatures".
There is exactly one candidate key packet that can have created a key binding signature (including embedded back signatures). Verifying the signature with that one candidate key is straightforward.

(This suggestion applies only to key binding signatures within TPKs, as done in `verify_key_binding_internal`. I would keep the `match_identity` check in the three other locations).

src/packet/signature/types.rs
hko-s: Alright, i pushed a separate commit that drops this block.

If CI is happy, and we have consensus, I can squash that commit up.

src/packet/signature/types.rs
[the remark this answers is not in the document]
hko-s: (linking to https://github.com/rpgp/rpgp/pull/262#pullrequestreview-1887781505, because that thread discusses the same point)

src/packet/signature/types.rs
```diff
@@ -80,20 +80,25 @@ impl Signature {
         self.config.typ()
     }
 
+    /// Does `key` match any issuer or issuer_fingerprint subpacket in `sig`?
+    fn match_identity(sig: &Signature, key: &impl PublicKeyTrait) -> bool {
+        sig.issuer().iter().any(|&key_id| key_id == &key.key_id())
```
hko-s, line 85: One remark about this: before this patch, rpgp would validate signatures if *no* issuer subpackets existed in both areas.

This implementation does not. If no matching issuer or issuer fingerprint is found, it refuses to verify signatures.

In the tests we have, this doesn't seem to cause problems. However, I'm not entirely sure if there are edge cases that this PR breaks.

A possible addition that seems potentially attractive to me, would be to also return `true` in this helper if the hashed area contains neither an `Issuer` nor an `IssuerFingerprint` subpacket (in that case, the implementation that made the signature apparently did not want to put it on the record who issued the signature, but if we have an idea which key to verify it with, there is no downside to use trying).

(`ensure!` ing on `match_identity` is only an optimization that avoids spending the CPU cycles to check the signature. But at the risk that we reject a signature as not valid, even though it cryptographically is)

hko-s: After some more thought and discussion with @wiktor-k, I propose we tweak the logic as shown in 3c24ff17142511ae438ad9ed645fd16bd487f8f6 (i'll squash that commit up, if there is agreement)

hko-s: This should mostly align the semantics of this PR with the previous state, but gracefully handle the existence of multiple `Issuer` fields and `IssuerFingerprint`s.

dignifiedquire (approved):
I think the final solution is a good one, thanks

hko-s:
> I think the final solution is a good one, thanks

Alright. Shall I squash the extra little commits?

hko-s:
Done. The commit history now seems reasonable to me.

src/packet/signature/config.rs
```diff
@@ -327,26 +339,71 @@ impl SignatureConfig {
...
-        if self.issuer.is_some() {
-            return self.issuer.as_ref();
+    /// Issuer.
+    ///
+    /// The OpenPGP Key ID of the key issuing the signature.
+    ///
+    /// https://datatracker.ietf.org/doc/html/rfc4880#section-5.2.3.5
+    ///
+    /// Returns Issuer subpacket data from both the hashed and unhashed area.
+    pub fn issuer(&self) -> Vec<&KeyId> {
```
dralley, line 368: I get that this is theoretically possible - but what is the situation in which it is actually reasonable to have more than one issuing key ID on a signature?  

My understanding of OpenPGP is kind of shallow, but I can't figure out why that would actually be a reasonable thing in practice.

hko-s: To first sidestep your question a little: Since the OpenPGP format allows signatures that contain multiple Issuers, it seems like the lesser evil for a low level API like rPGP to expose this to applications.

That said, the format generally allows extreme flexibility for signature subpackets, and I don't personally love this aspect of the format.

I believe there was some historical edge case where some people actually added multiple Issuer subpackets to signatures, to model something at least half reasonable (maybe to do with the migration from v3 to v4 keys?).
But I am entirely unsure about details right now.

For any half modern data, I think it's reasonable for applications to consider the existence of more than one Issuer subpacket an error (so, edge cases aside: I agree with your assessment that it is usually unreasonable to have)

dralley: Thanks for the response!

DemiMarie: Is it possible to have _no_ issuer subpackets?

hko-s: Yes.

Note that the "issuer" subpacket contains a 64 bit KeyId. An alternative (and newer) subpacket is the "issuer fingerprint", which contains a full fingerprint (Sha1 for v4 keys, Sha256 for v6 keys).

As I understand it, applications that want to deal with general OpenPGP data should check for both of these subpackets (and check for them in both the hashed and unhashed areas).
As always, the unhashed subpackets could be attacker controlled, but they can still serve as a useful hint to figure out which public key might verify that signature (the ultimate signal is when signature validation is happy. All of the issuer and issuer fingerprint subpackets only serve as hints that help consumers to arrive at happy signature validation).

(As far as OpenPGP the format is concerned, I believe an application is free to produce signature packets without any hints about the signer, if it believes that the consumer will be able to figure out which public key to verify the signature with)

wiktor-k: > All of the issuer and issuer fingerprint subpackets only serve as hints that help consumers to arrive at happy signature validation).

Agreed. On one hand these could help applications discover and fetch signing keys. On the other hand the fingerprint subpacket can be used to filter out signing keys that have the [same key ID](https://evil32.com/) but different fingerprints.

Maybe it'd be good to clarify [on the OpenPGP Mailing List](https://mailman3.ietf.org/mailman3/lists/openpgp@ietf.org/) what happens in the case where the signature is cryptographically valid but the issuer subpacket contains wrong data?

PR #263, opened 2024-01-15
fix: Normalize line endings when hashing data for a Text signature

hko-s:
See https://www.rfc-editor.org/rfc/rfc4880.html#section-5.2.1, which says that for signature type 0x01, "The signature is calculated over the text data with its line endings converted to `<CR><LF>`."


hko-s:
This PR is unfortunate in that it makes a copy of the literal data. If there is a reasonable way to have the same semantics in a streaming way, that would of course be much better (I failed to find one, given the existing data types and function signatures, but that might be entirely my mistake).

src/composed/message/types.rs
```diff
@@ -398,7 +400,21 @@ impl Message {
             } => {
                 if let Some(message) = message {
                     match **message {
-                        Message::Literal(ref data) => signature.verify(key, data.data()),
+                        Message::Literal(ref data) => {
+                            match signature.typ() {
```
wiktor-k, line 404: Instead of two arm match maybe a simpler `if let` would be sufficient?

src/composed/message/types.rs
```diff
@@ -398,7 +400,21 @@ impl Message {
...
+                                SignatureType::Text => {
+                                    // Normalize line endings for "Signature of a canonical text document",
+                                    // see https://www.rfc-editor.org/rfc/rfc4880.html#section-5.2.1
+                                    let normalized: Vec<_> = Normalized::new(
+                                        data.data().iter().cloned(),
+                                        LineBreak::Crlf,
+                                    ).collect();
+
+                                    signature.verify(key, &*normalized)
```
wiktor-k, line 413: The problem here seems to be that `Normalized` doesn't implement `Read` but rather is an iterator of u8s. There are ways to adapt iterator like so: https://crates.io/crates/iter_read but I'm not sure additional dependency is worth it :man_shrugging:

src/composed/message/types.rs
hko-s: > Instead of two arm match maybe a simpler `if let` would be sufficient?

Note that the outer match is pre-existing code. I don't think this PR should change that code?
I pushed a variation of my change that uses `matches!`, which I guess is slightly nicer to read.

src/composed/message/types.rs
hko-s: Thanks for that pointer! I'd be happy to adjust the PR to use that crate (or another similar mechanism).
My tendency would be to try and enable streaming operation.

wiktor-k:
Btw I don't know if it's just me but it'd be cool to have a test with a signature that used this kind of data which broke previously but would work now (as a "regression test").

src/composed/message/types.rs
wiktor-k: Maybe wait until @dignifiedquire decides... I'm just throwing random ideas here :)

hko-s:
I'm going to replace the copying with use of `iter_read` (as suggested by @wiktor-k), it looks tiny and the copying seems wasteful.

And I'll add a unit test for this.

dignifiedquire:
yes avoiding the copy would be very nice

hko-s:
It took me a while to recall when this patch is relevant:

When a signature is made with type 0x01 ("text"), but the data packet is marked "binary", then the data packet may contain non-normalized line endings, but the hash must (presumably) be calculated over normalized line endings.

I also added a unit test. The test reflects a message that fails to verify without this PR, and succeeds, with it. If we want keep it (see caveat below), I assume it would be good to move the test-key and -message to a file in the filesystem.
But am unsure about the organization of test files in the rpgp repo. What would be good folders/naming conventions for this test?

## Caveat

Note that while https://tests.sequoia-pgp.org/#Signed_messages (test case "Text sig / Binary") shows that some implementations can interoperably verify this type of signature, GnuPG is not tested there.

My attempt to manually test verification with GnuPG 2.4.3 just now showed that GnuPG doesn't consider this message correctly signed:

```
$ gpg --verify msg
gpg: Signature made Tue 23 Jan 2024 08:59:12 PM CET
gpg:                using EDDSA key 97DC159B71246121F2207E2264357EB6BB55DE12
gpg: BAD signature from "<alice@example.org>" [unknown]
```

It's not entirely clear to me what should follow from this set of facts.

hko-s:
Ah, no. Sorry, the unit test is still broken. This will take a bit more thought.

dignifiedquire:
> I assume it would be good to move the test-key and -message to a file in the filesystem.
> But am unsure about the organization of test files in the rpgp repo. What would be good folders/naming conventions for this test?

Given that the unit tests so far haven't used files, there is no good structure yet. Something sensible inside the `tests` folder seems good to me

dignifiedquire:
> When a signature is made with type 0x01 ("text"), but the data packet is marked "binary", then the data packet may contain non-normalized line endings, but the hash must (presumably) be calculated over normalized line endings.

why would the data be touched in any way for hashing? I would as a user assume the data is left untouched, given it is marked as binary, and I am not seeing anything in the RFC which would indicate otherwise

hko-s:
> why would the data be touched in any way for hashing? I would as a user assume the data is left untouched, given it is marked as binary, and I am not seeing anything in the RFC which would indicate otherwise

I believe the reasoning *for* normalizing in in https://www.rfc-editor.org/rfc/rfc4880.html#section-5.2.1:

```
"0x01: Signature of a canonical text document.
       This means the signer owns it, created it, or certifies that it
       has not been modified.  The signature is calculated over the text
       data with its line endings converted to <CR><LF>."
```

This is definitely an edge case of sorts. And I'll grant that I might be confused about some aspects (so I appreciate pushback).

There is some discussion here: https://gitlab.com/sequoia-pgp/openpgp-interoperability-test-suite/-/issues/63 (including links to more discussion from the original post).

hko-s:
> why would the data be touched in any way for hashing? I would as a user assume the data is left untouched, given it is marked as binary, and I am not seeing anything in the RFC which would indicate otherwise

One more argument why hashing the normalized literal data packet is correct:

The signature packet can be split out of this inline signed message. It can the be verified as a detached signature. After detaching it, GnuPG consider the signature in the test valid.

I'd count this as an argument that GnuPG is incorrectly processing the inline signed message, when the Literal Data Packet is set to "binary".

Two things follow from that, to my mind:
- It's better not to produce messages with SignatureType Text and Literal Data Packets set to Binary, with non-normalized line endings
- But when one encounters such a packet, it's more defensive to hash a line-ending normalized version of the data

hko-s:
For future reference, an example run of handling this type of signature, with the `sqop` SOP implementation and with GnuPG:

# Setting up a key and signed text message with non-normalized line endings

```shell
$ sqop version --extended
sqop 0.31.0
sop-rs 0.7.0
Nettle 3.9 (Cv448: true, OCB: true)
Sequoia 1.17.0

$ sqop generate-key "<alice@example.org>" > alice.priv
$ echo -ne "foo\nbar" | sqop inline-sign --as text alice.priv > msg
$ sq packet dump --hex msg
One-Pass Signature Packet, new CTB, 2 header bytes + 13 bytes
    Version: 3
    Type: Text
    Pk algo: EdDSA
    Hash algo: SHA512
    Issuer: A347C9B887511DAC
    Last: true

    00000000  c4                                                 CTB
    00000001     0d                                              length
    00000002        03                                           version
    00000003           01                                        type
    00000004              0a                                     hash_algo
    00000005                 16                                  pk_algo
    00000006                    a3 47  c9 b8 87 51 1d ac         issuer
    0000000e                                             01      last

Literal Data Packet, new CTB, 2 header bytes + 13 bytes
    Format: Binary data
    Content: "foo\nbar"

    00000000  cb                                                 CTB
    00000001     0d                                              length
    00000002        62                                           format
    00000003           00                                        filename_len
    00000004              00 00 00 00                            date
    00000008                           66 6f 6f 0a 62 61 72              foo.bar

Signature Packet, new CTB, 2 header bytes + 189 bytes
    Version: 4
    Type: Text
    Pk algo: EdDSA
    Hash algo: SHA512
    Hashed area:
      Signature creation time: 2024-01-23 21:45:51 UTC (critical)
      Issuer: A347C9B887511DAC
      Notation: salt@notations.sequoia-pgp.org
        00000000  d8 67 8b 82 95 71 37 d1  46 84 20 77 98 4f 0a 13
        00000010  0b da 27 cd a5 db 08 ee  f8 d2 b3 44 a7 81 b6 c8
      Issuer Fingerprint: 0432632608AA555574F154EAA347C9B887511DAC
    Digest prefix: 11FB
    Level: 0 (signature over data)

    00000000  c2                                                 CTB
    00000001     bd                                              length
    00000002        04                                           version
    00000003           01                                        type
    00000004              16                                     pk_algo
    00000005                 0a                                  hash_algo
    00000006                    00 6f                            hashed_area_len
    00000008                           05                        subpacket length
    00000009                              82                     subpacket tag
    0000000a                                 65 b0 33 8f         sig creation time
    0000000e                                             09      subpacket length
    0000000f                                                10   subpacket tag
    00000010  a3 47 c9 b8 87 51 1d ac                            issuer
    00000018                           47                        subpacket length
    00000019                              14                     subpacket tag
    0000001a                                 00 00 00 00         flags
    0000001e                                             00 1e   name len
    00000020  00 20                                              value len
    00000022        73 61 6c 74 40 6e  6f 74 61 74 69 6f 6e 73   notation name
    00000030  2e 73 65 71 75 6f 69 61  2d 70 67 70 2e 6f 72 67
    00000040  d8 67 8b 82 95 71 37 d1  46 84 20 77 98 4f 0a 13   notation value
    00000050  0b da 27 cd a5 db 08 ee  f8 d2 b3 44 a7 81 b6 c8
    00000060  16                                                 subpacket length
    00000061     21                                              subpacket tag
    00000062        04                                           version
    00000063           04 32 63 26 08  aa 55 55 74 f1 54 ea a3   issuer fp
    00000070  47 c9 b8 87 51 1d ac
    00000077                       00  00                        unhashed_area_len
    00000079                              11                     digest_prefix1
    0000007a                                 fb                  digest_prefix2
    0000007b                                    00 fc            eddsa_sig_r_len
    0000007d                                          09 65 5b   eddsa_sig_r
    00000080  4a 04 80 39 96 42 7c e9  c9 49 5e 0c 3a 87 76 70
    00000090  e1 9d c9 fa de 44 1a c7  a3 c4 92 0e 98
    0000009d                                          01 00      eddsa_sig_s_len
    0000009f                                                b3   eddsa_sig_s
    000000a0  6e 65 4f 08 d2 e9 54 a5  1e 96 3e 4b b2 82 ff 96
    000000b0  5f 94 36 f4 12 2a c1 c2  ca b0 84 82 6e 89 00

$ cat msg | sqop inline-verify alice.pub
foo
bar
```

# Breaking out the signature as a detached signature

```shell
$ cat msg | sqop inline-detach --signatures-out sig

$ echo -ne "foo\r\nbar" | sqop verify sig alice.priv
2024-01-23T21:45:51Z 0432632608AA555574F154EAA347C9B887511DAC 54AEF4F6D668594B4991460ED99F5E3E1D069B0D mode:text

$ echo -ne "foo\nbar" | sqop verify sig alice.priv
2024-01-23T21:45:51Z 0432632608AA555574F154EAA347C9B887511DAC 54AEF4F6D668594B4991460ED99F5E3E1D069B0D mode:text
```

# GnuPG

```shell
$ echo -ne "foo\nbar" > msg.txt
$ cat alice.priv | sqop extract-cert > alice.pub
$ gpg --import alice.pub
```

## gpg doesn't like the inline signature

```shell
$ gpg --verify msg
gpg: Signature made Tue 23 Jan 2024 10:45:51 PM CET
gpg:                using EDDSA key 0432632608AA555574F154EAA347C9B887511DAC
gpg: BAD signature from "<alice@example.org>" [unknown]
```

## gpg likes the split out signature

```shell
$ gpg --verify sig msg.txt
gpg: Signature made Tue 23 Jan 2024 10:45:51 PM CET
gpg:                using EDDSA key 0432632608AA555574F154EAA347C9B887511DAC
gpg: Good signature from "<alice@example.org>" [unknown]
gpg: WARNING: This key is not certified with a trusted signature!
gpg:          There is no indication that the signature belongs to the owner.
Primary key fingerprint: 54AE F4F6 D668 594B 4991  460E D99F 5E3E 1D06 9B0D
     Subkey fingerprint: 0432 6326 08AA 5555 74F1  54EA A347 C9B8 8751 1DAC
```

hko-s:
> Ah, no. Sorry, the unit test is still broken. This will take a bit more thought.

The test is now fixed (with a bit of a mean hack: the primary is not the signing key, in this case, so I'm manually grabbing the signing subkey for signature verification)

hko-s:
> why would the data be touched in any way for hashing? I would as a user assume the data is left untouched, given it is marked as binary, and I am not seeing anything in the RFC which would indicate otherwise

An update: Only with this change does rpgp pass the "Text mode" interoperability test cases in https://tests.sequoia-pgp.org/#Signed_messages, including the base case (second line).

hko-s:
> Given that the unit tests so far haven't used files, there is no good structure yet. Something sensible inside the `tests` folder seems good to me

I've moved the test artifacts into files (happy to rename/move them, if the naming scheme feels off to you).

hko-s:
For reference, discussion about this normalization point, on the IETF list:
https://mailarchive.ietf.org/arch/msg/openpgp/2KN_wxt4CMoh_9-Qb5kp7epLWYQ/

dignifiedquire (approved):
lgtm, more interop is good, even if I really dislike the rfc here..

hko-s:
> lgtm, more interop is good, even if I really dislike the rfc here..

(Anecdotal evidence seems to suggest that most implementers have hated this "one weird trick" for a little while, after first running into the question of when and how to normalize.)

To spell out the reasoning for the normalization some more, here (to add to the already extensive wall of text):

With detached OpenPGP signatures over text documents, it makes sense to define a normalization step in the course of hashing: It is assumed that the line endings inside a signed text document may change (e.g. by transferring a text document to a DOS machine by FTP). To maintain a stable hash digest for a text file, the line endings are normalized during hashing, to compensate for potential platform-specific alterations to the line endings inside the file.

At least historically, this was crucial for stable signatures on detached text files (I'm actually not sure if modern systems still modify line endings routinely).

For inline signed messages, the case for normalization gets weirder, and less convincing. The literal data is presumably not going get altered once it has been wrapped in OpenPGP format.

However, if we consider transformations between inline signed messages and detached signatures a reasonable operation, then handling inline signed messages in the same way as detached signatures seems like the most reasonable approach (which means: normalizing line endings for signature type 0x01, even when the payload is inlined into the OpenPGP message).

hko-s:
After some more consideration, I now wonder if the normalization should in fact happen in ` Signature::verify`.

hko-s:
> After some more consideration, I now wonder if the normalization should in fact happen in ` Signature::verify`.

I've pushed an extra commit that does this. However, I am struggling to use `Normalized` in a non-awkward way :confused: 

Awkwardness in the code aside, though, I don't see any regressions from this change, in the SOP testsuite.
And the new semantics of the rpgp API seem cleaner to me.

dignifiedquire:
code looks reasonable to me, so lets get this in

hko-s:
> code looks reasonable to me, so lets get this in

If you're not bothered by `let normalized = Normalized::new(data.bytes().flat_map(|b| b.ok()), LineBreak::Crlf);`, then yes, let's go!

dignifiedquire:
I can live with it :)

PR #265, opened 2024-01-20
Handle other RevocationCode values

hko-s:
Currently, when reading a revocation signature with an unknown "revocation code" value, the "revocation code" field is dropped during parsing, so it's missing from the resulting `Signature` struct. As a result, `Signature::verify_key()`  considers the cryptographic signature to be wrong.

So applications have to assume such a revocation is "fake", and ignore it, because the self-signature is broken. In such a case, applications will ignore a legitimately self-signed revocation, just because the "revocation code" field contains an unexpected value. This seems like an unfortunate failure mode.

This PR implements parsing (and serialization) for any u8 value in the code field.
If there is a better way to implement this, I'd be happy about pointers, and to improve this PR.

### DIFF
--- a/src/packet/signature/de.rs
+++ b/src/packet/signature/de.rs
@@ -242,10 +242,7 @@ fn features(body: &[u8]) -> IResult<&[u8], SubpacketData> {
 /// Ref: https://tools.ietf.org/html/rfc4880.html#section-5.2.3.23
 fn rev_reason(i: &[u8]) -> IResult<&[u8], SubpacketData> {
     map(
-        pair(
-            map_res(be_u8, RevocationCode::try_from),
-            map(rest, BString::from),
-        ),
+        pair(map(be_u8, RevocationCode::from), map(rest, BString::from)),
         |(code, reason)| SubpacketData::RevocationReason(code, reason),
     )(i)
 }
--- a/src/packet/signature/de.rs
+++ b/src/packet/signature/de.rs
@@ -489,6 +489,8 @@ mod tests {
     #![allow(clippy::unwrap_used)]
 
     use super::*;
+    use crate::{Deserializable, StandaloneSignature};
+    use std::io::Cursor;
 
     #[test]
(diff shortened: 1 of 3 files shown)


dignifiedquire:
code looks great, this is the same pattern used for similar structures already 👍 would be great to have a test case for this as well

hko-s:
> code looks great, this is the same pattern used for similar structures already 👍 would be great to have a test case for this as well

I will definitely add a test.

---

A possible alternative approach:

In a branch that I'm still exploring, I've experimented with `num_enum::FromPrimitive` (see https://github.com/rpgp/rpgp/commit/d3639bc111448604c72825d8f1723ff6d2996b8e) that allows this kind of catch_all matching, and doesn't require two match-functions:

```
    #[num_enum(catch_all)]
    Other(u8)
```

If adding [num_enum](https://crates.io/crates/num_enum) is an option for `rpgp`, I could adjust this PR to use it.
(However, besides adding dependencies, `num_enum` also raises the minimal Rust version to 1.67)

Let me know what you think?

dignifiedquire:
I think in principle that would be fine, i wonder if `num_enum` would allow us to replace the other usages of `num_derive` given their similarities

hko-s:
> I think in principle that would be fine, i wonder if `num_enum` would allow us to replace the other usages of `num_derive` given their similarities

I'd assume that's possible. If it sounds like a good direction to you, I'm totally into replacing `num_derive` with `num_enum`, and would happy to do it. Probably it would be good to first do a mechanical replacement PR, then, before also changing business logic?

dignifiedquire:
> Probably it would be good to first do a mechanical replacement PR, then, before also changing business logic?

sounds good

hko-s:
See #268 for switch to `num_enum`

hko-s:
I've changed the PR to what I think is a good shape, based on `num_enum` (and manually verified that the change does what it should).

I'll now look into making a unit test for this.

dignifiedquire (approved):

PR #267, opened 2024-01-20
make dearmoring more robust

hko-s:
Allow spaces in the blank line(s) after the armor header.


dignifiedquire:
can you add a test for this please?

hko-s:
As an aside, while looking at this I was wondering if the choice of `many0` was intentional.

https://www.rfc-editor.org/rfc/rfc4880.html#section-6.2 says:

> A blank (zero-length, or containing only whitespace) line

So it seems that RFC 4880 expects *exactly* one such line. I have no strong feelings about this, but I wanted to raise this point here.

dignifiedquire:
yes it was intentional, if I remember correctly, there were just too many instances where the rfc was not followed in files that I found

PR #268, opened 2024-01-22
switch from num-derive to num_enum

hko-s:
This change is in preparation for using additional features in `num_enum`, in particular `#[num_enum(catch_all)]`.


dignifiedquire (approved):
nice

PR #270, opened 2024-01-24
fix: return an error in Message::verify if we don't know what to verify

hko-s:
This change makes it harder for users of the API to mistakenly think they have successfully validated a message, when in fact they have used the API wrong.

### DIFF
--- a/src/composed/message/types.rs
+++ b/src/composed/message/types.rs
@@ -413,9 +413,10 @@ impl Message {
                 let msg = Message::from_bytes(data.decompress()?)?;
                 msg.verify(key)
             }
-            // Nothing to do for others.
-            // TODO: should this return an error?
-            _ => Ok(()),
+            // We don't know how to verify a signature for other Message types, and shouldn't return Ok
+            _ => Err(Error::Unsupported(format!(
+                "Unexpected message format: {self:?}",
+            ))),
         }
     }
 


src/composed/message/types.rs
```diff
@@ -413,9 +413,8 @@ impl Message {
                 let msg = Message::from_bytes(data.decompress()?)?;
                 msg.verify(key)
             }
-            // Nothing to do for others.
-            // TODO: should this return an error?
-            _ => Ok(()),
+            // We don't know how to verify a signature for other Message types, and shouldn't return Ok
+            _ => Err(Error::Unsupported("Unexpected message format".to_string())),
```
dignifiedquire, line 417: lets add the message type to the error

hko-s: Good point. I added `Debug` output for the `Message` (which might be long, but I couldn't find anything better to print?)

dignifiedquire (approved):

src/composed/message/types.rs
dignifiedquire: good enough for now, though we should probably add a display impl for these that is a bit more usable

dignifiedquire: (in a follow up/later I mean)

hko-s: sounds good

Issue #269, opened 2024-01-24
Fail on truly unexpected packets, but recognize types of packets that rpgp can safely ignore


hko-s:
Currently, the packet parsing machinery opts to accept all kinds of unexpected packets by just ignoring them. This is good for not crashing, but comes with a risk of ignoring packets that would be important to handle, and are dangerous to ignore.

Ideally, I think, unknown packets should cause parsing to stop and fail. Except in explicitly good-listed cases.
I'd propose to use `Error::Unsupported` as a mechanism for the packet parsing machinery to signal packets that it saw,  doesn't handle but understood what they are, and knows they are safe to ignore.
This could, for example, be the case for a (sub)key packet that is older or newer than what rpgp supports. Or for a signature packet that uses an algorithm that rpgp doesn't support.

All other, truly unexpected, types of packets should lead to an error, and abort the parsing process.

I've done a rough sketch of such a change, and I think it's promising.
However, this is a pretty invasive change, and I wanted to get feedback on the general direction first.

dignifiedquire:
I agree the handling should be improved here, this was mainly done in early stages to make sure we don't reject sth just because we don't have support for it yet. Now that general support is much better we should be more strict about this.

hko-s:
Great, then I'll clean up my experiments and prepare a concrete suggestion that we can discuss and test.

hko-s:
This is done and has been merged via #271.

PR #271, opened 2024-01-24
Draft: Bad packet handling

hko-s:
This is an outline for an approach to #269.

It's still incomplete, and needs polish, but shows the direction I had in mind. I'd appreciate thoughts.


src/composed/shared.rs
```diff
@@ -78,12 +79,21 @@ pub trait Deserializable: Sized {
...
+                        let err: &Error = b; // unbox
+                        if let Error::Unsupported(_e) = err {
+                            // Error::Unsupported signals packet versions etc that we explicitly detected and can safely ignore
+                            warn!("skipping unsupported packet: {:?}", p);
+                            return None;
+                        }
+                    }
+
+                    // Fail on packets that we failed to parse for unspecific reasons
+                    panic!("unexpected packet data: {p:?}");
```
dignifiedquire, line 95: we really shouldn't panic by design, especially not on content controlled by an external party

dignifiedquire:
general approach looks good to me, main thing is to remove any `panic` calls from my side

hko-s:
> general approach looks good to me, main thing is to remove any `panic` calls from my side

Great, thanks. I'll cover some more cases that should also go into the "Unknown but ok" bucket, and will clean up panics after.

src/crypto/ecc_curve.rs
```diff
@@ -1,6 +1,7 @@
 use crate::crypto::public_key::PublicKeyAlgorithm;
+use asn1_rs::FromDer;
```
dignifiedquire, line 2: I'd prefer to use https://docs.rs/der/ from rustcrypto

hko-s: I pushed a variant that uses the `der` crate. However, that crate seems to produce a differing String format, and I don't have a good enough grasp of OIDs and ASN.1 to make sense of what's going wrong there:

```
test crypto::ecc_curve::tests::test_unknown_ecc_curve ... FAILED

failures:

---- crypto::ecc_curve::tests::test_unknown_ecc_curve stdout ----
thread 'crypto::ecc_curve::tests::test_unknown_ecc_curve' panicked at 'assertion failed: `(left == right)`

Diff < left / right > :
<0.6.9.43.36.3.3.2.8.1.1.1
>1.3.36.3.3.2.8.1.1.1

', src/crypto/ecc_curve.rs:214:9
```

src/crypto/ecc_curve.rs
```diff
@@ -11,12 +12,13 @@ pub enum ECCCurve {
     BrainpoolP384r1,
     BrainpoolP512r1,
     Secp256k1,
+    Unknown(String),
```
dignifiedquire, line 14: I would prefer to store the parsed Oid here and not the string

hko-s: I'd also prefer that, but couldn't think of a way to implement `ECCCurve::oid_str` (which returns a `&str`), then. Maybe the API of `ECCCurve` needs some more adjustment?

hko-s: (I did already incompatibly change that part of the public API anyway, by removing the `Copy` derive, anyway. And changed some functions to take a reference instead of an owned `ECCCurve`)

dignifiedquire: returning String is fine

src/crypto/ecc_curve.rs
```diff
@@ -191,6 +203,17 @@ mod tests {
...
 
+    #[test]
+    fn test_unknown_ecc_curve() {
+        // brainpoolP160r1 (an OID we don't handle explicitly in ECCCurve)
+        const OID: &[u8] = &[
+            0x06, 0x09, 0x2B, 0x24, 0x03, 0x03, 0x02, 0x08, 0x01, 0x01, 0x01,
+        ];
+
+        let bp_p160r1 = ecc_curve_from_oid(OID).unwrap();
+        assert_eq!(bp_p160r1.oid_str(), "1.3.36.3.3.2.8.1.1.1");
```
tarcieri, line 214: This test seems to be misinterpreting a DER header as part of the OID.

I'm not sure how `ecc_curve_from_oid` is defined, but it seems like you are using the parser for an OID body, but giving it a DER-encoded OID instead of just the body, i.e. with a header containing the leading `0x06` = OID tag, and `0x09` length octets.

If you want to parse a DER encoded OID, you need to use `ObjectIdentifier::from_der`. Otherwise, strip the leading first two bytes since they contain a DER header.

hko-s: Oh. I see, thank you for the explanation!

hko-s: I've dropped the DER header from the test-OID, and of course that fixes the issue.

src/crypto/ecc_curve.rs
hko-s: I've adjusted the change to `ECCCurve` accordingly.

src/crypto/ecc_curve.rs
hko-s: This PR now uses the `der` crate for OID handling.

The test issue is resolved, based on Tony's input, below (my test was wrong).

tarcieri: Also if you just need OID support, there's the `const-oid` crate

hko-s: Thanks! For the specific use case in this PR, `const-oid` is sufficient.
I'll switch to that, for now, until more features are required in rpgp.

src/composed/shared.rs
hko-s: Fully agreed, of course.

I've commited an update to this draft with explicit error handling instead of those (placeholder) panics.
My current version of an attempt at orderly error handling logic is in 54e4c162c30987d398a6f16caf36c447a89b0063

hko-s:
CI complains about code in `src/armor/reader.rs` (see https://github.com/rpgp/rpgp/actions/runs/7702519672/job/20990984070?pr=271):

```
error: read amount is not handled
   --> src/armor/reader.rs:391:13
    |
391 |             base_decoder.read(into)?
    |             ^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: use `Read::read_exact` instead, or handle partial reads
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unused_io_amount
```

Looks like a regression in clippy. Probably this one: https://github.com/rust-lang/rust-clippy/issues/12208

src/composed/message/parser.rs
```diff
@@ -8,13 +8,16 @@ use crate::errors::Result;
...
+pub struct MessageParser<I: Sized + Iterator<Item = Result<Packet>>> {
     source: Peekable<I>,
 }
 
-fn next<I: Iterator<Item = Packet>>(packets: &mut Peekable<I>) -> Option<Result<Message>> {
+fn next<I: Iterator<Item = Result<Packet>>>(packets: &mut Peekable<I>) -> Option<Result<Message>> {
     while let Some(packet) = packets.by_ref().next() {
-        // for packet in packets.by_ref() {
+        let Ok(packet) = packet else {
+            return Some(Err(packet.expect_err("must be err")));
```
dignifiedquire, line 18: rather than using `expect` just have a match statement

```rust
let packet = match packet {
  Ok(packet) => packet,
  Err(err) => return Some(Err(err)),
};
```

src/composed/message/parser.rs
```diff
@@ -39,21 +42,33 @@ fn next<I: Iterator<Item = Packet>>(packets: &mut Peekable<I>) -> Option<Result<
...
-                        while packets.peek().map(|p| {
-                            p.tag() == Tag::SymEncryptedData
-                                || p.tag() == Tag::SymEncryptedProtectedData
+                        // while edata take em (FIXME: the message grammar only allows one "Encrypted Data" packet)
+                        while packets.peek().map(|p| match p.as_ref() {
+                            Ok(p) => {
+                                p.tag() == Tag::SymEncryptedData
+                                    || p.tag() == Tag::SymEncryptedProtectedData
+                            }
+                            Err(_) => false,
```
dignifiedquire, line 65: maybe write a helper function 

```rust
fn check_packet(f: F) -> bool 
where F: Fn(&Packet) -> bool {
  match packet {
    Ok(p) => f(p),
    Err(_) => false,
  }
}
```

src/composed/shared.rs
```diff
@@ -77,21 +78,42 @@ pub trait Deserializable: Sized {
...
+            };
+
+            match &p {
+                Ok(_) => Some(p),
+                Err(e) => {
+                    if let InvalidPacketContent(b) = e {
+                        let err: &Error = b; // unbox
+                        if let Error::Unsupported(_e) = err {
+                            // Error::Unsupported signals packet versions etc that we explicitly detected and can safely ignore
+                            warn!("skipping unsupported packet: {:?}", p);
```
dignifiedquire, line 97: should probably print the error as well, otherwise debugging will be hard

dignifiedquire: as a `debug` though

src/composed/signed_key/key_parser_macros.rs
```diff
@@ -39,14 +39,25 @@ macro_rules! key_parser {
...
+                while packets.peek().map(|p| {
+                    match p.as_ref() {
+                        Ok(p) => p.tag() != $key_tag,
+                        Err(_) => false,
+                    }
+                }) == Some(true) {
+                    match packets.next().expect("peeked") {
+                        Ok(p) => {
+                            warn!("ignoring unexpected packet: expected {:?}, got {:?}", $key_tag, p.tag());
+                            // FIXME: return error?
```
dignifiedquire, line 51: probably should have the same logic re unsupported here?

src/composed/signed_key/key_parser_macros.rs
```diff
@@ -83,10 +103,19 @@ macro_rules! key_parser {
...
+                            Ok(packet) => {
+                                debug!("peek {:?}", packet.tag());
+                                 packet.tag() == Tag::UserId || packet.tag() == Tag::UserAttribute
+                            }
+                            Err(_) => false
+                        }
                     }) == Some(true) {
                     let packet = packets.next().expect("peeked");
+                    let Ok(packet) = packet else {
+                        return Some(Err(packet.err().unwrap()));
```
dignifiedquire, line 116: match statement would be nicer

src/composed/signed_key/parse.rs
```diff
@@ -54,18 +54,38 @@ pub fn from_armor_many<'a, R: io::Read + io::Seek + 'a>(
...
                     None
                 }
-                Ok(p) => Some(p),
-                Err(_) => {
-                    // for now we are skipping any packets that we failed to parse
-                    warn!("skipping packet: {:?}", p);
-                    None
+                Ok(p) => Some(Ok(p)),
+                Err(e) => {
+                    if let Error::InvalidPacketContent(b) = &e {
```
dignifiedquire, line 74: maybe this could be a helper function, given it is used in multiple places already, and we will want to keep this logic consistent likely

src/packet/public_key_parser.rs
```diff
@@ -191,13 +196,13 @@ pub(crate) fn parse(
...
+    let (i, key_ver) = map(be_u8, KeyVersion::from)(i)?;
     let (i, key) = match &key_ver {
         &KeyVersion::V2 | &KeyVersion::V3 => old_public_key_parser(&key_ver)(i)?,
         &KeyVersion::V4 => new_public_key_parser(&key_ver)(i)?,
-        KeyVersion::V5 => {
-            return Err(nom::Err::Error(crate::errors::Error::ParsingError(
-                nom::error::ErrorKind::Switch,
+        KeyVersion::V5 | KeyVersion::Other(_) => {
+            return Err(nom::Err::Error(crate::errors::Error::Unsupported(
+                "Unsupported key version".to_string(),
```
dignifiedquire, line 205: add the key version to the error message

src/packet/secret_key_parser.rs
```diff
@@ -86,13 +86,13 @@ pub(crate) fn parse(
...
+    let (i, key_ver) = map(be_u8, KeyVersion::from)(i)?;
     let (i, key) = match &key_ver {
         &KeyVersion::V2 | &KeyVersion::V3 => old_private_key_parser(&key_ver)(i)?,
         &KeyVersion::V4 => new_private_key_parser(&key_ver)(i)?,
-        KeyVersion::V5 => {
-            return Err(nom::Err::Error(crate::errors::Error::ParsingError(
-                nom::error::ErrorKind::Switch,
+        KeyVersion::V5 | KeyVersion::Other(_) => {
+            return Err(nom::Err::Error(Error::Unsupported(
+                "Unsupported key version".to_string(),
```
dignifiedquire, line 95: add the version to the error

src/composed/signed_key/key_parser_macros.rs
hko-s: I think this `match` statement does what it should. My intention for this PR is that the low-level parser distinguishes three cases (instead of formerly just two):

1. Packets that it can parse and handle, and hand up to the meta-parser (we get those as `Ok`-wrapped packets, here)
2. Packets that the low level parser recognizes as safe, but `Unknown` will not be returned to the meta-parser (we don't see these packets here at all: the low-level parser skips over them, for us)
3. Packets that the low level parser can't explicitly recognize as "safe" will bubble up to the meta-parser as an error (we fail on these, here)

hko-s: (Before, no errors were ever bubbled up from the low-level parser to the meta-parsers, all "unknown" packets were skipped. So the third case didn't exist, and we didn't have the error branch here at all)

src/composed/message/parser.rs
hko-s: I think we're half re-inventing https://doc.rust-lang.org/std/result/enum.Result.html#method.is_ok_and (from Rust 1.70.0), there :sweat_smile: 

0ef10613701b3ae049c2b0c10a011b9a6c370433 is a bit less verbose, but not entirely satisfying. I keep wondering if I'm missing some additional simplification that can be done with `std`.

hko-s:
I've processed your feedback (note that I wrote two replies that are still open, we might want to continue discussing those two threads).

I've made one more substantial change to the logic of this RP: Two more error types are now considered "soft packet parser failures" (see 5491bda201f66e0bce8f3b134d940c6b5e7d97ca).
I think these errors happen with "weird packets" that can be safely skipped, and can't imagine a case where it's necessary to hard-error on them.

With these changes, CI is green again, after bumping the numbers of expected parsable packets (see 807bba718422f80f873145bf4e2f8eea2a07d916).
I have not verified which exact packets can now be parsed, but the increase in numbers feels plausible, and I don't know that it's worth investigating more deeply.
I was surprised that for `test_parse_dumps_0` the test outcome for `actual_count` increased by one, and wonder what might have caused this.

hko-s:
I also added an `Other`  value to `CompressionAlgorithm` and `SymmetricKeyAlgorithm`, following the same pattern

dignifiedquire:
any reason this is still in draft mode?

hko-s:
> any reason this is still in draft mode?

no good reason. i un-drafted it.

src/composed/message/parser.rs
dignifiedquire: sure

hko-s:
rebased, and adjusted the expected `test_parse_dumps` numbers

src/composed/message/parser.rs
```diff
@@ -106,21 +121,25 @@ fn next<I: Iterator<Item = Packet>>(packets: &mut Peekable<I>) -> Option<Result<
...
 
-                        let signature =
-                            if packets.peek().map(|p| p.tag() == Tag::Signature) == Some(true) {
-                                packets.next().expect("peeked").try_into().expect("peeked")
-                            } else {
-                                return Some(Err(format_err!(
-                                    "missing signature for, one pass signature"
-                                )));
-                            };
+                        let signature = if packets.peek().map_or(false, |res| {
```
dignifiedquire, line 131: I think all these `peek` calls could be simplified with this: https://doc.rust-lang.org/stable/std/iter/struct.Peekable.html#method.next_if not sure why I didn't use this back then.

src/composed/signed_key/key_parser_macros.rs
```diff
@@ -57,20 +68,29 @@ macro_rules! key_parser {
...
+                            debug!("parsing signature {:?}", packet.tag());
+                            let sig: Signature = err_opt!(packet.try_into());
+                            let typ = sig.typ();
+
+                            if typ == SignatureType::KeyRevocation {
+                                revocation_signatures.push(sig);
+                            } else {
+                                if primary_key.version() != KeyVersion::V4 {
+                                    // no direct signatures on V2|V3 keys
+                                    info!("WARNING: unexpected signature: {:?}", typ);
```
dignifiedquire, line 88: this really should be `warn` or not say `WARNING` 🤣

src/composed/signed_key/key_parser_macros.rs
```diff
@@ -137,26 +182,40 @@ macro_rules! key_parser {
...
+                    }
                 }) == Some(true) {
                     // -- Only V4 keys should have sub keys
                     if primary_key.version() != KeyVersion::V4 {
                         return Some(Err(format_err!("only V4 keys can have subkeys")));
                     }
 
                     let packet = packets.next().expect("peeked");
+                    let Ok(packet) = packet else {
+                        return Some(Err(packet.err().unwrap()));
```
dignifiedquire, line 200: `match` to avoid the `unwrap`?

dignifiedquire (approved):
some last notes, but overall looks good, thanks for working through this

src/composed/message/parser.rs
hko-s: are you voting to bump rust to 1.70 right away, or to change the code later?

dignifiedquire: lets do it now

hko-s:
> some last notes, but overall looks good, thanks for working through this

Ok, I think I've done these bits of cleanup. Thanks for talking me through this :)

dignifiedquire:
thanks alot for working though this

PR #274, opened 2024-01-25
fix: handle expiration times as Duration

hko-s:


dignifiedquire:
why?

hko-s:
The values of the two expiration time subpackets are defined as duration in seconds, not as absolute dates. Their names are misleading, but the text is clear. e.g., from 4880:

> The validity period of the key.  This is the number of seconds after the key creation time that the key expires.

dignifiedquire:
oh, I must have missed that 😅 thanks

src/packet/signature/de.rs
```diff
@@ -38,6 +38,11 @@ fn dt_from_timestamp(ts: u32) -> Option<DateTime<Utc>> {
         .map(|ts| DateTime::<Utc>::from_naive_utc_and_offset(ts, Utc))
 }
 
+/// Convert a u32 to a `Duration`
+fn duration_from_timestamp(ts: u32) -> Option<Duration> {
+    Some(Duration::seconds(ts as i64))
```
dignifiedquire, line 43: please use `i64::from` instead

hko-s:
> oh, I must have missed that 😅 thanks

The naming is utterly confusing

src/packet/signature/de.rs
hko-s: i changed it

dignifiedquire (approved):

PR #279, opened 2024-02-02
implement From<&SignedSecretKey> for SignedPublicKey

hko-s:
Does some version of this functionality exist already?


src/composed/signed_key/secret.rs
```diff
@@ -315,3 +315,23 @@ impl PublicKeyTrait for SignedSecretSubKey {
         self.key.to_writer_old(writer)
     }
 }
+
+impl From<&SignedSecretKey> for SignedPublicKey {
```
dignifiedquire, line 319: I would prefer doing this on the non ref version and making the clone explicit on the caller side

dignifiedquire:
I am surprised sth like this doesn’t have exist

dignifiedquire:
wait https://github.com/rpgp/rpgp/blob/b11ed268d9d1ea38dab56977a94408b787d0dab9/src/composed/signed_key/secret.rs#L180 isn’t this the same?

hko-s:
> wait
> 
> https://github.com/rpgp/rpgp/blob/b11ed268d9d1ea38dab56977a94408b787d0dab9/src/composed/signed_key/secret.rs#L180
> isn’t this the same?

I think that function drops the signatures. It produces only a `PublicKey`, not a `SignedPublicKey`.

dignifiedquire:
right, that's probably why this never existed

src/composed/signed_key/secret.rs
```diff
@@ -315,3 +315,23 @@ impl PublicKeyTrait for SignedSecretSubKey {
...
+        let public_primary = value.primary_key.public_key();
+        let details = value.details;
+
+        let mut public_subkeys = value.public_subkeys;
+
+        for subkey in value.secret_subkeys {
+            let pk = subkey.key.public_key();
+            let sig = subkey.signatures;
+
+            let spsk = SignedPublicSubKey::new(pk, sig);
```
dignifiedquire, line 330: maybe the `From` should exist for subkeys as well? cleaning this code up a bit

eg 

```rust
let public_subkeys: Vec<SignedPublicSubKey> = value.secret_subkeys.into_iter().map(Into::into).collect();
```

hko-s: Good idea. Will do.

hko-s: I've adjusted the code (it's a bit less "neat", because two formerly separate sets of subkeys need to be joined together into one combined `Vec`).

dignifiedquire (approved):

dignifiedquire:
sth just broke all of ci

hko-s:
> sth just broke all of ci

I can reproduce those errors locally when I run `cargo +nightly clippy --tests`.

That was ok with nightly from `5518eaa94 2024-01-29`, but isn't ok after update to `f067fd608 2024-02-05`.

(It's currently unclear to me if clippy has a point, or if its complaints are baseless.)

hko-s:
It's apparently a reasonable complaint. Looks like `curve25519-dalek` will get a fix for it: https://github.com/dalek-cryptography/curve25519-dalek/issues/618

I don't know what we should do until such a release is made, though?

dignifiedquire:
lets just pin to an older working nightly for now until the fix is out

hko-s:
> lets just pin to an older working nightly for now until the fix is out

Right. Let's do that in #283

dignifiedquire:
needs rebase onto fixed `master`

hko-s:
> needs rebase onto fixed `master`

done

dignifiedquire (approved):

PR #282, opened 2024-02-03
make low level ESK decryption functions public

hko-s:
For users of rPGP to directly handle decrypt pkesk and skesk, `decrypt_session_key` and `decrypt_session_key_with_password` are required.

I think it makes sense to make these two functions public: to my mind they are roughly analogous to `Signature::verify`.

One way to make them public is to make all of `pgp::composed::message::decrypt` public (that's what this PR proposes).

The only other thing in this module is the type `MessageDecrypter`, which is already kind of part of rPGP's public API as a return type of `Message::decrypt` (I was surprised that it's even legal that `MessageDecrypter` is currently private O_o)


dignifiedquire:
lets explicitly export them and not make the whole thing public

hko-s:
> lets explicitly export them and not make the whole thing public

I changed the PR to be similar to other instances of `pub use` in rpgp. Is that what you had in mind?

I'm a little unsure about the namespacing - this re-export makes the `decrypt` functions appear one module level up from where they are actually implemented.

dignifiedquire (approved):

dignifiedquire:
fine for now, longterm the whole public api needs probably some reworking

PR #283, opened 2024-02-06
chore: use a variable for "nightly", use a fixed nightly version for now

hko-s:
This nightly version avoids triggering the errors from https://github.com/dalek-cryptography/curve25519-dalek/issues/618

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest]
-        rust: [$RUST_NIGHTLY, beta, stable]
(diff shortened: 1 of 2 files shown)


hko-s:
Let's see if I can figure out how to do this in github CI :sweat_smile:

dignifiedquire (approved):

dignifiedquire:
I think needs some more github actions foo 🤣

hko-s:
> I think needs some more github actions foo 🤣

If you have a rough idea how to do this, I'd be happy to hand this over to you to fix?
I'm totally out of my element here, and might take a few hundred iterations :scream_cat:

dignifiedquire:
I‘ll take a look tomorrow

PR #287, opened 2024-02-13
fix: do not allow decryption with "Plaintext" algorithm

link2xt:
According to
<https://datatracker.ietf.org/doc/html/rfc4880#section-13.4> "plaintext" MUST NOT be used to encrypt data packets.

We check that "plaintext" algorithm is not used
to decrypt messages and is not used to decrypt session keys
when the key is derived using S2K.

### DIFF
--- a/src/composed/message/types.rs
+++ b/src/composed/message/types.rs
@@ -270,7 +270,7 @@ impl Message {
         self.encrypt_symmetric(rng, esk, alg, session_key)
     }
 
-    /// Encrytp the message using the given password.
+    /// Encrypt the message using the given password.
     pub fn encrypt_with_password<R, F>(
         &self,
         rng: &mut R,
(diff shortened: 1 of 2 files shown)


dignifiedquire:
I wish we had a test for this

dignifiedquire:
thanks :

src/composed/message/types.rs
```diff
@@ -803,6 +818,50 @@ mod tests {
...
+                data += hashlib.new("SHA1", inner_data + b"\xd3\x14").digest()
+                print(data)
+        */
+
+        let msg = Message::from_bytes(&msg_raw[..]).unwrap();
+
+        // Before the fix message eventually decrypted to
+        //   Literal(LiteralData { packet_version: New, mode: Binary, created: 1970-01-01T00:00:00Z, file_name: "", data: "48656c6c6f20776f726c6421" })
+        // where "48656c6c6f20776f726c6421" is an encoded "Hello world!" string.
+        assert!(msg.decrypt_with_password(|| "foobarbaz".into()).is_err());
```
dignifiedquire, line 862: maybe check the correct error is returned, at least on a string basis?

dignifiedquire (approved):

PR #290, opened 2024-02-19
Do not allow recursive decompression

link2xt:


dignifiedquire (approved):

dignifiedquire:
I don't see the need for even two layers, so I think this is a good approach.

PR #294, opened 2024-02-20
fix: enforce Partial Body Length "MUST" from RFC 4880

hko-s:


dignifiedquire:
any chance we can get a small test for this?

hko-s:
> any chance we can get a small test for this?

Good point. I added three test packet files, each containing the same `LiteralData` information (but one of them in an illegal format):

- One with regular two byte length encoding,
- one with a legal first partial length part of 512 bytes, followed by a fixed length part, and 
- one with an illegal first partial length part of 256 bytes, followed by a fixed length part.

Without this PR, all three files get parsed into the same literal data. With this PR, as specified in 4880, the 256 byte partial part throws an error.

dignifiedquire (approved):

dignifiedquire:
thanks

PR #296, opened 2024-02-21
Fix CI status image

wiktor-k:
I've noticed that the image shows "build failing" but the CI succeeds and works (as evidenced by clicking the link) so I went to Actions and pressed "Create status badge" and replaced the link below with what was suggested by GitHub.

Edit: I've noticed the Rust version badge was also out of date so I've fixed that too to reflect what is written as text.


dignifiedquire (approved):
thanks

PR #307, opened 2024-03-06
feat: obfuscate symmetric key length with ecdh padding

### LINKED ISSUE #292: When generating ECDH padding: Pad to "largest expected size"
Also see https://github.com/rpgp/rpgp/pull/280

The padding mechanism described in RFC 6637 "allows the sender to obfuscate the size of the symmetric encryption key used  to encrypt the data". We can and should pad to the biggest expected size.

It seems this size is 40 (based on AES 256 as the longest type of symmetric key we use).

We should adjust the padding that rpgp emits.

hko-s:
Padding in https://www.rfc-editor.org/rfc/rfc6637 is described as being based on PKCS5. However, RFC 6337 allows padding that exceeds the size of one block, with the goal of obfuscating the size of the symmetric key material.

This commit changes the default size of padded payloads to 40 bytes, which is the regular size of a padded payload that contains a AES256 key.

fixes #292

### DIFF
--- a/src/crypto/ecdh.rs
+++ b/src/crypto/ecdh.rs
@@ -1,5 +1,3 @@
-use block_padding::{Padding, Pkcs7};
-use generic_array::{typenum::U8, GenericArray};
 use rand::{CryptoRng, Rng};
 use x25519_dalek::{PublicKey, StaticSecret};
 use zeroize::{Zeroize, Zeroizing};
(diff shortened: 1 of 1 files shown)


src/crypto/ecdh.rs
```diff
@@ -252,25 +250,41 @@ pub fn encrypt<R: CryptoRng + Rng>(
...
+    let padded_len = if len < PAD_DEFAULT_TARGET {
+        // Normally, we just pad to the default target size ...
+        PAD_DEFAULT_TARGET
+    } else {
+        // ... but if `plain` isn't shorter than our target size, we pad to the next full block
+        let remainder = len % 8; // e.g. 3 for len==19
+
+        len + 8 - remainder // (e.g. "8 + 8 - 0 => 16", or "19 + 8 - 3 => 24")
     };
+    ensure!(padded_len % 8 == 0, "Unexpected padded_len {}", padded_len);
```
dignifiedquire, line 271: maybe this should be a `debug_assert`?

src/crypto/ecdh.rs
```diff
@@ -252,25 +250,41 @@ pub fn encrypt<R: CryptoRng + Rng>(
...
+
+    // The value we'll use for padding (must be between 1 and 255)
+    let padding = padded_len - len;
+    ensure!(
+        padding > 0 && padding < 256,
+        "Unexpected padding value {}",
+        padding
+    );
```
dignifiedquire, line 279: same here

src/crypto/ecdh.rs
```diff
@@ -252,25 +250,41 @@ pub fn encrypt<R: CryptoRng + Rng>(
...
+        let remainder = len % 8; // e.g. 3 for len==19
+
+        len + 8 - remainder // (e.g. "8 + 8 - 0 => 16", or "19 + 8 - 3 => 24")
     };
+    ensure!(padded_len % 8 == 0, "Unexpected padded_len {}", padded_len);
+
+    // The value we'll use for padding (must be between 1 and 255)
+    let padding = padded_len - len;
+    ensure!(
+        padding > 0 && padding < 256,
```
dignifiedquire, line 276: you could use `u8::try_from(padding)` for the size check instead

src/crypto/ecdh.rs
hko-s: right, sounds good

src/crypto/ecdh.rs
hko-s: i also want >0, so `u8::try_from` is not exactly shorter - but i agree that it's more expressive. changed.

PR #309, opened 2024-03-07
s2k: restructure, implement argon2

hko-s:


dignifiedquire:
do you have a test case for a full message using argon2 that we could use for an e2e test?

hko-s:
> do you have a test case for a full message using argon2 that we could use for an e2e test?

yes, we can use the tests from: https://www.ietf.org/archive/id/draft-ietf-openpgp-crypto-refresh-13.html#name-sample-messages-encrypted-u (i extracted the unit test vectors from those)

i'll add a message decryption test.

dignifiedquire (approved):

PR #310, opened 2024-03-07
chore: drop workaround, should be resolved

hko-s:


hko-s:
while this seems to be resolved in nightly, it's still currently broken in other toolchains

PR #311, opened 2024-03-07
Detect armor

hko-s:
A possible approach to autodetecting armored vs. binary OpenPGP streams.

Whatever approach we take, we should implement it for all types of data: keys, messages and raw signatures.
But I wanted to solicit thoughts, first.

### DIFF
--- a/src/composed/shared.rs
+++ b/src/composed/shared.rs
@@ -1,5 +1,6 @@
+use byteorder::ReadBytesExt;
 use std::collections::BTreeMap;
-use std::io::{Cursor, Read, Seek};
+use std::io::{Cursor, Read, Seek, SeekFrom};
 
 use crate::armor::{self, BlockType};
 use crate::errors::{Error, Result};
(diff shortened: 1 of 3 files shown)


src/composed/signed_key/parse.rs
```diff
@@ -9,49 +11,64 @@ use crate::errors::{Error, Result};
...
+                unimplemented_err!("key format {:?}", typ);
+            }
         }
+    } else {
+        Ok((from_bytes_many(input), None))
     }
 }
 
 /// Parses a list of secret and public keys from raw bytes.
-pub fn from_bytes_many<'a>(
```
dignifiedquire, line 54: I think in any case we should still keep explicit versions public, such that places where the format is already known can call those

hko-s: Right, agreed

hko-s:
Shifted to a very minimalist approach, which I think covers this one goal well enough.

src/composed/shared.rs
```diff
@@ -86,6 +87,41 @@ pub trait Deserializable: Sized {
...
     ) -> Box<dyn Iterator<Item = Result<Self>> + 'a>;
+
+    /// Parses a single composition, from either ASCII-armored or binary OpenPGP data.
+    ///
+    /// Returns a composition and a BTreeMap containing armor headers
+    /// (None, if the data was unarmored)
+    #[cfg_attr(feature = "cargo-clippy", allow(clippy::type_complexity))]
+    fn from_reader_single<'a, R: Read + Seek + 'a>(
+        mut input: R,
+    ) -> Result<(Self, Option<BTreeMap<String, String>>)> {
```
dignifiedquire, line 98: note for the future, we should really make a new type for the armor headers

src/composed/shared.rs
```diff
@@ -134,3 +170,17 @@ pub(crate) fn filter_parsed_packet_results(p: Result<Packet>) -> Option<Result<P
         }
     }
 }
+
+/// Check if the OpenPGP data in `input` seems to be ASCII-armored or binary (by looking at the
+/// highest bit of the first byte)
+pub(crate) fn is_binary<R: Read + Seek>(input: &mut R) -> Result<bool> {
+    // Peek at the first byte in the reader
+    let first = input.read_u8()?;
+    let _ = input.seek(SeekFrom::Start(0))?;
```
dignifiedquire, line 179: This is not correct, we don't know if the source is at `0`. We should use `stream_position` before reading to store the current position

dignifiedquire: or just use `SeekFrom::Current(-1)`

src/composed/shared.rs
hko-s: right, I opened #313 to track

dignifiedquire (approved):

src/composed/shared.rs
hko-s: good catch. i changed this to the relative-seek approach.

PR #314, opened 2024-03-19
Refactor code to allow using hardware tokens

wiktor-k:
This is super early PR refactoring of the decryption API to give it a bit more flexibility. 

I've tested it with my yubikey through this example project: https://codeberg.org/wiktor/openpgp-card-rpgp/commit/b3d791d5decbcccf9be4ed7aed657e40d81cc8b5 and it worked:)

This PR has numerous issues now:
- since the unlock API is also used in signing I've mocked that out
- unlock is also used in tests to check the parameters of the keys, I've commented it out for the time being
- the naming could be improved and a bit of code shuffling is still necessary 
- quite a lot of code still uses SecretKeyRepr which makes that code usable only with software keys
- the code changes need to be split into multiple commits

Well, I guess that's it. Thanks for your time! 👋

### DIFF
--- a/src/composed/message/decrypt.rs
+++ b/src/composed/message/decrypt.rs
@@ -1,8 +1,8 @@
 use crate::crypto::sym::SymmetricKeyAlgorithm;
-use crate::crypto::{checksum, ecdh, rsa};
+
 use crate::errors::Result;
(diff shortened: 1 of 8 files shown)


dignifiedquire:
Merged in parts & ideas now

PR #315, opened 2024-03-19
[WIP] Prepare for hardware support

dignifiedquire:
Based on #314 

Starts to prepare more generic abstractions of the crypto backends to allow for supporting hardware tokens.

### DIFF
--- a/src/composed/message/decrypt.rs
+++ b/src/composed/message/decrypt.rs
@@ -1,62 +1,26 @@
 use crate::crypto::sym::SymmetricKeyAlgorithm;
-use crate::crypto::{checksum, ecdh, rsa};
 use crate::errors::Result;
 use crate::packet::SymKeyEncryptedSessionKey;
 use crate::types::{KeyTrait, Mpi, SecretKeyRepr, SecretKeyTrait};
 
 /// Decrypts session key using secret key.
-pub fn decrypt_session_key<F>(
-    locked_key: &(impl SecretKeyTrait + KeyTrait),
+pub fn decrypt_session_key<F, L>(
+    locked_key: &L,
     key_pw: F,
     mpis: &[Mpi],
 ) -> Result<(Vec<u8>, SymmetricKeyAlgorithm)>
 where
     F: FnOnce() -> String,
+    L: SecretKeyTrait<Unlocked = SecretKeyRepr> + KeyTrait,
 {
     debug!("decrypting session key");
 
     let mut key: Vec<u8> = Vec::new();
     let mut alg: Option<SymmetricKeyAlgorithm> = None;
     locked_key.unlock(key_pw, |priv_key| {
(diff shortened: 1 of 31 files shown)


src/composed/message/decrypt.rs
```diff
@@ -1,62 +1,26 @@
...
-pub fn decrypt_session_key<F>(
-    locked_key: &(impl SecretKeyTrait + KeyTrait),
+pub fn decrypt_session_key<F, L>(
+    locked_key: &L,
     key_pw: F,
     mpis: &[Mpi],
 ) -> Result<(Vec<u8>, SymmetricKeyAlgorithm)>
 where
     F: FnOnce() -> String,
+    L: SecretKeyTrait<Unlocked = SecretKeyRepr> + KeyTrait,
```
wiktor-k, line 14: I'm using `decrypt_session_key` in here: https://codeberg.org/wiktor/openpgp-card-rpgp/src/branch/hack/examples/decrypt.rs#L27

But the `Unlocked` needing a `SecretKeyRepr` constraint here means the `decrypt_session_key` is still bound to software keys... I know this is a work in progress but maybe I'm holding it wrong somehow?

dignifiedquire: you are not holding it wrong, you can't use this yet for other implementations, I just added this as a placeholder until I have a constraint that I am happy with

wiktor-k: Ack. Thanks! :+1: 

I think in the ideal world the `SecretKeyRepr` would never be passed anywhere but some trait which `SecretKeyRepr` (or the Secret key packet?) implements would be used everywhere.

From what I can see there's just a couple of bits needed to pass around:
  - ECDH key params (even though in v6 they'd be fixed),
  - some metadata e.g. for signing one embeds the signing key fingerprint in the signature

wiktor-k:
By the way, I've made a small piece of code that could be used in unit tests:

This fragment uses fixed shared secret that I've captured from my tests: https://codeberg.org/wiktor/openpgp-card-rpgp/commit/40affaf19302978b1020be87111a411d7c4575bb#diff-b24749917179fb5e3e613ed2a703fcdcc6cdf9da (card code commented out)

And it successfully decrypts the sample message2.asc: https://codeberg.org/wiktor/openpgp-card-rpgp/commit/40affaf19302978b1020be87111a411d7c4575bb#diff-b5b530e096e7ef5a4f1c052a38d8585f96a77fb0 given that this tiny fixup code is applied: https://codeberg.org/wiktor/openpgp-card-rpgp/commit/288a63f38c2db77a43aafefa7717b1afaceb09d8

This is only for ed25519 so it'd be good to have a test case for RSA too but it's at least something.

What do you think about creating a PR against *this* PR to add the test case? I guess I could do that or @hko-s :)

PR #316, opened 2024-03-20
feat: basic support for AEAD

dignifiedquire:
The primitives for encryption are there, but this will be implemented once proper v5/v6 support has landed, as there are too many missing pieces.

- messages
  - [x] parsing
  - [x] writing
  - [ ] encrypting
  - [x] decrypting
- key params
  - [x] parsing
  - [x] writing
  - [ ] encrypting
  - [x] decrypting

- [x] ocb
  - [x] aes128
  - [x] aes256
- [x] eax
  - [x] aes128
  - [x] aes256
- [x] gcm
  - [x] aes128
  - [x] aes256


src/composed/message/types.rs
```diff
@@ -164,19 +164,88 @@ impl Edata {
...
+                    Self::SymEncryptedProtectedData(p) => {
+                        let decrypted_packet = p.decrypt(&key, None)?;
+
+                        // TODO: handle multiple messages (padding packets can appear here)
 
-        Message::from_bytes(Cursor::new(decrypted_packet.to_vec()))
+                        Message::from_bytes(Cursor::new(decrypted_packet))
+                    }
+                    Self::SymEncryptedData(_) => {
+                        // TODO: is this a valid case?
```
hko-s, line 238: I don't think this is a valid case:

> A v6 SKESK packet MUST NOT precede a v1 SEIPD packet or a deprecated Symmetrically Encrypted Data packet (see [Section 10.3.2.1](https://www.ietf.org/archive/id/draft-ietf-openpgp-crypto-refresh-13.html#encrypted-message-versions)).

(from https://www.ietf.org/archive/id/draft-ietf-openpgp-crypto-refresh-13.html#name-version-6-symmetric-key-enc)

So `SymEncryptedProtectedData` with version one must also be rejected here.

src/types/s2k.rs
```diff
@@ -5,13 +5,81 @@ use nom::combinator::{map, rest};
...
+    /// 0
+    Unprotected,
+    /// 1..253
+    LegacyCfb(SymmetricKeyAlgorithm),
+    /// 253
+    Aead,
+    /// 254
+    Cfb,
+    /// 255
+    MaleableCfb,
```
hko-s, line 31: Typo, should be `MalleableCfb`

PR #317, opened 2024-03-20
Fix allocation bug

LozzaTray:
The buffer allocator does not correctly respect the size increase. This PR aims to fix that by growing the buffer by the requested amount.

### DIFF
--- a/src/packet/many.rs
+++ b/src/packet/many.rs
@@ -143,12 +143,24 @@ impl<R: Read> Iterator for PacketParser<R> {
             }
 
             // if the parser returned `Incomplete`, and it needs more data than the buffer can hold, we grow the buffer.
-            if let Some(Needed::Size(sz)) = needed {
-                if b.usable_space() < sz.into() && self.capacity * 2 < MAX_CAPACITY {
-                    self.capacity *= 2;
-                    let capacity = self.capacity;
-                    b.make_room();
-                    b.reserve(capacity);
+            if let Some(needs) = needed {
+                match needs {
+                    Needed::Size(sz) => {
+                        let sz: usize = sz.into();
+                        if b.usable_space() < sz && self.capacity + sz < MAX_CAPACITY {
+                            self.capacity += sz;
+                            b.make_room();
+                            b.reserve(self.capacity);
+                        }
+                    }
+                    Needed::Unknown => {
+                        warn!("Doubling buffer size to accommodate unknown extra quantity of data");
+                        if self.capacity * 2 < MAX_CAPACITY {
+                            self.capacity *= 2;
+                            b.make_room();
+                            b.reserve(self.capacity);
+                        }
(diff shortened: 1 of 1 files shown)


dignifiedquire:
Thanks,  have implemented a variant of this in #319 

Did you run into an error with this or did you find this?

LozzaTray:
> Thanks, have implemented a variant of this in #319
> 
> Did you run into an error with this or did you find this?

First of all, thanks for taking a look at this PR. I did actually run into this error while trying to decrypt some data encrypted by this library. Interestingly, my issue was not with the `Needed::Unknown` codepath that your PR deals with but instead with the `Needed::Size(sz).

There are only a maximum of 2 rounds of buffer increase allowed (https://github.com/rpgp/rpgp/pull/317/files#diff-0c2121684730f47b19eac98dd7042f60a1376f48e9d5a285e5719fd500d9c562L63-L71) and each round just doubles the buffer size. So if `sz > self.capacity * 4`, the code will quit out before the buffer has grown enough to accomodate the packet.

I don't know if it's intentional that we always just double the buffer size rather than increasing it to fit `sz`. You certainly have more context on this codebase so would welcome your input. I've verified that the change in this PR successfully decrypts the data I was after but not sure if there is a more appropriate way.

dignifiedquire:
> I don't know if it's intentional that we always just double the buffer size rather than increasing it to fit `sz`.

Many small allocations tend to be expensive, so we optimistically double the size. The other thing is that the needed size is not always 100% accurate. 

Your PR pointed me to a couple of edge cases that weren't dealt with properly, which is why I believe your decoding now works. 



> I've verified that the change in this PR successfully decrypts the data I was after but not sure if there is a more appropriate way.
Glad to hear it

LozzaTray:
So the "this PR" I was referring to that fixed my problems was #317 rather than your #319. I haven't tested with #319 but I imagine I would still get the error as the `Needed::Size` behaviour is unchanged. Is there a reason to error after two doublings of the buffer size? Could we instead only error once `MAX_CAPACITY` is exceeded?

dignifiedquire:
it shouldn’t stop after two doublings anymore only at the max capacity, which is much larger than that

LozzaTray:
I don't think that's true. The `second_round` boolean gets set to `True` after the first pass of the loop that doubles the buffer and then we return `Some(Err(Error::PacketIncomplete))` after the second: https://github.com/rpgp/rpgp/blob/master/src/packet/many.rs#L63-L71

Happy to add a test-case that covers this by trying to decrypt a packet that is larger than 4x the initial capacity but still smaller than MAX_CAPACITY

dignifiedquire:
> Happy to add a test-case that covers this by trying to decrypt a packet that is larger than 4x the initial capacity but still smaller than MAX_CAPACITY

yes that would be great

PR #320, opened 2024-03-23
refactor: break out ecdh unwrap

hko-s:


src/crypto/ecdh.rs
```diff
@@ -91,70 +89,87 @@ impl Decryptor for SecretKey {
...
+            fingerprint,
+        )
+    }
+}
 
-            // Padding length seems to exceed size of the padded message
-            if *pad as usize > len {
-                return Err(Error::UnpadError);
-            }
+pub fn ecdh_decrypt_unwrap_unpad(
```
dignifiedquire, line 110: no need for the `ecdh` prefix, the module is already called `ecdh`

dignifiedquire: also the`unwrap` is a bit unfortunate, as in the rust context this means panic on error

dignifiedquire: also if this becomes public it should have a doc comment describing what exactly it does

hko-s: Good points. I've made adjustments.

I'm torn about my change to use `KeyParams`: It makes the function parameters a little more expressive, but adds a pointless `oid.clone()`. What do you think?

dignifiedquire: more expressive sounds good

dignifiedquire (approved):

PR #328, opened 2024-03-31
feat: cleartext framework support

### LINKED ISSUE #122: Cleartext signature framework support (gpg --clearsign)
Currently there does not appear to be support for the plaintext signature framework as described in [RFC4880 section 7](https://tools.ietf.org/html/rfc4880.html#section-7).

This style of signature involves a message which starts with a "`-----BEGIN PGP SIGNED MESSAGE-----`" header, followed by an inline, non-base64-encoded cleartext message body, followed by a regular ASCII armored PGP signature.

The GPG command line tool produces such signatures when invoked with `--clearsign`.

dignifiedquire:
thanks to @hko-s for tips and existing work.

- Introduces `composed::CleartextSignedMessage`
- Introduces `composed::Any` for dealing with arbitrary armor based content

Closes #122


src/composed/any.rs
```diff
@@ -0,0 +1,68 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Any {
+    Cleartext(CleartextSignedMessage),
+    PublicKey(SignedPublicKey),
+    SecretKey(SignedSecretKey),
+    Message(Message),
+    Signature(StandaloneSignature),
+}
+
+impl Any {
```
hko-s, line 22: This name clashes with `std::any::Any`. I wonder if the name is worth changing.

I fear the RFC doesn't offer terminology for "any kind of PGP thing". The most generic objects it names are "Messages" and "Keys" (neither of which includes cleartext messages).

Attempting to brainstorm alternative names: `(Open)PgpData`, `(Open)PgpContent`.

I currently have no idea that I'm excited about.

(I don't think this kind of generic reader needs to be limited to armored data, so I would tend to not use "armor" in the naming)

dignifiedquire: I think I am going to stick with `Any` for now, and wait for better suggestions :)

PR #330, opened 2024-04-02
csf: add test

hko-s:


src/cleartext.rs
```diff
@@ -128,46 +132,46 @@ impl CleartextSignedMessage {
...
+    /// The output is normalized to "\r\n" line endings.
+    ///
+    /// Returns `None` if the text isn't valid UTF8
+    pub fn signed_text(&self) -> Option<String> {
+        let unescaped = dash_unescape(&self.csf_encoded_text);
+
+        let normalized: Vec<u8> = Normalized::new(unescaped.bytes(), LineBreak::Crlf).collect();
 
-        out
+        std::str::from_utf8(&normalized).map(str::to_owned).ok()
```
dignifiedquire, line 169: you can use `.expect` here, it is guranteed that this is valid utf8, because `csf_encoded_text` already is

hko-s: great, done

dignifiedquire (approved):
thanks

PR #337, opened 2024-04-08
Freshen the repo's stale scent

StevenBlack:
This PR is purely to freshen the scent conveyed by the readme. If merged, this PR removes the false stale scent therein.

**Previously**: the readme displays "Last updated: October 2019" in Heading 2 font, above the fold as the visitor starts reading.

This conveyed that this repo might be abandonware, which is the impression I conjured on my first TWO visits to the repo home page.

Then after repeatedly seeing links to this repo on socials I thought, why are all these people linking to abandonware?  It's only on my THIRD time that I realized, it's just the H2 that's sending the stale signal.

This PR updates the last time this particular section was updated, February 8 2024.

A pertinent question may be: why display a date in the H2 heading at all?

This PR freshens the date so at least the incorrect stale scent is removed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ For ECC crypto support we are using [Curve25519-dalek](https://crates.io/crates/
 > the tests which exercise the API. Please open issues here if if you are
 > attempting to use rPGP and need help.
 
-## Status (Last updated: October 2019)
+## Status (Last updated: February 2024)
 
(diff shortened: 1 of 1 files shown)


dignifiedquire:
Thanks a lot, took this as inspiration to write: https://github.com/rpgp/rpgp/pull/338, so will close this in favor of that

PR #338, opened 2024-04-08
docs: update and expand status and readme

dignifiedquire:


r10s:
very nice, what a fresh breeze :)

dignifiedquire:
> @dignifiedquire Why remove deps.rs badge? Because of the scary red `rsa` package at https://deps.rs/repo/github/rpgp/rpgp ?

not really, I just don't find it super useful, and never look at it when looking at repos

PR #345, opened 2024-04-13
add pgp::packet::sym_encrypted_protected_data::Data to the public API

hko-s:
I don't think making `SymEncryptedProtectedData::data` pub is right, but `::data()` is already taken (and it doesn't return that field).

We probably want to shuffle this namespace around, some. Do you have an idea/preference?
Maybe `SymEncryptedProtectedData::data()` should be renamed to `::data_as_slice()` or something like that?


hko-s:
I adjusted the PR to an API that seems reasonable to me (and works for using in rsop).

dignifiedquire (approved):
thanks

PR #349, opened 2024-05-10
feat: add APIs for third-party certifications

### LINKED ISSUE #347: Should add an API to issue third-party certifications (e.g. over a UserId)
The function `UserId::sign` (and similar ones) currently assumes that the signer `key` is also the key that the `UserId` should be bound to. This is fine for self-signatures, but doesn't allow third-party signatures.

We should add an API that allows passing in two separate key parameters: a signer and a key to bind (e.g.) the UserId to.

I can propose a PR soonish.

(Cc @kyrias, fyi)

hko-s:
Fixes #347

cc: @kyrias


dignifiedquire:
Api looks good to me, needs a conflict resolution and ideally also a test that signs and verifies such a third party construction

hko-s:
> Api looks good to me, needs a conflict resolution and ideally also a test that signs and verifies such a third party construction

Cool, thanks! I'll add a test soon.

hko-s:
> > Api looks good to me, needs a conflict resolution and ideally also a test that signs and verifies such a third party construction

Added a test for UserId (and APIs for verification of third-party certifications).
But I can't decide if I find the added noise for a UserAttribute unit test worth it %)

dignifiedquire (approved):

PR #351, opened 2024-05-11
Add  NIST P curve support to ECDH

hko-s:


src/crypto/ecdh.rs
```diff
@@ -19,87 +19,196 @@ const ANON_SENDER: [u8; 20] = [
...
 #[derive(Clone, PartialEq, Eq, Zeroize, ZeroizeOnDrop)]
-pub struct SecretKey {
-    /// The secret point.
-    pub secret: [u8; 32],
-    pub hash: HashAlgorithm,
-    pub oid: Vec<u8>,
-    pub alg_sym: SymmetricKeyAlgorithm,
+pub enum SecretKey {
+    // ECDH with Curve25519
+    Curve25519Legacy {
```
dignifiedquire, line 26: why legacy?

hko-s: I was going with the naming in: https://www.ietf.org/archive/id/draft-ietf-openpgp-crypto-refresh-13.html#section-5.5.5.6.1.1

However, I'm unsure if the two Curve 25519 formats in that draft even need to be modeled (and named) as separate `SecretKey` variants. Maybe they don't, then this name would be wrong.

dignifiedquire: I think we should keep calling this `Curve25519` and only rename it to legacy when introducing v6 keys and the non legacy format

src/crypto/ecdh.rs
```diff
@@ -19,87 +19,175 @@ const ANON_SENDER: [u8; 20] = [
...
+            SecretKey::P384 { hash, alg_sym, .. } => (ECCCurve::P384, *alg_sym, *hash),
+            SecretKey::P521 { hash, alg_sym, .. } => (ECCCurve::P521, *alg_sym, *hash),
+        }
     }
 }
 
 impl Decryptor for SecretKey {
     fn decrypt(&self, mpis: &[Mpi], fingerprint: &[u8]) -> Result<Vec<u8>> {
+        /// Generic for Rust Crypto curves
+        fn derive_shared_secret<C>(
```
dignifiedquire, line 89: lets put this function outside of the function, not a big fan of nesting function defs, other for very short helpers

src/crypto/ecdh.rs
```diff
@@ -176,38 +264,99 @@ pub fn derive_session_key(
 
 /// Generate an ECDH KeyPair.
 /// Currently only support ED25519.
```
dignifiedquire, line 266: outdated comment

src/crypto/ecdh.rs
```diff
@@ -176,38 +264,99 @@ pub fn derive_session_key(
...
-            alg_sym,
-        },
-        PlainSecretParams::ECDH(Mpi::from_raw(q)),
-    )
+pub fn generate_key<R: Rng + CryptoRng>(
+    mut rng: R,
+    curve: &ECCCurve,
+) -> Result<(PublicParams, PlainSecretParams)> {
+    /// Generic for Rust Crypto curves
+    fn keygen<C, R: Rng + CryptoRng>(
```
dignifiedquire, line 272: put this outside please

src/crypto/ecdh.rs
```diff
@@ -361,45 +558,52 @@ mod tests {
...
+                        )
+                        .unwrap(),
+                        _ => panic!("invalid key generated"),
+                    };
+
+                    let mpis = mpis.into_iter().map(Into::into).collect::<Vec<Mpi>>();
+
+                    let decrypted = match skey.as_ref().as_repr(&pkey).unwrap() {
+                        SecretKeyRepr::ECDH(ref skey) => skey.decrypt(&mpis, &fingerprint).unwrap(),
+                        _ => panic!("invalid key generated"),
```
dignifiedquire, line 602: wrong error

src/types/params/plain_secret.rs
```diff
@@ -139,18 +139,63 @@ impl<'a> PlainSecretParamsRef<'a> {
...
+                            },
+                        ))
+                    }
+                    ECCCurve::P256 => {
+                        const SEC_LEN: usize = ECCCurve::P256.secret_key_length();
+
+                        ensure!(d.len() <= SEC_LEN, "invalid secret");
+
+                        let mut secret = [0u8; SEC_LEN];
+                        secret[SEC_LEN - d.len()..].copy_from_slice(d.as_bytes());
```
dignifiedquire, line 165: maybe add a helper, sth like this

```rust
fn pad_key<const SIZE: usize>(val: &[u8]) -> Result<[u8; SIZE]> {
  ensure!(d.len() <= SIZE, "invalid key size");
  let mut key = [0u8; SIZE];
  key[SIZE - val.len()..].copy_from_slice(val);
  Ok(key)
}
```

dignifiedquire:
would be nice to add  benchmarks and tests for these new formats, following the existing ones

src/crypto/ecdh.rs
hko-s: Agreed

src/types/params/plain_secret.rs
```diff
@@ -160,10 +193,12 @@ impl<'a> PlainSecretParamsRef<'a> {
...
                     ECCCurve::Ed25519 => {
-                        ensure!(d.len() <= 32, "invalid secret");
+                        const SEC_LEN: usize = ECCCurve::Ed25519.secret_key_length();
+
+                        ensure!(d.len() <= SEC_LEN, "invalid secret");
 
-                        let mut secret = [0u8; 32];
-                        secret[32 - d.len()..].copy_from_slice(d.as_bytes());
+                        let mut secret = [0u8; SEC_LEN];
+                        secret[SEC_LEN - d.len()..].copy_from_slice(d.as_bytes());
```
dignifiedquire, line 201: use the helper here as well?

src/types/params/plain_secret.rs
hko-s: Right.

(Aside: I was surprised to notice that the compiler accepts calls to this kind of function without an explicit value for `SIZE`. It's legal to call just `pad_key(&val)`.
I failed to find Rust documentation that explains this, and if there is a way to tell the compiler that the parameter must be explicitly set.)

hko-s:
> would be nice to add benchmarks and tests for these new formats, following the existing ones

I extended the benchmarks, which seemed straightforward.

Regarding tests i am unsure what to extend.  I did already add nist curve tests to https://github.com/rpgp/rpgp/blob/3b91018aaf2f2ac5a4d1a49f70756e153d63fc18/src/crypto/ecdh.rs#L543-L548

It looks like the corpus in `tests` doesn't have any nist artifacts.
Did you have specific ideas what kinds of tests should be added?

dignifiedquire:
overall looking great, will take some time in the next day to go over the actual crypto algorithms once more

dignifiedquire (approved):

dignifiedquire:
@hko-s want to move it out of draft, or is there sth else you want to do?

dignifiedquire:
thanks  :)

PR #353, opened 2024-05-13
use sha1_checked for all Sha1 hashing, except for MDC

hko-s:


src/crypto/dsa.rs
```diff
@@ -41,7 +41,10 @@ impl Signer for SecretKey {
 
         let signature = match hash_algorithm {
             HashAlgorithm::MD5 => signing_key.sign_prehashed_rfc6979::<md5::Md5>(digest),
+
+            // TODO? we can't easily use use sha1_checked here, because it doesn't implement BlockSizeUser
```
dignifiedquire, line 45: should be easy to add, want to make a PR for this?

src/crypto/sym.rs
```diff
@@ -164,6 +162,21 @@ impl SymmetricKeyAlgorithm {
...
+            T: AsRef<[u8]>,
+            I: IntoIterator<Item = T>,
+        {
+            use sha1::{Digest, Sha1};
+
+            let mut digest = Sha1::new();
+            for chunk in data {
+                digest.update(chunk.as_ref());
+            }
+            digest.finalize()[..20].to_vec()
```
dignifiedquire, line 177: you can return an array here, no need to allocate

src/types/params/plain_secret.rs
```diff
@@ -110,6 +110,7 @@ impl<'a> PlainSecretParamsRef<'a> {
         hasher.finalize().to_vec()
     }
 
+    /// Uses sha1_checked
```
dignifiedquire, line 113: this means these functions panic now on a collision, right?

src/crypto/sym.rs
hko-s: right, cool

src/crypto/dsa.rs
hko-s: see https://github.com/RustCrypto/hashes/pull/582

src/types/params/plain_secret.rs
hko-s: I'll take a stab at adjusting all relevant functions to return a "collision detected" error, instead of doing the hash digest mitigation thing.

(I think that's what you had in mind?)

src/crypto/dsa.rs
hko-s: It's merged :tada:

src/types/params/plain_secret.rs
hko-s: I've adjusted almost all uses of `sha1-checked` to return a new `Error::HashCollision` if a collision is detected.

One intentional deviation from that pattern is fingerprint calculation: I think it is unreasonable to expect consumers of an OpenPGP library to handle the possibility that fingerprint calculation might error out.

My current feeling is that it's a better tradeoff to return the "mitigated" hash digest as a fingerprint instead of returning errors.

src/crypto/dsa.rs
dignifiedquire: nice

src/types/params/plain_secret.rs
dignifiedquire: sounds reasonable

src/crypto/checksum.rs
```diff
@@ -80,16 +79,26 @@ impl Hasher for SimpleChecksum {
     }
 }
 
-/// SHA1 checksum, first 20 octets.
+/// SHA1 checksum, using sha1_checked, first 20 octets.
+///
+/// Fails with `Error::HashCollision` if a SHA1 collision was detected.
 #[inline]
-pub fn calculate_sha1<I, T>(data: I) -> Vec<u8>
+pub fn calculate_sha1<I, T>(data: I) -> Result<Vec<u8>>
```
dignifiedquire, line 86: return an array, and only convert into vec when needed at call site?

src/errors.rs
```diff
@@ -85,6 +85,8 @@ pub enum Error {
     Eax,
     #[error("OCB")]
     Ocb,
+    #[error("Hash collision detected")]
+    HashCollision,
```
dignifiedquire, line 89: maybe `Sha1HashCollision`?

dignifiedquire (commented):
looking good, some nitpicks

src/crypto/checksum.rs
hko-s: Right, sounds good.

(Initially I thought it's better to keep the `Vec<u8>` return type just to avoid gratuitous API breakage. But we're now changing the type to `Result` anyway.)

dignifiedquire (approved):
thanks a lot

PR #365, opened 2024-07-04
feat(secret_key_macros): impl set_password and set_password_and_s2k

hko-s:
API sketch for changing secret key packet locking

### DIFF
--- a/src/packet/key/secret.rs
+++ b/src/packet/key/secret.rs
@@ -5,15 +5,14 @@ use crate::{
     crypto::{hash::HashAlgorithm, public_key::PublicKeyAlgorithm},
     errors::Result,
     packet::{
-        PacketTrait, Signature, SignatureConfigBuilder, SignatureType, Subpacket, SubpacketData,
+        PacketTrait, PublicKey, PublicSubkey, Signature, SignatureConfigBuilder, SignatureType,
+        Subpacket, SubpacketData,
     },
     types::{
         KeyId, KeyVersion, Mpi, PublicKeyTrait, PublicParams, SecretKeyRepr, SecretKeyTrait,
         SecretParams, Tag, Version,
     },
 };
-
-use super::{PublicKey, PublicSubkey};
 
 #[derive(Debug, PartialEq, Eq, Clone, zeroize::Zeroize, zeroize::ZeroizeOnDrop)]
 pub struct SecretKey(SecretKeyInner<PublicKey>);
@@ -37,6 +36,57 @@ impl<D> zeroize::Zeroize for SecretKeyInner<D> {
 impl<D> Drop for SecretKeyInner<D> {
     fn drop(&mut self) {
         self.zeroize()
+    }
+}
+
+impl<D: PublicKeyTrait + crate::ser::Serialize> SecretKeyInner<D> {
+    fn remove_password<P>(&mut self, password: P) -> Result<()>
+    where
+        P: FnOnce() -> String,
+    {
+        if let SecretParams::Encrypted(enc) = &self.secret_params {
+            let unlocked = enc.unlock(
+                password,
+                self.details.algorithm(),
+                self.details.public_params(),
+            )?;
+            self.secret_params = SecretParams::Plain(unlocked);
+        }
+
+        Ok(())
+    }
+
+    fn set_password<P, R>(&mut self, password: P, rng: R) -> Result<()>
+    where
+        P: FnOnce() -> String,
+        R: rand::Rng + rand::CryptoRng,
+    {
+        let s2k = crate::types::S2kParams::new_default(rng);
+        Self::set_password_with_s2k(self, password, s2k)
+    }
+
+    fn set_password_with_s2k<P>(
+        &mut self,
+        password: P,
+        s2k_params: crate::types::S2kParams,
+    ) -> Result<()>
+    where
+        P: FnOnce() -> String,
+    {
+        let plain = match &self.secret_params {
+            SecretParams::Plain(plain) => plain,
+            SecretParams::Encrypted(_) => {
+                bail!("Secret Key packet must be unlocked")
+            }
+        };
+
+        self.secret_params = SecretParams::Encrypted(plain.clone().encrypt(
+            &password(),
+            s2k_params,
+            self.version(),
+        )?);
+
+        Ok(())
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/packet/secret_key_macro.rs
```diff
@@ -150,6 +150,50 @@ macro_rules! impl_secret_key {
...
+                s2k_params: $crate::types::S2kParams,
+            ) -> $crate::errors::Result<()>
+            where
+                O: FnOnce() -> String,
+                N: FnOnce() -> Option<String>,
+            {
+                use $crate::types::SecretParams;
+
+                let plain = match &self.secret_params {
+                    SecretParams::Plain(plain) => plain.clone(),
```
dignifiedquire, line 182: would be nice to avoid this clone (but not critical)

hko-s: The clone also bugged me, but I can't think of a DRY way to avoid it.
It's of course avoidable, but I'm not sure that it's worth avoiding at the cost of messier code.

If you have an idea for a neat approach, I'd be more than happy to change it.

hko-s: I just realized that `PlainSecretParams::encrypt` (which I call, here) takes a `self` too. Which I don't think is necessary, a `&self` should be enough, there.

But unfortunately that makes my (mostly silly) cloning here more necessary.

src/packet/secret_key_macro.rs
```diff
@@ -150,6 +150,50 @@ macro_rules! impl_secret_key {
...
+
+            pub fn set_password<O, N, R>(
+                &mut self,
+                pw_old: O,
+                pw_new: N,
+                rng: R,
+            ) -> $crate::errors::Result<()>
+            where
+                O: FnOnce() -> String,
+                N: FnOnce() -> Option<String>,
```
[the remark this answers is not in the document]
dignifiedquire: that should be fine, yes

src/packet/secret_key_macro.rs
dignifiedquire: ok :/ lets leave it for now

src/packet/secret_key_macro.rs
```diff
@@ -150,6 +150,50 @@ macro_rules! impl_secret_key {
                     .build()?
                     .sign_key(key, key_pw, &self)
             }
+
+            pub fn set_password<O, N, R>(
```
dignifiedquire, line 154: needs some doc comment :)

src/packet/secret_key_macro.rs
```diff
@@ -150,6 +150,50 @@ macro_rules! impl_secret_key {
...
+            where
+                O: FnOnce() -> String,
+                N: FnOnce() -> Option<String>,
+                R: rand::Rng + rand::CryptoRng,
+            {
+                let s2k = $crate::types::S2kParams::new_default(rng);
+                Self::set_password_and_s2k(self, pw_old, pw_new, s2k)
+            }
+
+            pub fn set_password_and_s2k<O, N>(
```
dignifiedquire, line 169: this one too

src/packet/secret_key_macro.rs
```diff
@@ -151,30 +151,44 @@ macro_rules! impl_secret_key {
...
+            /// `pw_old`.
+            ///
+            /// The s2k mechanism is set to `s2k_params`, if `pw_new` is `Some`.
+            /// If `pw_new` is `None`, `s2k_params` is not used.
+            pub fn set_password_and_s2k<P>(
                 &mut self,
-                pw_old: O,
-                pw_new: N,
+                pw_old: P,
+                pw_new: Option<P>,
```
hko-s, line 187: I am tempted to change the API to

`pw_new: Option<(P, $crate::types::S2kParams)>`

Tuples are a bit odd as fn parameters. But it would remove the weird situation of passing an `S2kParams` that is not used when `pw_new` is `None`, at the type level.

wdyt?

dignifiedquire:
we should have a test for this ;) 

also `Option<T>` is very annoying if you pass `None`, unless you set a default type parameter

src/packet/secret_key_macro.rs
dignifiedquire: I wonder if it isn't easier at this point to have a specific call for the `None` case

hko-s: Good point. I've pushed a change that explores that approach.

I'm half tempted to drop `pw_old` from the setter functions, now.
Callers could then unlock and lock explicitly, as needed.

hko-s: Pushed a commit that completely separates locking and unlocking.
I like that change - in part because it allows aligning function names with the terms in the RFC.

src/packet/secret_key_macro.rs
```diff
@@ -151,18 +151,42 @@ macro_rules! impl_secret_key {
                     .sign_key(key, key_pw, &self)
             }
 
+            /// Unlock the private key material of this secret key packet.
+            /// This removes password protection ("locking") of the secret key material.
+            ///
+            /// If the Secret Key packet is not locked, it is left unchanged.
+            ///
+            /// The current locking password for this key must be provided in `password`.
+            pub fn unset_password<P, R>(&mut self, password: P) -> $crate::errors::Result<()>
```
dignifiedquire: bikeshed `remove_password`?

dignifiedquire: no need for the rng parameter

hko-s: > no need for the rng parameter

oops. of course. removed.

src/packet/secret_key_macro.rs
```diff
@@ -150,6 +150,75 @@ macro_rules! impl_secret_key {
...
                     .sign_key(key, key_pw, &self)
             }
+
+            /// Unlock the private key material of this secret key packet.
+            /// This removes password protection ("locking") of the secret key material.
+            ///
+            /// If the Secret Key packet is not locked, it is left unchanged.
+            ///
+            /// The current locking password for this key must be provided in `password`.
+            pub fn password_unlock<P>(&mut self, password: P) -> $crate::errors::Result<()>
```
dignifiedquire, line 160: bikeshedding: `unlock` to me sounds very temporary

hko-s: I see what you mean. The RFC uses "lock" and "unlock", and I like that use of these two terms is specific to s2k-based protection of secret key material.

However, the RFC only really uses "unlock" in the context of a temporary operation. So I guess you're right.
Names like `remove_password` and `set_password` are probably better, after all.

hko-s: pushed a naming change (and more explicit documentation for the "change password" use case)

hko-s:
> we should have a test for this ;)

Absolutely, will make one, once the API starts to settle ;o)

src/packet/key/secret.rs
```diff
@@ -39,6 +39,75 @@ impl<D> Drop for SecretKeyInner<D> {
     }
 }
 
+impl<D: PublicKeyTrait + crate::ser::Serialize> SecretKeyInner<D> {
+    /// Remove the password protection of the private key material in this secret key packet.
+    /// This permanently "unlocks" the secret key material.
+    ///
+    /// If the Secret Key material in the packet is not locked, it is left unchanged.
+    ///
+    /// The current locking password for this key must be provided in `password`.
```
dignifiedquire, line 48: the docs should be on the public version, as these will not show up for users

hko-s: good point.

moved the password-related functions into a trait, in part to have a shared place for the docs.
i'm very open for thoughts and bikeshedding :)

dignifiedquire: why make this a new trait, instead of adding to the existing `SecretKeyTrait`?

hko-s: These password operations are kind of specific to Secret Key Packets.
They don't apply to some other implementations of `SecretKeyTrait`, such as HSM-backed keys

hko-s: Maybe the trait should be named `SecretKeyPacketProtection`

dignifiedquire: then lets not use a trait but impl them directly, I don’t think it’s good to have an api with a bunch of traits

hko-s: That way we'll have to duplicate the doc comments, but I'm not super opposed to the approach

src/packet/key/secret.rs
```diff
@@ -38,48 +38,7 @@ impl<D> Drop for SecretKeyInner<D> {
     }
 }
 
-/// Altering the password protection state of a Secret Key packet.
-/// Removing or adding password protection for the private key material in a Secret Key packet.
-///
-/// See https://www.rfc-editor.org/rfc/rfc4880.html#section-3.7.2.1
-pub trait SecretKeyProtection {
```
hko-s: Final bit of musing on the trait: I had wondered if/hoped that this trait might enable users of the library to abstract over `SecretKey` and `SecretSubkey`. But am unsure if that would actually ever be useful in practice.

src/packet/key/secret.rs
```diff
@@ -451,3 +501,161 @@ impl<D: PublicKeyTrait + crate::ser::Serialize> PublicKeyTrait for SecretKeyInne
...
+        s2k_params: crate::types::S2kParams,
+    ) -> Result<()>
+    where
+        P: FnOnce() -> String,
+    {
+        self.0.set_password_with_s2k(password, s2k_params)
+    }
+}
+
+#[test]
```
dignifiedquire, line 597: please use 

```rust
#[cfg(test)]
mod tests {
...
}
```

dignifiedquire (approved):
one last nit

src/packet/key/secret.rs
hko-s: right, thanks

PR #366, opened 2024-07-06
refactor: remove bigger macros in favor of direct types

dignifiedquire:


src/composed/signed_key/key_parser.rs
```diff
@@ -0,0 +1,229 @@
...
+use crate::{SignedKeyDetails, SignedPublicSubKey, SignedSecretSubKey};
+
+#[allow(clippy::complexity)]
+pub fn next<I, IKT>(
+    packets: &mut std::iter::Peekable<I>,
+    key_tag: Tag,
+    parse_secrect_subkeys: bool,
+) -> Option<
+    Result<(
+        IKT,
```
hko-s, line 13: I'm unsure how to think about this generic.

I think the `IKT` should definitely never be a subkey type. Shouldn't it always be a `packet::PublicKey`?

hko-s: Ah. Or a `packet::SecretKey`?

Still, it seems to me the type system isn't currently guaranteeing that we won't get a subkey packet in this location? ... which would be a bit sad.

src/packet/key/public.rs
```diff
@@ -0,0 +1,506 @@
...
+        Ok(Self(inner))
+    }
+
+    /// Parses a `PublicKeyKey` packet from the given slice.
+    pub fn from_slice(packet_version: Version, input: &[u8]) -> Result<Self> {
+        let inner = PubKeyInner::from_slice(packet_version, input)?;
+        Ok(Self(inner))
+    }
+
+    pub fn sign<F>(&self, key: &impl SecretKeyTrait, key_pw: F) -> Result<Signature>
```
hko-s, line 45: I wonder if this function (and its subkey sibling, below) should both be named `bind`, to be slightly more self-explanatory.

src/composed/signed_key/key_parser.rs
dignifiedquire: I am confused, it is quite strictly reduced type wise in the where clause, what are you worried about?

hko-s: I was wondering if it'd be possible and useful to disallow subkey types in this place. But I have no strong feelings about this.

hko-s:
I haven't tried to verify that the transformations from macro to plain code are 1:1 the same, but assume they are.

That said, this PR seems like a great improvement to me - and using it as a dependency worked well, where i tried.

src/packet/key/public.rs
hko-s: Moving this point to https://github.com/rpgp/rpgp/issues/367, it is separate from this PR

PR #371, opened 2024-07-11
fix: decrypted data must contain exactly one message

hko-s:
I think performing these checks for all versions of encrypted messages is the right thing to do.

### DIFF
--- a/src/composed/message/types.rs
+++ b/src/composed/message/types.rs
@@ -167,6 +167,25 @@ impl Edata {
         }
     }
 
+    /// Transform decrypted data into a message.
+    /// Bails if the packets contain no message or multiple messages.
+    fn process_decrypted(packet_data: &[u8]) -> Result<Message> {
+        let mut messages = Message::from_bytes_many(packet_data);
+        // First message is the one we want to return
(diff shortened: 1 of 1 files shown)


src/composed/message/types.rs
```diff
@@ -167,6 +167,25 @@ impl Edata {
...
+        let mut messages = Message::from_bytes_many(packet_data);
+        // First message is the one we want to return
+        let Some(message) = messages.next() else {
+            bail!("no valid message found");
+        };
+        let message = message?;
+
+        // The only other message allowed is a padding packet, which will be skipped
+        // by the parser, so check that we have only a single message.
+        if let Some(msg) = messages.next() {
```
dignifiedquire, line 182: what would happen if we have multiple padding packets?

hko-s: Note that the logic here is unchanged from the V6 session key branch, before. The functionality is just factored out to be easy to call from all version-specific branches.

That said, the `messages` iterator should return `None` for any number of padding packets.

dignifiedquire (approved):

PR #372, opened 2024-07-25
Update Cargo.toml

Dylan-DPC:
Bumps the dependency to at least 4.1.3 as per the advisory

[Advisory](https://rustsec.org/advisories/RUSTSEC-2024-0344.html) 

(even though this crate may not be using the function mentioned in the advisory, it is better to bump it to ensure that there is no chance of other dependencies pulling it)


dignifiedquire (commented):
thanks

dignifiedquire (approved):

PR #373, opened 2024-07-26
Re-export bstr for user to use

DoumanAsh:
bstr is used in public API therefore it must be re-exported to alloow user to use it without need to manually check which version you use


dignifiedquire:
Thanks, but we have multiple crates that are used in the public API, and it has been a conscious decision against reexporting them for now. This might change in the future, but if so we will reexport all of them at once.

PR #374, opened 2024-07-26
Improve API of LIteralData to provide more flexibility

DoumanAsh:


src/packet/literal_data.rs
```diff
@@ -84,6 +84,24 @@ impl LiteralData {
...
 
+    #[inline]
+    /// Extracts data in to raw data
+    pub fn into_bytes(self) -> Vec<u8> {
+        self.data
+    }
+
+    #[inline]
+    /// Extracts data as string, returning raw bytes as Err if not valid utf-8 string
+    pub fn into_string(self) -> Result<String, Vec<u8>> {
```
dignifiedquire, line 95: Given that `to_string` returns `None` in the error case, I think this api should mirror that. If you want this exact behaviour, I would suggest renaming this method into sth like `try_into_string`

DoumanAsh: This behavior is not very useful for owned version because there is no way to examine what is wrong with data

hko-s: Renaming to `pub fn try_into_string(self) -> Result<String, Vec<u8>>` seems like a good approach to me, then

hko-s: From the point of view of OpenPGP semantics, I tend to think that `DataMode::Binary` should also be handled like the `_` case (i.e.: I think the `match` should be dropped).

As far as I can tell, the `DataMode` of a literal packet isn't very intimately linked with the expected content of the packet. So if we're willing to try `String::from_utf8` for `DataMode::Binary`, I think we should do for all cases.

src/packet/literal_data.rs
```diff
@@ -84,6 +84,24 @@ impl LiteralData {
...
+    /// Extracts data in to raw data
+    pub fn into_bytes(self) -> Vec<u8> {
+        self.data
+    }
+
+    #[inline]
+    /// Extracts data as string, returning raw bytes as Err if not valid utf-8 string
+    pub fn try_into_string(self) -> Result<String, Vec<u8>> {
+        match self.mode {
+            DataMode::Binary => Err(self.data),
```
hko-s, line 97: From the point of view of OpenPGP semantics, I tend to think that DataMode::Binary should also be handled like the _ case (i.e.: I think the match should be dropped).

As far as I can tell, the DataMode of a literal packet isn't very intimately linked with the expected content of the packet. So if we're willing to try String::from_utf8 for DataMode::Binary, I think we should do for all cases.

DoumanAsh: I followed `to_string` implementation which ignores Binary variant 🤔

hko-s: Hm, I guess that's fair enough. Sorry, ignore this point :)

dignifiedquire (approved):
thanks

PR #379, opened 2024-08-01
feat: rfc9580 keys

hko-s:


src/crypto/eddsa.rs
```diff
@@ -48,62 +58,57 @@ impl Signer for SecretKey {
...
         ECCCurve::Ed25519 => {
-            ensure_eq!(sig.len(), 2);
-
-            let r = sig[0].as_bytes();
-            let s = sig[1].as_bytes();
-
-            ensure!(r.len() < 33, "invalid R (len)");
-            ensure!(s.len() < 33, "invalid S (len)");
-            ensure_eq!(q.len(), 33, "invalid Q (len)");
-            ensure_eq!(q[0], 0x40, "invalid Q (prefix)");
```
dignifiedquire, line 93: where did this logic go?

src/packet/key/public.rs
```diff
@@ -394,7 +485,24 @@ impl PublicKeyTrait for PubKeyInner {
...
+
+        match self.version {
+            KeyVersion::V2 | KeyVersion::V3 | KeyVersion::V4 => {
+                // When a v4 signature is made over a key, the hash data starts with the octet 0x99,
+                // followed by a two-octet length of the key, and then the body of the key packet.
+                writer.write_all(&[0x99, (key_buf.len() >> 8) as u8, key_buf.len() as u8])?;
+            }
+
+            KeyVersion::V6 => {
+                // When a v6 signature is made over a key, the hash data starts with the salt,
```
dignifiedquire, line 497: where is the salt?

src/packet/one_pass_signature.rs
```diff
@@ -32,14 +34,14 @@ impl OnePassSignature {
...
-        if pk.version != 2 && pk.version != 3 && pk.version != 4 && pk.version != 5 {
+        if pk.version != 2 && pk.version != 3 && pk.version != 4 && pk.version != 6 {
             unsupported_err!("unsupported signature version {}", pk.version);
         }
 
         Ok(pk)
     }
 
-    pub fn from_details(
+    pub fn from_details_v3(
```
dignifiedquire, line 44: don't we need a `v4`?

src/crypto/eddsa.rs
hko-s: To `src/composed/signed_key/public.rs`. My idea was that the "legacy" format specific stuff is abstracted away there, while `eddsa.rs` can more format-agnostic.

(Not sure if that's the right idea though, seeing how key generation in essda.rs does OpenPGP format-specific things :thinking:)

src/packet/secret_key_parser.rs
```diff
@@ -38,12 +39,22 @@ fn new_private_key_parser(
     |i: &[u8]| {
         let (i, created_at) = map_opt(be_u32, |v| Utc.timestamp_opt(i64::from(v), 0).single())(i)?;
         let (i, alg) = map(be_u8, PublicKeyAlgorithm::from)(i)?;
-        let (i, params) = parse_pub_priv_fields(alg)(i)?;
+
+        let i = if *key_ver == KeyVersion::V6 {
+            // "scalar octet count for the following public key material" -> skip
+            let (i, _len) = be_u32(i)?;
```
dignifiedquire, line 45: shouldn't be skipped, but validated

src/packet/one_pass_signature.rs
hko-s: OnePassSignature is only defined for v3 and v6

src/packet/signature/types.rs
```diff
@@ -713,7 +749,7 @@ impl SubpacketType {
             SubpacketType::SignatureTarget => 31,
             SubpacketType::EmbeddedSignature => 32,
             SubpacketType::IssuerFingerprint => 33,
-            SubpacketType::PreferredAead => 34,
+            SubpacketType::PreferredAead => 39,
```
dignifiedquire, line 752: did this change? what about `34`?

src/packet/key/public.rs
hko-s: In `src/packet/signature/config.rs`. But that's silly. I'll move it.

src/packet/signature/types.rs
```diff
@@ -822,11 +859,8 @@ pub enum SubpacketData {
     TrustSignature(u8, u8),
     RegularExpression(BString),
     ExportableCertification(bool),
-    IssuerFingerprint(
-        KeyVersion,
-        #[debug("{}", hex::encode(_1))] SmallVec<[u8; 20]>,
-    ),
-    PreferredAeadAlgorithms(SmallVec<[AeadAlgorithm; 2]>),
+    IssuerFingerprint(KeyVersion, #[debug("{}", hex::encode(_1))] Vec<u8>),
```
dignifiedquire, line 862: why not a smallvec anymore?

src/packet/signature/types.rs
hko-s: In RFC 9580, subpacket type 34 is "Reserved": https://www.rfc-editor.org/rfc/rfc9580.html#name-signature-subpacket-specifi

This change of ID is part of the IETF process' shift from v5 to v6, and vacating numbers that were previously used in 4880bis.

src/crypto/eddsa.rs
dignifiedquire: ah okay, makes sense, just trying to understand

src/packet/one_pass_signature.rs
dignifiedquire: right.. that is not confusing at all..

src/packet/signature/types.rs
hko-s: Good question. I changed Fingerprints to use Vec instead of SmallVec, because we now need Fingerprints to be dynamically either 20 or 32 bytes long, and I wasn't sure if that's reasonable to do with SmallVec.

If it is, I should change Fingerprints back to using smallvec.

src/crypto/hash.rs
```diff
@@ -38,6 +38,22 @@ impl Default for HashAlgorithm {
     }
 }
 
+impl HashAlgorithm {
+    /// V6 signature salt size
+    /// https://www.rfc-editor.org/rfc/rfc9580.html#hash-algos
+    pub fn salt_len(&self) -> usize {
```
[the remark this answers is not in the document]
hko-s: right, done

src/packet/signature/types.rs
dignifiedquire: I wonder if we shouldn't just do sth like this

```rust
pub enum Fingerprint {
    V4([u8; 20]),
    V6([u8; 32]),
}
```

src/packet/key/public.rs
hko-s: On reflection, the existing code is right and should not be changed:

The v6 salt is part of the `SignatureConfig` and gets hashed in that top-level context. When a key is serialized for hashing (that's what we do here), the serialized key data comes after that salt.

For example, a key binding signature will hash roughly the following data:

- The salt
- Some more signature metadata from the `SignatureConfig`
- The first key (serialized with this code)
- The second key (serialized with this code)

src/packet/one_pass_signature.rs
hko-s: right, we definitely need more documentation pointing people into the RFC.
i haven't spent a lot of time thinking about that, yet.

src/packet/signature/de.rs
```diff
@@ -268,17 +267,32 @@ fn embedded_sig(i: &[u8]) -> IResult<&[u8], SubpacketData> {
 /// Parse an issuer subpacket
 /// Ref: https://tools.ietf.org/html/draft-ietf-openpgp-rfc4880bis-05#section-5.2.3.28
 fn issuer_fingerprint(i: &[u8]) -> IResult<&[u8], SubpacketData> {
-    map(
-        pair(map(be_u8, KeyVersion::from), rest),
-        |(version, fingerprint)| {
-            SubpacketData::IssuerFingerprint(version, SmallVec::from_slice(fingerprint))
-        },
-    )(i)
+    let (i, version) = be_u8(i)?;
```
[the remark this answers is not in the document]
hko-s: good idea, done.

(fwiw, i expect to transform this code some more, when introducing a Fingerprint type)

src/packet/secret_key_parser.rs
hko-s: i've taken a stab at using this data, and like the general shape it's taking.

however, i feel like i'm doing error handling wrong, in the nom code: https://github.com/rpgp/rpgp/pull/379/commits/6440eb831ed231ac261fbbca0347fd9678a85f0d#diff-a690e568e7a7967c9ea9d6230ca7798089e653a14d39e45b7cfce5ac262b39eeR21-R30

it looks very clumsy - is there a way to do this more nicely?

src/packet/signature/types.rs
hko-s: > I wonder if we shouldn't just do sth like this
> 
> ```rust
> pub enum Fingerprint {
>     V4([u8; 20]),
>     V6([u8; 32]),
> }
> ```

I've taken a stab at such a change in 4e6dc901ba583406fcb3c1683219d2a827895ada, and would love to hear your thoughts on the direction.

The change feels in part reasonable, to me. But in some places we now have redundancy between the fingerprint type (which encodes a key version), and a separate key version field. I haven't yet decided how I feel about that.

dignifiedquire: overall I think this makes the state strictly better

src/packet/key/public.rs
hko-s: I (hopefully) clarified this comment

src/packet/secret_key_parser.rs
dignifiedquire: no, this is unfortunately the way it works atm

src/composed/cleartext.rs
```diff
@@ -295,7 +299,7 @@ fn validate_headers(headers: Headers) -> Result<Vec<HashAlgorithm>> {
 ///
 /// This implementation is implicitly agnostic between "\n" and "\r\n" line endings.
 ///
-/// Ref https://www.ietf.org/archive/id/draft-ietf-openpgp-crypto-refresh-13.html#name-dash-escaped-text
+/// Ref https://www.rfc-editor.org/rfc/rfc9580.html#name-dash-escaped-text
```
dignifiedquire, line 302: why the change to the URLs?

src/composed/cleartext.rs
```diff
@@ -554,8 +559,13 @@ mod tests {
     fn test_sign() {
         let key_data = std::fs::read_to_string("./tests/unit-tests/cleartext-key-01.asc").unwrap();
         let (key, _) = SignedSecretKey::from_string(&key_data).unwrap();
-        let msg = CleartextSignedMessage::sign("hello\n-world-what-\nis up\n", &key, String::new)
-            .unwrap();
+        let msg = CleartextSignedMessage::sign(
+            &mut thread_rng(),
```
dignifiedquire, line 563: ideally we would switch to a seeded rng in all tests, so that they are reproducible

dignifiedquire: chacha8 is crypto + seeded

src/composed/cleartext.rs
hko-s: RFC 9580 was finally released a few days ago. I think we should point to the finalized RFC URLs, not the (pre-publication) draft.

src/composed/message/decrypt.rs
```diff
@@ -14,7 +14,7 @@ where
     debug!("decrypt session key");
 
     locked_key.unlock(key_pw, |priv_key| {
-        let (key, sym_alg) = priv_key.decrypt(mpis, &locked_key.fingerprint())?;
+        let (key, sym_alg) = priv_key.decrypt(mpis, locked_key.fingerprint().as_bytes())?;
         // TODO: handle other versions
```
dignifiedquire, line 18: uhm, what about this todo?

src/composed/message/types.rs
```diff
@@ -403,28 +403,29 @@ impl Message {
     }
 
     /// Sign this message using the provided key.
-    pub fn sign<F>(
+    pub fn sign<R, F>(
         self,
+        rng: &mut R,
```
dignifiedquire, line 408: should normalize to `mut rng:  R`

src/composed/signed_key/mod.rs
```diff
@@ -87,6 +88,7 @@
 //!         packet::Subpacket::regular(packet::SubpacketData::Issuer(signing_key.key_id())),
 //!     ],
 //!     vec![],
+//!     SignatureVersionSpecific::V4 {}
```
dignifiedquire, line 91: this is unfortunate

src/composed/signed_key/public.rs
```diff
@@ -83,6 +83,7 @@ impl SignedPublicKey {
         mut public_subkeys: Vec<SignedPublicSubKey>,
     ) -> Self {
         public_subkeys.retain(|key| {
+            // FIXME: is it useful to make subkey retention depend on unvalidated signatures?
```
dignifiedquire, line 86: issue?

src/composed/key/builder.rs
```diff
@@ -6,11 +6,12 @@ use rand::{thread_rng, CryptoRng, Rng};
 use smallvec::SmallVec;
 
 use crate::composed::{KeyDetails, SecretKey, SecretSubkey};
+use crate::crypto::aead::AeadAlgorithm;
```
dignifiedquire, line 9: can you expand the tests in this file please to cover all current versions & algorithms please

src/crypto/eddsa.rs
```diff
@@ -8,6 +8,11 @@ use crate::crypto::Signer;
 use crate::errors::Result;
 use crate::types::{Mpi, PlainSecretParams, PublicParams};
 
+pub enum Mode {
```
dignifiedquire, line 11: docs

src/crypto/hash.rs
```diff
@@ -38,6 +38,22 @@ impl Default for HashAlgorithm {
...
+    /// https://www.rfc-editor.org/rfc/rfc9580.html#hash-algos
+    pub const fn salt_len(&self) -> usize {
+        match self {
+            Self::SHA2_224 => 16,
+            Self::SHA2_256 => 16,
+            Self::SHA2_384 => 24,
+            Self::SHA2_512 => 32,
+            Self::SHA3_256 => 16,
+            Self::SHA3_512 => 32,
+            _ => 0,
```
dignifiedquire, line 52: maybe Option, and return None here?

src/crypto/x25519.rs
```diff
@@ -0,0 +1,57 @@
...
+
+/// Secret key for X25519
+#[derive(Clone, PartialEq, Eq, Zeroize, ZeroizeOnDrop)]
+pub struct SecretKey {
+    pub(crate) secret: [u8; 32],
+}
+
+impl fmt::Debug for SecretKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("X25519SecretKey")
```
dignifiedquire, line 19: use `derive_more`

src/crypto/x25519.rs
```diff
@@ -0,0 +1,57 @@
...
+    fn key_params(&self) {}
+}
+
+impl Decryptor for SecretKey {
+    // - Ephemeral X25519 public key (32 bytes).
+    // - Recipient public key (32 bytes). [FIXME: use PublicParams and expect PublicParams::X25519?]
+    // - Encrypted and wrapped session key.
+
+    fn decrypt(&self, _mpis: &[Mpi], _fingerprint: &[u8]) -> Result<Vec<u8>> {
+        unimplemented_err!("decrypt for x25519")
```
dignifiedquire, line 37: what's the plan here?

src/packet/one_pass_signature.rs
```diff
@@ -3,97 +3,168 @@ use std::io;
...
     pub typ: SignatureType,
     pub hash_algorithm: HashAlgorithm,
     pub pub_algorithm: PublicKeyAlgorithm,
-    pub key_id: KeyId,
     pub last: u8,
+    pub version_specific: OpsVersionSpecific,
+}
+
+#[derive(derive_more::Debug, Clone, PartialEq, Eq)]
+pub enum OpsVersionSpecific {
```
dignifiedquire, line 28: docs

src/crypto/x25519.rs
hko-s: I plan to do encryption-related functionality right after we're doing with this PR, including x25519 and seipdv2

src/types/mod.rs
```diff
@@ -21,3 +25,75 @@ pub use self::s2k::*;
...
+///
+/// It is an element of a [pgp::packet::Signature] packet.
+/// Historically, cryptographic signatures in OpenPGP were encoded as [pgp::types::Mpi],
+/// however, in RFC 9580, native encoding is used for the modern Ed25519 and Ed448 signatures.
+///
+/// This type can represent both flavors of cryptographic signature data.
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub enum Sig {
+    Mpis(Vec<Mpi>),
+    Native(Vec<u8>),
```
dignifiedquire, line 39: docs

src/composed/key/public.rs
```diff
@@ -76,7 +82,7 @@ impl PublicKeyTrait for PublicKey {
     fn algorithm(&self) -> PublicKeyAlgorithm {
         self.primary_key.algorithm()
     }
-    fn verify_signature(&self, hash: HashAlgorithm, data: &[u8], sig: &[Mpi]) -> Result<()> {
+    fn verify_signature(&self, hash: HashAlgorithm, data: &[u8], sig: &Sig) -> Result<()> {
```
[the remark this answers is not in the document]
hko-s: I don't like `Sig` either, and like `SignatureBytes` a lot better. Let's do that.

(It's a bit sad that the RFCs have no term for .. this)

src/composed/message/decrypt.rs
dignifiedquire, line 18: what about those other versions?

src/packet/one_pass_signature.rs
```diff
@@ -3,97 +3,168 @@ use std::io;
...
+    pub fn version(&self) -> u8 {
+        match self.version_specific {
+            OpsVersionSpecific::V3 { .. } => 3,
+            OpsVersionSpecific::V6 { .. } => 6,
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct OnePassSignatureCommon {}
```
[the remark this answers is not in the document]
hko-s: Ugh, this struct is not needed at all, it's an artifact of a previous state of refactoring. I'll kick it out.

(A similar situation exists for `SignatureVersionSpecific::V4 {}`, which you made a separate review comment about)

hko-s:
It might be good to rename all functions that take an rng parameter into `*_with_rng`, and add a variant function with the short name that implicitly uses `thread_rng`.

src/composed/cleartext.rs
hko-s: I've adjusted the new tests to use ChaCha8.

While looking around, I noticed that some older tests use `StdRng` and `XorShiftRng`. Should we leave those, or normalize them to all use ChaCha8?

src/composed/message/decrypt.rs
hko-s: The missing functionality for this todo will be part of the followup PR that will handle v6 encryption

src/composed/message/types.rs
hko-s: :+1: done

src/composed/signed_key/public.rs
hko-s: right, this is of course unrelated to v6. moved to https://github.com/rpgp/rpgp/issues/383

src/packet/mod.rs
```diff
@@ -48,18 +48,13 @@
...
 //! let now = chrono::Utc::now();
 //!
-//! let mut sig_cfg_bldr = SignatureConfigBuilder::default();
+//! let mut sig_cfg_bldr = SignatureConfigBuilder::v4(packet::SignatureType::Binary,PublicKeyAlgorithm::RSA,HashAlgorithm::SHA2_256);
 //! let sig_cfg = sig_cfg_bldr
-//!      .version(packet::SignatureVersion::V4)
-//!      .typ(packet::SignatureType::Binary)
-//!      .pub_alg(PublicKeyAlgorithm::RSA)
-//!      .hash_alg(HashAlgorithm::SHA2_256)
 //!      .unhashed_subpackets(vec![]) // must be initialized
```
dignifiedquire, line 53: they don't need to be initialized anymore

src/composed/key/builder.rs
hko-s: good point, thanks. i've tried to cover all new variants in 7cd74f89bbeab68b6aafac666f2ea855040524f7.

we can't currently do the RSA test (key locking isn't fully in this PR, for v6 keys), i'd  just leave that as a TODO until the followup PR that implements the new encryption mechanisms (including "v6 key locking").

orthogonally: it occurred to me that it would be nice if these tests also did an encrypt-decrypt roundtrip with the encryption subkey (if you agree, i'd propose i open a ticket about that, and do that outside of this PR).

dignifiedquire: yes, sounds good

src/crypto/hash.rs
hko-s: i was worried that change might be messy, but after an attempt, i like it: 081e1f631c14c796a563ba3b72a761132dd73b09

src/composed/signed_key/mod.rs
hko-s: I think this has been conclusively resolved by our iterative changes to Builder/SignatureConfig creation

src/composed/signed_key/mod.rs
```diff
@@ -87,6 +88,7 @@
 //!         packet::Subpacket::regular(packet::SubpacketData::Issuer(signing_key.key_id())),
 //!     ],
 //!     vec![],
+//!     SignatureVersionSpecific::V4,
```
dignifiedquire, line 91: can we instead have `Signature::v4`?

src/packet/key/secret.rs
```diff
@@ -451,24 +511,48 @@ impl PublicKeyTrait for SecretSubkey {
...
-        plain: &[u8],
-    ) -> Result<Vec<Mpi>> {
+    fn encrypt<R: rand::Rng + rand::CryptoRng>(&self, rng: R, plain: &[u8]) -> Result<Vec<Mpi>> {
         self.details.encrypt(rng, plain)
     }
 
     fn serialize_for_hashing(&self, writer: &mut impl std::io::Write) -> Result<()> {
+        use byteorder::{BigEndian, WriteBytesExt};
+
+        // FIXME: avoid duplication with key/public.rs!
```
dignifiedquire, line 530: still want to do this?

hko-s: oh right. i'd like to, but at some point confused myself into being unsure if it's easily possible.

src/packet/signature/config.rs
```diff
@@ -24,42 +22,185 @@ pub struct SignatureConfig {
...
+        match version {
+            SignatureVersion::V6 => Ok(SignatureVersionSpecific::V6 {
+                salt: Self::v6_salt_for(rng, hash_alg)?,
+            }),
+            SignatureVersion::V4 => Ok(SignatureVersionSpecific::V4),
+            _ => bail!("Unsupported signature version {:version?}"),
+        }
+    }
+
+    pub fn new_v4_v6<R>(
```
dignifiedquire, line 161: do we need this? it reads really odd, and with the more specific `v4` and `v6` methods it seems redundant

src/composed/signed_key/mod.rs
hko-s: So, to clarify: you're thinking of adding `::vX()` constructors to the `SignatureVersionSpecific` enum?

We're not constructing that type a lot, anymore, I believe.

src/packet/signature/config.rs
```diff
@@ -2,20 +2,18 @@ use std::io::Read;
...
 use crate::ser::Serialize;
-use crate::types::{KeyId, PublicKeyTrait, SecretKeyTrait, Tag};
+use crate::types::{Fingerprint, KeyId, PublicKeyTrait, SecretKeyTrait, Tag};
 
-#[derive(Clone, PartialEq, Eq, Debug, Builder)]
-#[builder(build_fn(error = "Error"))]
+#[derive(Clone, PartialEq, Eq, Debug)]
 pub struct SignatureConfig {
-    #[builder(default)]
     pub version: SignatureVersion,
```
dignifiedquire, line 17: I don't think we need this anymore, `version_specific` already encodes this

hko-s: Almost, with the exception of the distinction between V2 and V3.

But it might be nicer to add `V2` to the version specific enum, and drop `version` here.

src/packet/signature/ser.rs
```diff
@@ -286,7 +315,16 @@ impl SignatureConfig {
             packet.to_writer(&mut unhashed_subpackets)?;
         }
 
-        writer.write_u16::<BigEndian>(unhashed_subpackets.len() as u16)?;
+        match self.version {
+            SignatureVersion::V4 => {
+                writer.write_u16::<BigEndian>(unhashed_subpackets.len() as u16)?
```
dignifiedquire, line 320: should use `try_from` for these

src/types/fingerprint.rs
```diff
@@ -0,0 +1,87 @@
...
+    ///
+    /// However, sometimes a fingerprint may be obtained where the key version is unknown.
+    /// Then, this is the only possible way to encode it.
+    #[allow(dead_code)]
+    pub(crate) fn new_unknown(fp: &[u8]) -> Result<Self> {
+        Ok(Fingerprint::Unknown(Box::from(fp)))
+    }
+
+    #[allow(clippy::len_without_is_empty)]
+    pub fn len(&self) -> usize {
```
dignifiedquire, line 59: docs here and below

src/types/packet.rs
```diff
@@ -142,17 +150,56 @@ pub enum KeyVersion {
...
 
+impl KeyVersion {
+    /// Size of OpenPGP fingerprint in bytes
+    /// (returns 0 for unknown versions)
+    pub const fn fingerprint_len(&self) -> usize {
+        match self {
+            KeyVersion::V2 | KeyVersion::V3 => 16, // MD5
+            KeyVersion::V4 => 20,                  // SHA1
+            KeyVersion::V5 | KeyVersion::V6 => 32, // SHA256
+            KeyVersion::Other(_) => 0,             // FIXME?
```
dignifiedquire, line 167: return Option/None ?

src/composed/signed_key/mod.rs
dignifiedquire: no, I mean adding a `vX` constructor to `Signature`, replacing the generic `new` constructor

src/packet/signature/config.rs
hko-s: f06fc9d5617ff9190028e2327e39c640319892bb

src/packet/signature/config.rs
hko-s: a while ago i thought it might be useful to have as a convenience constructor.
but dropping it seems right, yes: 33e884c701599218cdc61fc0d5b50e37993335d0

src/composed/signed_key/mod.rs
hko-s: right, that looks good 3750727b0684f7c64b6cec086a1cb028b205d5aa

src/packet/key/secret.rs
hko-s: yay, that's less annoying. 80a8a2d691de5b457c5ad9e8809828de22896cc9

src/types/packet.rs
hko-s: i'm not sure if there's a more reasonable way to do this than e3f6222e52824a3d1a912246d72a754189f1659d (i have mixed feelings about `expect`ing vs erroring, there)

src/types/fingerprint.rs
hko-s: cab9dc33bcff7e214cc131b306da529002895cde

src/packet/signature/config.rs
```diff
@@ -24,42 +22,185 @@ pub struct SignatureConfig {
...
+        match value {
+            SignatureVersionSpecific::V3 { .. } => SignatureVersion::V3,
+            SignatureVersionSpecific::V4 => SignatureVersion::V4,
+            SignatureVersionSpecific::V6 { .. } => SignatureVersion::V6,
+        }
+    }
 }
 
 impl SignatureConfig {
+    pub fn v3(
```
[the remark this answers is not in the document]
hko-s: cab9dc33bcff7e214cc131b306da529002895cde

src/composed/message/types.rs
```diff
@@ -433,34 +438,57 @@ impl Message {
...
-                    algorithm,
-                    hash_algorithm,
-                    hashed_subpackets,
-                    unhashed_subpackets,
-                );
-                (typ, signature_config.sign(key, key_pw, l.data())?)
+                let mut config = match sig_version {
+                    SignatureVersion::V4 => SignatureConfig::v4(typ, algorithm, hash_algorithm),
+                    SignatureVersion::V6 => {
+                        SignatureConfig::v6_with_rng(rng, typ, algorithm, hash_algorithm)?
```
dignifiedquire, line 444: not a fan of this name, maybe make `v6` by default take an `rng` and make a `v6_with_salt`?

src/composed/message/types.rs
```diff
@@ -433,34 +438,57 @@ impl Message {
...
+                let signature = config.sign(key, key_pw, &data[..])?;
 
                 (typ, signature)
             }
         };
-        let ops = OnePassSignature::from_details(typ, hash_algorithm, algorithm, key_id);
+
+        let ops = match sig_version {
+            SignatureVersion::V4 => {
+                OnePassSignature::from_details_v3(typ, hash_algorithm, algorithm, key_id)
```
dignifiedquire, line 475: given that the others are called `v4()` `v6()` etc, should we call this just `v3()`?

src/composed/message/types.rs
```diff
@@ -433,34 +438,57 @@ impl Message {
...
+            }
```
dignifiedquire, line 476: should we check for the matching config version here too?

src/packet/key/public.rs
```diff
@@ -394,7 +514,27 @@ impl PublicKeyTrait for PubKeyInner {
         self.to_writer(&mut key_buf)?;
 
         // old style packet header for the key
-        writer.write_all(&[0x99, (key_buf.len() >> 8) as u8, key_buf.len() as u8])?;
+        match self.version() {
+            KeyVersion::V2 | KeyVersion::V3 | KeyVersion::V4 => {
+                // When a v4 signature is made over a key, the hash data starts with the octet 0x99,
+                // followed by a two-octet length of the key, and then the body of the key packet.
+                writer.write_all(&[0x99])?;
```
dignifiedquire, line 521: isn't there `write_u8`?

src/packet/signature/config.rs
```diff
@@ -2,64 +2,181 @@ use std::io::Read;
...
+            bail!("Unknown v6 signature salt length for hash algorithm {hash_alg:?}");
+        };
+
+        let mut salt = vec![0; salt_len];
+        rng.fill_bytes(&mut salt);
+
+        Ok(salt)
+    }
+
+    pub fn v6_with_rng<R: CryptoRng + Rng>(
```
dignifiedquire, line 142: missing docs

src/types/params/plain_secret.rs
```diff
@@ -328,6 +360,47 @@ impl PlainSecretParams {
...
+                aead_mode,
+                s2k,
+                nonce,
+            } => {
+                let key = s2k.derive_key(passphrase, sym_alg.key_size())?;
+
+                let enc_data = match version {
+                    KeyVersion::V2 => unsupported_err!("Encryption for V2 keys is not available"),
+                    KeyVersion::V3 => unimplemented_err!("v3 encryption"),
+                    KeyVersion::V4 => unimplemented_err!("v4 aead encryption"), // FIXME: implement
```
dignifiedquire, line 380: don't we already have aead? or is this for the next round ?

src/types/params/plain_secret.rs
```diff
@@ -336,23 +409,23 @@ impl PlainSecretParams {
...
         writer.write_all(&[self.string_to_key_id()])?;
+
         let mut hasher = checksum::SimpleChecksum::default();
         {
             let mut tee = TeeWriter::new(&mut hasher, writer);
-            self.to_writer_raw(&mut tee)?;
+            self.as_ref().to_writer_raw(&mut tee)?;
+        }
+
+        if version != KeyVersion::V6 {
```
dignifiedquire, line 422: should this be more explicit checking for v3 and v4?

src/types/params/plain_secret.rs
```diff
@@ -389,3 +470,39 @@ fn rsa_secret_params(i: &[u8]) -> IResult<&[u8], PlainSecretParams> {
...
+    aead_mode: AeadAlgorithm,
+) -> Result<([u8; 32], Vec<u8>)> {
+    // HKDF to derive output keying material
+    let hk = Hkdf::<Sha256>::new(None, derived);
+    let mut okm = [0u8; 32];
+
+    let type_id = u8::from(secret_tag) | 0xc0;
+
+    // HKDF info parameter
+    let info = &[
```
dignifiedquire, line 491: this is a bit odd, I would skip the reference here, and only use it at call site

src/composed/key/builder.rs
hko-s: Opened https://github.com/rpgp/rpgp/issues/384 for encryption roundtripping, later

src/composed/message/types.rs
hko-s: good catch. done.

src/packet/key/public.rs
hko-s: I've made a branch that uses write_u8 a lot, starting from the code in this PR: https://github.com/hko-s/rpgp/pull/2

My suggestion would be that we merge that after this one.

src/types/params/plain_secret.rs
hko-s: key encryption with aead will be handled in the followup PR

src/packet/signature/ser.rs
hko-s: (footnote: I've caught a few more "as" casts, in the meantime, and fixed them in https://github.com/hko-s/rpgp/pull/2)

src/composed/message/types.rs
hko-s: yes, sounds good (71070ad0ff4fcaba297cc697fc6347cb30827e31)

src/composed/message/types.rs
hko-s: We create the `SignatureConfig` in the block above this one. The versions for both packets are chosen based on the key version. I don't see how the `SignatureConfig` and `Ops` versions could fall out of sync, here.

But maybe I'm misreading what you are suggesting?

Fwiw: v6 keys can only make v6 `Signature`/`Ops` packets (and v4 keys must always make v4 `Signature` and v3 `Ops` packets). So picking the versions is nicely straightforward.

hko-s: After staring at this code for a while, I think it's clearer to `match` directly on the signer key version, when picking signature versions. See a45b2dff72b2e984268bcfd8f3195bcb392b53f2

hko-s:
> add benchmarks for v6 keys and messages

:+1:  I opened https://github.com/rpgp/rpgp/issues/388 for this

src/types/packet.rs
dignifiedquire: yeah that looks good to me, totally fine to `expect` if you know that the versions are already checked at that point

dignifiedquire (approved):

src/types/packet.rs
hko-s: ftr: i replaced the expects (see 53214a9d9e54757b6372c37b8697000cb469e753) based on our brief discussion of the brittleness of this approach

PR #380, opened 2024-08-05
Add special error message for packet 20

### LINKED ISSUE #376: Improve error message about packet type 20
Add a short text that explains that we don't support LibrePGP's AEAD format, to help users to understand what's going on

DoumanAsh:
Fixes #376

### DIFF
--- a/src/packet/single.rs
+++ b/src/packet/single.rs
@@ -127,6 +127,9 @@ pub fn body_parser(ver: Version, tag: Tag, body: &[u8]) -> Result<Packet> {
         }
         Tag::ModDetectionCode => ModDetectionCode::from_slice(ver, body).map(Into::into),
         Tag::Padding => Padding::from_slice(ver, body).map(Into::into),
+        Tag::Other(20) => {
+            unimplemented_err!("GnuPG-proprietary 'OCB Encrypted Data Packet' is unsupported")
+        }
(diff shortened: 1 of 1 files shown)


src/packet/single.rs
```diff
@@ -127,6 +127,7 @@ pub fn body_parser(ver: Version, tag: Tag, body: &[u8]) -> Result<Packet> {
         }
         Tag::ModDetectionCode => ModDetectionCode::from_slice(ver, body).map(Into::into),
         Tag::Padding => Padding::from_slice(ver, body).map(Into::into),
+        Tag::Other(20) => unimplemented_err!("GnuPG AEAD is unsupported and not part of OpenGPG"),
```
hko-s, line 130:
```suggestion
        Tag::Other(20) => unimplemented_err!("GnuPG-prorietary 'OCB Encrypted Data Packet' is unsupported"),
```

hko-s: Thanks for adding a message, this is useful! I suggested a minor change, above, to reflect the exact name of packet type 20 in the draft text of GnuPG's format.

hko-s: Oops, thanks for catching my typo :relaxed:

dignifiedquire (approved):
thanks, just one more `cargo fmt` is missing

Issue #388, opened 2024-08-16
Add benchmarks for v6 keys and messages


hko-s:
(after encryption has landed)

PR #392, opened 2024-08-17
RFC 9580 encryption

hko-s:


src/composed/signed_key/secret.rs
```diff
@@ -234,8 +234,8 @@ impl PublicKeyTrait for SignedSecretKey {
         self.primary_key.verify_signature(hash, data, sig)
     }
 
-    fn encrypt<R: Rng + CryptoRng>(&self, rng: R, plain: &[u8]) -> Result<Vec<Mpi>> {
-        self.primary_key.encrypt(rng, plain)
+    fn encrypt<R: Rng + CryptoRng>(&self, rng: R, plain: &[u8], v6_esk: bool) -> Result<EskBytes> {
```
dignifiedquire, line 237: this is very unfortunte

src/crypto/aes_kw.rs
```diff
@@ -3,6 +3,10 @@ use generic_array::GenericArray;
 
 use crate::errors::Result;
 
+/// Default initial value (IV)
+/// https://www.rfc-editor.org/rfc/rfc3394#section-2.2.3.1
+const IV: [u8; 8] = [0xA6, 0xA6, 0xA6, 0xA6, 0xA6, 0xA6, 0xA6, 0xA6];
```
dignifiedquire, line 8: can be deleted

src/crypto/x25519.rs
```diff
@@ -46,3 +91,198 @@ pub fn generate_key<R: Rng + CryptoRng>(mut rng: R) -> (PublicParams, PlainSecre
...
+        let sk = SecretKey {
+            secret: long_lived_private,
+        };
+        let decrypted2 = sk.decrypt((ephemeral_key, public_key, &esk)).unwrap();
+
+        assert_eq!(decrypted_key, decrypted2);
+    }
+
+    // #[test]
+    // fn test_encrypt_decrypt() {
```
dignifiedquire, line 238: what about this test?

src/packet/public_key_encrypted_session_key.rs
```diff
@@ -47,61 +61,181 @@ impl PublicKeyEncryptedSessionKey {
...
+
+    // FIXME: helper fn to check if a Key matches with this PKESK's target (can we remove id/fp?)
+    // v3: keyid or wildcard match
+    // v6: fp or implicit wildcard match
+
+    pub fn match_identity(&self, key_id: &KeyId, fp: &Fingerprint) -> bool {
+        match self {
+            Self::V3 { id, .. } => {
+                let wildcard_id =
+                    KeyId::from_slice(&[0, 0, 0, 0, 0, 0, 0, 0]).expect("wildcard_id");
```
dignifiedquire, line 124: should be a function on `KeyId`, `fn is_wildcard(&self) -> bool`

src/packet/public_key_encrypted_session_key.rs
```diff
@@ -47,61 +61,181 @@ impl PublicKeyEncryptedSessionKey {
...
-        &self.id
+        match self {
+            Self::V3 { id, .. } => id,
+            _ => unimplemented!(), // FIXME
+        }
     }
 
-    pub fn mpis(&self) -> &[Mpi] {
-        &self.mpis
+    pub fn fp(&self) -> Option<&Fingerprint> {
```
dignifiedquire, line 149: expand to `fingerprint`?

src/packet/public_key_encrypted_session_key.rs
```diff
@@ -113,62 +247,172 @@ fn parse(
...
+                PublicKeyAlgorithm::RSA
+                | PublicKeyAlgorithm::RSASign
+                | PublicKeyAlgorithm::RSAEncrypt
+                | PublicKeyAlgorithm::Elgamal
+                | PublicKeyAlgorithm::ElgamalSign,
+                _, // FIXME!
+            ) => {
+                // FIXME: impl Elgamal
+                let EskBytes::Rsa { mpi } = self.values() else {
+                    unimplemented!("FIXME")
```
dignifiedquire, line 362: sad

src/packet/sym_encrypted_protected_data.rs
```diff
@@ -67,12 +67,120 @@ impl SymEncryptedProtectedData {
...
+        debug!("info: {} - hkdf: {}", hex::encode(info), hex::encode(okm));
+        let message_key = &okm[..sym_alg.key_size()];
+        let raw_iv_len = aead.nonce_size() - 8;
+        let iv = &okm[sym_alg.key_size()..sym_alg.key_size() + raw_iv_len];
+        let mut nonce = vec![0u8; aead.nonce_size()];
+        nonce[..raw_iv_len].copy_from_slice(iv);
+
+        debug!("message_key: {}", hex::encode(message_key));
+        debug!("iv: {}", hex::encode(iv));
+        debug!("nonce: {}", hex::encode(&nonce));
```
dignifiedquire, line 109: note: before merging please make a pass to not log any sensitive material at any log level

src/types/secret_key_repr.rs
```diff
@@ -19,47 +21,114 @@ pub enum SecretKeyRepr {
...
-                let key_size = session_key_algorithm.key_size();
-                (
-                    &decrypted_key[1..=key_size],
-                    &decrypted_key[key_size + 1..key_size + 3],
-                )
+                let key = priv_key.decrypt((ephemeral.to_owned(), *public, session_key))?;
+
+                // FIXME: no postprocessing for X25519 (especially: no checksum)
+
+                // We expect `algo` to be set for v3 PKESK, and unset for v6 PKESK
```
dignifiedquire, line 70: should be cross checked with flag

src/crypto/x25519.rs
hko-s: hadn't implemented it yet. it's there now: 8990cf99324d271e5a9144717a0025ecf6edd6f8

src/packet/public_key_encrypted_session_key.rs
hko-s: good point (ae28f112b8ea74d6ca7070bd330e9d08506a5e2c)

src/packet/public_key_encrypted_session_key.rs
hko-s: > sad

true. fixed in b737a345c42e746b4a73836f7a157d8bf373486c

src/packet/sym_encrypted_protected_data.rs
hko-s: right. i think 62ab81dd17bb79fdb79c8dc760e6c6b20560b3e9 drops all of them.

src/packet/public_key_encrypted_session_key.rs
```diff
@@ -113,62 +238,167 @@ fn parse(
...
+                    sym_alg,
+                    session_key,
+                },
+            ) => {
+                writer.write_all(ephemeral)?;
+
+                // Unlike the other public-key algorithms, in the case of a v3 PKESK packet,
+                // the symmetric algorithm ID is not encrypted.
+                //
+                // https://www.ietf.org/archive/id/draft-ietf-openpgp-crypto-refresh-13.html#name-algorithm-specific-fields-for-
```
hko-s, line 386: replace all draft links with RFC 9580 links

src/crypto/x448.rs
```diff
@@ -0,0 +1,261 @@
...
+    type KeyParams = ();
+
+    fn key_params(&self) {}
+}
+
+impl Decryptor for SecretKey {
+    // - Ephemeral X448 public key (56 bytes).
+    // - Recipient public key (56 bytes). [FIXME: use PublicParams and expect PublicParams::X448?]
+    // - Encrypted and wrapped session key.
+    type Data<'a> = ([u8; 56], [u8; 56], &'a [u8]);
```
dignifiedquire, line 28: would be nicer to make this at least a struct

src/crypto/x448.rs
```diff
@@ -0,0 +1,261 @@
...
+    use std::ops::Deref;
+
+    use rand::{RngCore, SeedableRng};
+    use rand_chacha::ChaChaRng;
+
+    use super::*;
+    use crate::types::SecretKeyRepr;
+
+    // #[test]
+    // fn x25519_hkdf() {
```
dignifiedquire, line 179: what about this?

hko-s:
> add fixtures for key and message tests for all the new combinations

src/crypto/x448.rs
hko-s: I refactored this a bit, but am very unsure what names I'd like, and if this is the structure we want: b877594112607c5c52a3aa9a88269fe035879247

src/composed/signed_key/secret.rs
hko-s: I made this slightly less horrible with an enum: 6cc62b1d00afc358dd26fe2094b2ff4acab92bbe ... but if we come up with better ideas, I'm all into iterating further.

src/composed/message/types.rs
```diff
@@ -397,12 +397,9 @@ impl Message {
     ) -> Result<Self> {
         let data = self.to_bytes()?;
 
-        let edata = Edata::SymEncryptedProtectedData(SymEncryptedProtectedData::encrypt_with_rng(
-            rng,
-            alg,
-            &session_key,
-            &data,
-        )?);
+        let edata = Edata::SymEncryptedProtectedData(
```
hko-s, line 400: - Add a check that the `esk` parameter contains `Esk` that are all consistently esk v3/4 or esk v6, and then
- use the appropriate seipd version

src/composed/message/decrypt.rs
```diff
@@ -3,26 +3,73 @@ use log::debug;
...
-        let (key, sym_alg) = priv_key.decrypt(mpis, locked_key.fingerprint().as_bytes())?;
-        // TODO: handle other versions
-        Ok(PlainSessionKey::V4 { key, sym_alg })
-    })
+    locked_key.unlock(key_pw, |priv_key| priv_key.decrypt(values, typ, locked_key))
+}
+
+/// Values comprising the encrypted session key
+///
+/// FIXME: extend for algorithm specific values? (and/or v3 vs. v6)
```
dignifiedquire, line 26: this sounds like a good plan, given they are differently structured depending on the version

src/composed/message/decrypt.rs
```diff
@@ -3,26 +3,73 @@ use log::debug;
...
 
 /// Decrypted session key.
+///
+/// A v4 session key can be used with a SED, or a v1 SEIPD.
+/// A v6 session key can only be used with a v2 SEIPD.
+///
+/// https://www.rfc-editor.org/rfc/rfc9580.html#name-packet-versions-in-encrypte
 #[derive(derive_more::Debug, Clone, PartialEq, Eq)]
 pub enum PlainSessionKey {
+    /// Obtained from a v3 PKESK or a v4 SKESK
```
dignifiedquire, line 72: is this correct `v3 PKESK or a v4 SKESK` ?

src/composed/signed_key/secret.rs
```diff
@@ -526,6 +526,8 @@ ruh8m7Xo2ehSSFyWRSuTSZe5tm/KXgYG
 
         msg.verify(&ssk.primary_key)?;
 
+        // FIXME: test roundtrip unlock/lock?
```
dignifiedquire, line 529: yes please

src/crypto/ecdh.rs
```diff
@@ -124,14 +126,20 @@ impl Decryptor for SecretKey {
...
-                derive_shared_secret_decryption::<p256::NistP256>(public_point, secret, &curve, 65)?
-            }
-            SecretKey::P384 { secret, .. } => {
-                derive_shared_secret_decryption::<p384::NistP384>(public_point, secret, &curve, 97)?
-            }
+            SecretKey::P256 { secret, .. } => derive_shared_secret_decryption::<p256::NistP256>(
+                data.public_point,
+                secret,
+                &curve,
+                65,
```
dignifiedquire, line 133: maybed make these constants

src/crypto/x448.rs
```diff
@@ -0,0 +1,220 @@
...
+    )
+}
+
+/// HKDF for X448
+/// https://www.rfc-editor.org/rfc/rfc9580.html#name-algorithm-specific-fields-for-x
+pub fn hkdf(
+    ephemeral: &[u8; 56],
+    recipient_public: &[u8; 56],
+    shared_secret: &[u8; 56],
+) -> Result<[u8; 32]> {
```
dignifiedquire, line 103: could this be shared with the hkdf in x25519 by passing in the info and sizes as generics/arguments?

src/packet/key/public.rs
```diff
@@ -429,6 +431,9 @@ impl PublicKeyTrait for PubKeyInner {
             PublicParams::X25519 { .. } => {
                 unimplemented_err!("verify X25519");
```
dignifiedquire, line 432: should this be unimplemented, or not available?

src/packet/public_key_encrypted_session_key.rs
```diff
@@ -11,34 +11,49 @@ use crate::crypto::checksum;
...
 /// <https://tools.ietf.org/html/rfc4880.html#section-5.1>
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct PublicKeyEncryptedSessionKey {
-    packet_version: Version,
-    version: u8,
-    id: KeyId,
-    algorithm: PublicKeyAlgorithm,
-    mpis: Vec<Mpi>,
+
+pub enum PublicKeyEncryptedSessionKey {
```
dignifiedquire, line 26: docs

src/packet/public_key_encrypted_session_key.rs
```diff
@@ -47,61 +62,200 @@ impl PublicKeyEncryptedSessionKey {
...
+            }
+            _ => false,
+        }
+    }
+
     pub fn id(&self) -> &KeyId {
-        &self.id
+        match self {
+            Self::V3 { id, .. } => id,
+            _ => unimplemented!(), // FIXME
```
dignifiedquire, line 137: what's missing here?

src/packet/public_key_encrypted_session_key.rs
```diff
@@ -47,61 +62,200 @@ impl PublicKeyEncryptedSessionKey {
...
+            _ => None,
+        }
     }
 
-    pub fn packet_version(&self) -> Version {
-        self.packet_version
+    pub fn values(&self) -> &EskBytes {
+        match self {
+            Self::V3 { values, .. } | Self::V6 { values, .. } => values,
+            _ => unimplemented!(),
```
dignifiedquire, line 151: not panic, but return an actual error?

src/packet/public_key_encrypted_session_key.rs
```diff
@@ -47,61 +62,200 @@ impl PublicKeyEncryptedSessionKey {
...
+        }
+    }
+
+    pub fn algorithm(&self) -> PublicKeyAlgorithm {
+        match self {
+            Self::V3 { pk_algo, .. } | Self::V6 { pk_algo, .. } => *pk_algo,
+            _ => unimplemented!(),
```
dignifiedquire, line 158: not panic, but return an actual error?

src/packet/sym_encrypted_protected_data.rs
```diff
@@ -67,12 +66,103 @@ impl SymEncryptedProtectedData {
...
+    }
+
+    /// Same as [`encrypt_with_rng_seipdv1`], but uses [`thread_rng`] for RNG.
     ///
-    /// [`encrypt_with_rng`]: SymEncryptedProtectedData::encrypt_with_rng
+    /// [`encrypt_with_rng_seipdv1`]: SymEncryptedProtectedData::encrypt_with_rng_seipdv1
     /// [`thread_rng`]: rand::thread_rng
-    pub fn encrypt(alg: SymmetricKeyAlgorithm, key: &[u8], plaintext: &[u8]) -> Result<Self> {
-        Self::encrypt_with_rng(&mut thread_rng(), alg, key, plaintext)
+    pub fn encrypt_seipdv1(
```
dignifiedquire, line 160: while we are renaming these, lets drop the convenience methods, and do only the ones that take in an rng

src/packet/sym_key_encrypted_session_key.rs
```diff
@@ -232,6 +229,52 @@ impl SymKeyEncryptedSessionKey {
             encrypted_key: Some(encrypted_key),
         })
     }
+
+    pub fn encrypt6<F, R: CryptoRng + Rng>(
```
dignifiedquire, line 233: maybe `encrypt_v6`?

src/packet/sym_key_encrypted_session_key.rs
```diff
@@ -232,6 +229,52 @@ impl SymKeyEncryptedSessionKey {
...
+        let mut iv = vec![0; aead.iv_size()];
+        rng.fill_bytes(&mut iv);
+
+        // AEAD encrypt
+        let mut encrypted_key = session_key.to_vec();
+        let auth_tag =
+            aead.encrypt_in_place(&sym_algorithm, &okm, &iv, &info, &mut encrypted_key)?;
+
+        Ok(SymKeyEncryptedSessionKey::V6 {
+            packet_version: Default::default(),
```
dignifiedquire, line 269: default is this correct?

src/types/key_id.rs
```diff
@@ -24,6 +24,10 @@ impl KeyId {
     pub fn to_vec(&self) -> Vec<u8> {
         self.0.to_vec()
     }
+
+    pub fn is_wildcard(&self) -> bool {
```
dignifiedquire, line 28: doc

src/types/secret_key_repr.rs
```diff
@@ -16,50 +18,152 @@ pub enum SecretKeyRepr {
...
+                return if let Some(sym_alg) = *sym_alg {
+                    Ok(PlainSessionKey::V4 { key, sym_alg })
+                } else {
+                    Ok(PlainSessionKey::V6 { key })
+                };
             }
+
+            (SecretKeyRepr::EdDSA(_), _) => bail!("EdDSA is only used for signing"),
+            _ => todo!(),
```
dignifiedquire, line 115: nicer error ;)

hko-s:
- can legacy packet format be combined with v6 artifacts? (make tests that pin down what we want to support)
- add tests that check we don't allow v6 keys that use "legacy" cryptographic mechanisms

src/composed/message/decrypt.rs
hko-s: I think so. For some historical reason, there is no v4 PKESK :shrug:

- PKESK was at v3 in RFC 2440/4880, and now went to v6.
- SKESK was at v4 in RFC 2440/4880, and now went to v6.

src/types/secret_key_repr.rs
hko-s: 23aaabfd

src/composed/message/decrypt.rs
hko-s: In the newly defined type `EskType` I've called one variant `V3_4` (read: V3 "PKESK or V4 SKESK").
If we are happy with this (kind of weird) name, then we could also rename `PlainSessionKey::V4` to `V3_4` for consistency?

Whatever we do, these versions are just weird. I guess there is no super clean and neat way to name them.

src/composed/signed_key/secret.rs
hko-s: f842c230fca89376a00a37adff82b7da10edd216

src/crypto/x448.rs
hko-s: I think it can. But the shared function would also need to take a `Digest` as parameter (`Sha256` vs. `Sha512`).

I'm sure this is possible, but I don't know how. And I wonder about the const-generics complexity vs. DRY-ness tradeoff.

src/packet/sym_key_encrypted_session_key.rs
hko-s: `PacketVersion` defaults to `New`. That's definitely what we want generate for V6 artifacts (and really, always).

src/composed/message/decrypt.rs
dignifiedquire: sounds all horrible, I guess `V3_4` helps a bit, to highlight you need to look at the docs

src/packet/sym_key_encrypted_session_key.rs
hko-s: I added a bit of documentation in 08e6e9bfceea1df71624340b85c56f308e90b20f to make a better connection between the terminology in RFC 9580 and the `Version` enum.

src/composed/message/decrypt.rs
hko-s: changed in 401f356fc138c1572066e211389357f669effc35, and i added a bit of docs

src/packet/key/public.rs
hko-s: Right, this should be "not available".

The errors around this one slowly all bitrotted in this wrong direction :sweat_smile: 
I normalized a number of errors, in fabe4c9b9567e99b907c137167563c569ffaad9a.

src/packet/public_key_encrypted_session_key.rs
hko-s: :+1: 17d1ba5f346019f8d2f3ce61e199c81b3082b13c

src/packet/public_key_encrypted_session_key.rs
```diff
@@ -47,61 +62,200 @@ impl PublicKeyEncryptedSessionKey {
...
-            algorithm: pkey.algorithm(),
-            mpis,
+            pk_algo: pkey.algorithm(),
+            values,
+        })
+    }
+
+    /// Encrypts the given session key to the passed in public key, as a v6 pkesk.
+    /// FIXME: cleanup/DRY with from_session_key
+    pub fn from_session_key6<R: CryptoRng + Rng>(
```
hko-s, line 91: rename to `_v6`

src/composed/message/decrypt.rs
hko-s: (Mostly independently, I just renamed `EskBytes` to `PkeskBytes` and moved it to `crate::types` in 6b66d8827f1913e253ff7d7618f0c28e6d6e7d14.
Maybe it should live in `crate::packet::public_key_encrypted_session_key`, though?)

That aside, I've thought about how to improve the shape of the enum a number of times now, and can't come up with a significant improvement.

I think the current enum models the algorithm-specific data reasonably.

The main oddity is that `SymmetricKeyAlgorithm` is an optional field for the `X25519` and `X448` variants. That field is only present for v3 PKESK, not for v6 PKESK.
Maybe the field name could/should be improved to make it clearer that this optionality is version-specific (currently only the docs explain this, not the field names).

Unless some epiphany strikes me unexpectedly, for now I'll give up trying to improve the enum, but would be very happy if you have ideas.

src/crypto/ecdh.rs
hko-s: Agreed, but I'd like to do that in a separate PR, to keep this PR somewhat focused.
I opened #402 to track our intention to improve the constants situation.

src/composed/message/types.rs
hko-s: Note to self: renamed the existing functions to `*_seipdv1` to make it explicit that they produce SEIPDv1 encrypted data packets.

Adding analogous mid-level APIs that produce SEIPDv2 encrypted data packets is a separate task.

src/packet/public_key_encrypted_session_key.rs
hko-s: 2e94501ed16eea209b9e395e5845ebca6be436f5

src/packet/public_key_encrypted_session_key.rs
hko-s: error handling was missing. fixed.

src/packet/sym_encrypted_protected_data.rs
hko-s: I've dropped all convenience methods that produce an implicit Rng, and also dropped all `with_rng` fragments from function names (934257b9e05a55fbce639612ed2b94928f74f07a).

(I think this aligns all function names with what we've done for v6 Signatures in the previous PR.)

src/packet/sym_key_encrypted_session_key.rs
hko-s: :+1:  8412305e8be6b7e27cd0631e8669caabaabe310b

hko-s:
> * can legacy packet format be combined with v6 artifacts? (make tests that pin down what we want to support)

RFC 9580 doesn't contain any hard requirements about the packet format. It strongly prefers generating the new packet format (which we do). So I don't think there's anything we can/should test for, regarding the packet format.

src/crypto/x448.rs
dignifiedquire: it's fine for now, maybe just add a todo to explore this for now

src/composed/message/decrypt.rs
dignifiedquire: seems fine for now

src/crypto/x448.rs
dignifiedquire: structure looks good to me

hko-s:
>     * add tests that check we don't allow v6 keys that use "legacy" cryptographic mechanisms

814b6eda0a2fe4bab375ec4692f47b481acb241d implements rejection of v6 keys that use "legacy" curve 25519 formats.
61671703503dda29d095917e5abf88af706a8570 contains the test rfc9580_legacy_25519_illegal_in_v6 that verifies this works.

hko-s:
> > add fixtures for key and message tests for all the new combinations

61671703503dda29d095917e5abf88af706a8570 and 08c779617ca576b17e9b24eed2fc2bc0ade264e2 contain keys and messages that exercise all new combinations of key types and encryption/data signatures that i could think of.

The tests do:
- decryption of seipdv1/v2 fixtures
- verification of a v6 cleartext signed message fixtures
- roundtripping of seipdv1 and seipdv2 encryption/decryption
- roundtripping of CSF and inline signature creation/verification

For two sets of keys (see https://github.com/rpgp/rpgp/blob/08c779617ca576b17e9b24eed2fc2bc0ade264e2/tests/rfc9580.rs#L17-L31). Each key comes with test-message fixtures.

For brevity/clarity I'm not including pre-existing combinations of key and message type: I see no point in testing seipdv1 decryption for a traditional v4 key, in this context. Or dealing with v4 signatures (except for the new Ed25519 signature format, which gets tested for signatures with a v4 key as well).

However, with regular "old-style" v4 keys, only seipdv2 decryption and encryption roundtripping is tested.

src/composed/message/types.rs
```diff
@@ -1061,25 +1061,28 @@ mod tests {
...
-                    &[&pkey][..],
-                )
-                .unwrap();
 
-            let armored = encrypted.to_armored_bytes(None.into()).unwrap();
-            fs::write("./message-x25519.asc", &armored).unwrap();
+        for aead in [AeadAlgorithm::Ocb, AeadAlgorithm::Eax, AeadAlgorithm::Gcm] {
+            for sym in [
+                SymmetricKeyAlgorithm::AES128,
+                // SymmetricKeyAlgorithm::AES192, // TODO: not currently supported
```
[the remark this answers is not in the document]
hko-s: That also surprised me, when i added this test :sweat_smile: 

AES192 was not implemented in https://github.com/rpgp/rpgp/blob/main/src/crypto/aead.rs.
When I wrote this test, I thought this AES192 can't be supported without changed to RustCrypto. However, now that I took a more serious look, I see that this is incorrect.

I added AES192 support in 9f2cb1174a0c902ed7790d080e6eed31abe47ed8, and enabled these tests.

I'm thinking of producing SEIPDv2 encrypted test-messages for all permutations of Eax, Ocb, Gcm x AES128/192/256. And I'm thinking it might be good to make those the payload big enough that the messages will be broken into at least two chunks (given a small chunk size that i can set when generating those messages). Does that sound reasonable/worthwhile?

benches/benchmarks/message.rs
```diff
@@ -59,7 +59,7 @@ fn bench_message(c: &mut Criterion) {
     for size in &sizes {
         g.throughput(Throughput::BytesDecimal(*size as u64));
         g.bench_with_input(
-            BenchmarkId::new("encrypt_password_s2k_iter_aes128", size),
+            BenchmarkId::new("encrypt_password_s2k_iter_aes128_seipdv1", size),
```
dignifiedquire, line 62: what about benchmarks for seipdv2?

src/composed/message/decrypt.rs
```diff
@@ -3,35 +3,51 @@ use log::debug;
...
+    /// (Note: for historical reasons, the OpenPGP format doesn't specify a v4 PKESK or a V3 SKESK)
+    V3_4 {
         sym_alg: SymmetricKeyAlgorithm,
         #[debug("..")]
         key: Vec<u8>,
     },
+
     V5 {
         #[debug("..")]
         key: Vec<u8>,
```
dignifiedquire, line 47: given that these are sensitive (all versions) we should wrap these into [`Zeroizing`](https://docs.rs/zeroize/latest/zeroize/struct.Zeroizing.html)

benches/benchmarks/message.rs
hko-s: There is #388, for this. I imagined we'd do that after landing this PR (but am open to doing it before, as well).

src/packet/public_key_encrypted_session_key.rs
```diff
@@ -13,95 +13,252 @@ use crate::crypto::sym::SymmetricKeyAlgorithm;
...
         data[0] = u8::from(alg);
         data[1..=len].copy_from_slice(session_key);
 
-        // and appended a checksum
-        BigEndian::write_u16(
-            &mut data[len + 1..],
-            checksum::calculate_simple(session_key),
-        );
+        // Append a checksum, except for x25519/x448
+        // FIXME: factor this difference out and up?
```
dignifiedquire, line 75: what about this?

src/packet/public_key_encrypted_session_key.rs
```diff
@@ -13,95 +13,252 @@ use crate::crypto::sym::SymmetricKeyAlgorithm;
...
+                    true // wildcard always matches
+                }
+            }
+            _ => false,
+        }
     }
 
-    pub fn packet_version(&self) -> Version {
-        self.packet_version
+    pub fn id(&self) -> Result<&KeyId> {
```
dignifiedquire, line 137: docs missing for this and following methods

src/packet/sym_encrypted_protected_data.rs
```diff
@@ -53,26 +52,105 @@ impl SymEncryptedProtectedData {
...
+    /// Encrypts the data using the given symmetric key.
+    pub fn encrypt_seipdv2<R: CryptoRng + Rng>(
+        mut rng: R,
+        sym_alg: SymmetricKeyAlgorithm,
+        aead: AeadAlgorithm,
+        chunk_size: u8,
+        session_key: &[u8],
+        plaintext: &[u8],
+    ) -> Result<Self> {
+        // TODO: DRY decrypt/encrypt code duplication?
```
dignifiedquire, line 78: what about this?

src/packet/sym_encrypted_protected_data.rs
```diff
@@ -53,26 +52,105 @@ impl SymEncryptedProtectedData {
...
+        let mut okm = [0u8; 42];
+        hk.expand(&info, &mut okm).expect("42");
+
+        let message_key = &okm[..sym_alg.key_size()];
+        let raw_iv_len = aead.nonce_size() - 8;
+        let iv = &okm[sym_alg.key_size()..sym_alg.key_size() + raw_iv_len];
+        let mut nonce = vec![0u8; aead.nonce_size()];
+        nonce[..raw_iv_len].copy_from_slice(iv);
+
+        let mut out = Vec::new();
```
dignifiedquire, line 106: can we preallocate this with a known size?

benches/benchmarks/message.rs
dignifiedquire: ah right, I forgot about it 😅 can land later

src/composed/message/decrypt.rs
hko-s: right.

i think 083274ed7d570bc77e10887393917c8fce4f4778 does the right thing, and is sufficient?

src/packet/public_key_encrypted_session_key.rs
hko-s: Looking into this sent me down a surprising rabbit-hole of zeroizing and DRY-ing: 2fb88ac6a3863f9cfd2d9ad78842861aa2c86b36

Please check if all of these changes looks reasonable :sweat_smile:

src/packet/public_key_encrypted_session_key.rs
hko-s: right, wrote docs in bb714ecdbe4d5d0ce7fb169a651969cf8257ab6b

src/packet/sym_encrypted_protected_data.rs
hko-s: Yes. I think I've done the right math in e18252c891bc988701a30508577cfd3b5ce1fcd5.

It's just a bit annoying that this calculation is pretty convoluted and takes up a lot of space (putting cognitive load on future readers), so I tried my best to comment what's going on.

(But if we switch to streaming processing, that whole allocation will go away, anyway :smiley:)

hko-s: (I'll do the equivalent preallocation for the decryption allocation, in the DRY [comment above](https://github.com/rpgp/rpgp/pull/392#discussion_r1760973235))

hko-s: I added a `debug_assert_eq` that the allocation length calculation matches the final len of `out`.

I'm unsure if that's reasonable. Nut it seemed silly if the preallocation calculation was wrong, and we had no chance to ever notice. Would you find a WARN (?) log output appropriate for this? (Rather than panicing in debug mode)

src/packet/sym_encrypted_protected_data.rs
hko-s: I've DRY'd this in aa47208034f1d42cc80e87ec6d99ca2b1c153845.

(Not sure about the name and return-type of the internal helper function, though :sweat:)

src/composed/message/decrypt.rs
dignifiedquire: looks good

dignifiedquire (approved):
🎉

PR #393, opened 2024-08-19
feat: Make `SignatureConfig::hash_signature_data` more flexible

wiktor-k:
Hashers from RustCrypto implement `std::io::Write` and making use of that trait in the `hash_signature_data` makes it possible to use other hashers which do not implement RustCrypto Hasher trait.

Additionally it allows using hashers which may fail during the digest update (such as OpenSSL hashers).

(just for the context I'm using this in my [hasher-test](https://github.com/wiktor-k/hasher-test?tab=readme-ov-file#hasher-test) repository and the hasher I'm passing happens to be from a new `-pre` version of Rust Crypto's sha2 crate. Due to versioning for Rust it's not the same Hasher trait but the same Write trait).

### DIFF
--- a/src/packet/signature/config.rs
+++ b/src/packet/signature/config.rs
@@ -182,7 +182,7 @@ impl SignatureConfig {
         }
 
         self.hash_data_to_sign(&mut *hasher, data)?;
-        let len = self.hash_signature_data(&mut *hasher)?;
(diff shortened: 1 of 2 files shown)


dignifiedquire (approved):
seems fine to me, given that we don't rely on anything else here

PR #396, opened 2024-08-30
Ignore another sks-dump test failure

dkg:
If the submodule hasn't been cloned, then test_many_parsers will also fail, so ignore this failing test like the other ones that depend on the sks dump.


dignifiedquire:
thanks

PR #397, opened 2024-08-30
Relax dependencies

dkg:
I've tested the builds with these marginally older versions of the dependencies, and the test suites pass without a problem.

These older versions are in debian unstable, so it makes it slightly easier to build rpgp for debian if the dependencies are looser.


dkg:
if you have any reason to believe that something in the version ranges indicated should *not* be used with rpgp (even if the tests all pass) please let me know!

dignifiedquire (approved):
cargo deny is happy, so should be fine

Issue #401, opened 2024-09-04
MPI interface is too easy to misuse


wiktor-k:
`Mpi` struct has several constructors: `from_raw` and `from_slice` and (at least to me) it's not immediately clear that one of them has a hidden footgun: leading zeros may result in invalid MPIs being generated.

The problem is doubly annoying that the user may not be aware that they're using the wrong function because most of the time they would pass a non-zero-padded slices but just rarely get a wrong thing (depending on what the underlying crypto library generates). (We found that in signstar by running the test suite 100 times over and finding one broken signature).

IMO the non-zero-stripping constructor should be more clearly marked as being dangerous. Maybe even more: since it can be used to construct malformed MPIs maybe it should be named `..._unchecked` and marked unsafe?

The problem is exacerbated by the [`bit_size`](https://docs.rs/pgp/latest/src/pgp/util.rs.html#67-73) function which is non-spec-compliant. Quoting [RFC 9580](https://www.rfc-editor.org/rfc/rfc9580.html#name-using-mpis-to-encode-other-):

> 2 octets of length in bits counted from the *first non-zero bit*

The current implementation assumes that the first non-zero bit will be in the first byte of the passed slice but this may not be true if the slice has leading zero *bytes*. (The function could be adjusted to look for first non-zero byte instead of `val[0]`).

The root of the problem is the underspecification of what happens with MPIs when there are zero-bytes in [RFC 4880](https://www.rfc-editor.org/rfc/rfc4880#section-3.2) (the new spec disallows leading zero bytes) and rpgp seems to assume that they are not allowed (`bit_size` and `mpi` parsing function seems to assume that). But if that is so then the MPI constructor function should *not* let users construct MPIs which then cannot be used by the library itself.

I think the following changes are a good idea:
  - remove or rename/mark unsafe functions that make it possible to construct invalid MPIs (with leading zero bytes) to adjust it to RFC 9580,
  - fix `bit_size` to return correct number of bits in case the input slice has a zero-byte OR document that the input needs to be a valid MPI,
  - consider adjusting `mpi` parsing function to properly support MPIs with leading zeros for implementation that are RFC 4880-compliant but not 9580-compliant.

Phew, okay, I guess this is clear now :sweat_smile: 

If there are any other good ideas I'm happy to hear them!

CC @hko-s @dvzrv

[...]

PR #410, opened 2024-09-18
fix: Remove the MPI constructor which may be misused

wiktor-k:
This patch removes the MPI constructor which may be used to construct invalid MPIs. It also adjusts the `From<u8>` implementation to use a safe variant.

@hko-s please see if this is what we discussed. Thanks!

Fixes: https://github.com/rpgp/rpgp/issues/401

### DIFF
--- a/src/types/mpi.rs
+++ b/src/types/mpi.rs
@@ -21,11 +21,12 @@ const MAX_EXTERN_MPI_BITS: u32 = 16384;
 ///
 /// ```rust
 /// use pgp::types::mpi;
+/// use pgp::types::MpiRef;
 ///
 /// // Decode the number `1`.
 /// assert_eq!(
 ///     mpi(&[0x00, 0x01, 0x01][..]).unwrap(),
-///     (&b""[..], (&[1][..]).into())
+///     (&b""[..], MpiRef::from_slice(&[1][..]))
 /// );
 /// ```
 pub fn mpi(input: &[u8]) -> IResult<&[u8], MpiRef<'_>> {
@@ -43,7 +44,7 @@ pub fn mpi(input: &[u8]) -> IResult<&[u8], MpiRef<'_>> {
             Err(needed) => Err(nom::Err::Incomplete(needed)),
             Ok(index) => {
                 let (rest, n) = number.take_split(index);
-                let n_stripped: MpiRef<'_> = strip_leading_zeros(n).into();
+                let n_stripped: MpiRef<'_> = MpiRef::from_slice(strip_leading_zeros(n));
 
                 Ok((rest, n_stripped))
             }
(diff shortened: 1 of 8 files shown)


src/types/mpi.rs
```diff
@@ -157,7 +153,7 @@ impl<'a> Serialize for MpiRef<'a> {
 
 impl From<&[u8]> for Mpi {
```
dignifiedquire, line 154: I wonder if we should remove this entirely and force direct calls instead

dignifiedquire: also this is problematic, as this now behaves differently to `From<Vec<u8>>` as well as the `From` variants for `MpiRef`

wiktor-k: I think it's a good idea to remove both `From`s since it's not a simple conversion.

hko-s: Removing all `From` implementations sounds potentially right to me. We could try out how that would look, in this PR.

wiktor-k: Maybe *all* is a bit too much since there are many (e.g. for `BigUint`). The other approach would be having `from_slice` but inside stripping leading zeros (basically renaming `from_raw_slice` to `from_slice`).

dignifiedquire (approved):
this is a good start, we can always remove/change more in follow ups

wiktor-k:
> this is a good start, we can always remove/change more in follow ups

I'm wondering about two further changes here:
  - rename `Mpi::from_raw_slice` to `Mpi::from_slice` since it sounds more idiomatic and the name is free now :)
  - make `MpiRef::from_slice` internally apply `strip_leading_zeros` to the input.

The docs of both `Mpi` and `MpiRef` both indicate that "The inner value is ready to be serialized, without the need to strip leading zeros" and making that constructor strip it, it would ensure that this invariant always holds.

If that doesn't sound like a good idea I'd propose alternative design: documenting that `MpiRef::from_slice` MUST be called with trailing zeros stripped (to preserve the docs invariant).

hko-s:
> > this is a good start, we can always remove/change more in follow ups
> 
> I'm wondering about two further changes here:
> 
>     * rename `Mpi::from_raw_slice` to `Mpi::from_slice` since it sounds more idiomatic and the name is free now :)
> 
>     * make `MpiRef::from_slice` internally apply `strip_leading_zeros` to the input.
> 
> 
> The docs of both `Mpi` and `MpiRef` both indicate that "The inner value is ready to be serialized, without the need to strip leading zeros" and making that constructor strip it, it would ensure that this invariant always holds.
> 
> If that doesn't sound like a good idea I'd propose alternative design: documenting that `MpiRef::from_slice` MUST be called with trailing zeros stripped (to preserve the docs invariant).

I like the idea to make the `Mpi` type guarantee this invariant (and documenting it in the API docs).
Removing the "unchecked" constructors sounds very attractive to me.

Alternatively: keeping unchecked constructors around, but giving them scarier names, possibly limiting their visibility in the API, and documenting what expectations these put on the caller sounds acceptable to me as well.

(I'm not super worried about the performance of always checking, but we could compare benchmark results of always checking in all `Mpi` constructors?)

hko-s:
(Closing this in favor of #416, to steal ~50% of the glory from @wiktor-k)

PR #411, opened 2024-09-18
Misc minor RFC 9580 work

hko-s:


src/packet/signature/config.rs
```diff
@@ -412,6 +414,16 @@ impl SignatureConfig {
                 let mut hashed_subpackets = Vec::new();
                 for packet in &self.hashed_subpackets {
                     debug!("hashing {:#?}", packet);
+
+                    // If a subpacket is encountered that is marked critical but is unknown to the
+                    // evaluating implementation, the evaluator SHOULD consider the signature to be
```
dignifiedquire, line 419: should is not must, so I am not sure we should do this

dignifiedquire (approved):

src/packet/signature/config.rs
hko-s: The argument for failing is:

> "[..] The purpose of the critical bit is to allow the signer to tell an evaluator that it would prefer a new, unknown feature to generate an error rather than being ignored."

(From https://www.rfc-editor.org/rfc/rfc9580.html#section-5.2.3.7-8)

I don't have extremely strong feelings about keeping this change in, but think it's better to keep than in:

If rPGP doesn't know a subpacket type, an application built on top of rPGP has no reasonable way to interpret this subpacket, which the signer explicitly marked as "if you don't understand this subpacket, please fail validation".

dignifiedquire: yeah that's fair, still unfortunate that it is should...

hko-s: I agree that the RFC is unfortunate and overly fuzzy there.

This whole subpacket/criticality business is from RFC 2440 (in 1998): https://www.rfc-editor.org/rfc/rfc2440.html#section-5.2.3.1 ... clearly the folks who wrote that text were into leaving a lot of the semantics fuzzy, and giving implementers much room to be very smart (tm).

I'll just imagine that the text says "MUST", here (as it probably should), and move on, for now :sweat_smile:

PR #416, opened 2024-09-23
fix: Remove the unchecked Mpi constructor (and other Mpi tweaking)

hko-s:
This is the same as #410, but rebased onto current main


hko-s:
I added two more commits as a suggestion:

- dropping the unchecked `impl From<Vec<u8>> for Mpi` (in favor of explicit calls for `from_raw`)
- dropping the unused `MpiRef::strip_trailing_zeroes`

dignifiedquire (approved):
thanks

PR #419, opened 2024-10-05
fix: make strip_leading_zeros_vec() work correctly

mrothNET:
Current implementation is wrong, it should have be `bytes.remove(0)`.
But removing each byte individually is sub optimal. The method `drain()` does the hard work instead.


dignifiedquire (approved):
thanks for catching this

PR #421, opened 2024-10-05
chore: remove unused and erratic end_of_line() fn

mrothNET:
- Looks like that is a typo left over from migrating `nom` to v7.
- Before migration, this was `alt!(input, eof!() | eol)`. 
- The migrated one looks wrongly (recursion eof).
- But anyway, this function is used nowhere, so remove it.


dignifiedquire:
thanks

PR #422, opened 2024-10-07
Relax dependencies

laniakea64:
Following up on https://github.com/rpgp/rpgp/pull/397#issuecomment-2378096286 :

The rationale for https://github.com/rpgp/rpgp/pull/397 only mentioned needing the _option_ of older dependencies to build for Debian using Debian-packaged Rust libraries.  It didn't mention any reason to _force_ downgrades e.g. to improve security or stability or address a bug.

This PR relaxes those dependency requirements to avoid unnecessary dependency downgrades relative to `pgp` 0.13.2 for projects building with crates directly from crates.io.

Tests and stress tests pass locally, and `cargo deny` is ok locally.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ rust-version = "1.75"
 [dependencies]
 aes = "^0.8"
 argon2 = "0.5"
-base64 = "0.21.7, < 0.23"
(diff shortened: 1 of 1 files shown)


dignifiedquire (approved):
thanks, makes sense

PR #425, opened 2024-10-21
chore: Optimize `Display` implementation for `BlockType`

wiktor-k:
This patch makes the `Display` implementation directly write static strings to the formatter instead of first allocating a `String` object. Call-sites have also been adjusted to take advantage of the formatter.

The `to_string` function is automatically attached by a `ToString` blanket implementation.


hko-s:
Looks straightforward enough to me :+1:

dignifiedquire (approved):

dignifiedquire:
thanks

PR #426, opened 2024-10-22
Gate x448 work behind an "unstable-curve448" feature.

dkg:
The x448 crate hasn't had a release in over 2 years, while several of its dependencies are still rather unstable (in particular, ed448-goldilocks and rand_core have had at least one semver-incompatible bump since the x448 release, and ed448-goldilocks itself probably needs another release as well, since there are many useful cleanup changes pending there).

This makes me think that x448 isn't quite ready for prime time in the existing rust ecosystem.  Given that ed448 itself is also not supported yet, it seems like any work with Curve448 might be premature to make available by default.

This patch proposes to put the x448 code behind a feature named "unstable-curve448".  It will also make it easier to land the rest of the RFC 9580 work in debian, which is currently struggling with the unmaintained dependency chain here.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -128,7 +128,7 @@ nightly = ["rsa/nightly", "rand/nightly", "num-bigint/nightly"]
 profile = ["gperftools"]
 asm = ["sha1/asm", "sha2/asm", "md-5/asm"]
 wasm = ["chrono/wasmbind", "getrandom", "getrandom/js"]
-unstable-curve448 = ["x448"]
+unstable-curve448 = ["dep:x448"]
 
 [profile.bench]
 debug = 2
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,7 +74,7 @@ eax = "0.5.0"
 ocb3 = "0.1"
 aes-kw = { version = "0.2.1", features = ["std"] }
 derive_more = { version = "1.0.0-beta.6", features = ["debug"] }
-x448 = "0.6"
+x448 = { version = "0.6", optional = true }
 
 [dependencies.buffer-redux]
 version = "1.0.0"
@@ -128,6 +128,7 @@ nightly = ["rsa/nightly", "rand/nightly", "num-bigint/nightly"]
 profile = ["gperftools"]
 asm = ["sha1/asm", "sha2/asm", "md-5/asm"]
 wasm = ["chrono/wasmbind", "getrandom", "getrandom/js"]
+unstable-curve448 = ["x448"]
 
 [profile.bench]
 debug = 2
--- a/README.md
+++ b/README.md
@@ -105,6 +105,17 @@ let public_key = SignedPublicKey::from_string(&key_string).unwrap().0;
 public_key.verify_signature(HashAlgorithm::default(), &data[..], &new_signature).unwrap();
 ```
 
+## Features
+
+### Experimental Cryptography
+
+Some cryptographic primitives are relatively new and under development.
+Those primitives are gated behind "unstable" features, so that they need to be explicitly enabled to make use of them.
+
+Currently, there is only one such feature:
+
+- The `unstable-curve448` enables public key encryption with the `x448` algorithm.
+
 ## Current Status
 
 > Last updated *September 2024*
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -170,6 +170,12 @@ jobs:
         target: ${{ matrix.target }}
         args: --benches
 
+    - name: check curve448
+      uses: actions-rs/cargo@v1
+      with:
+        command: test
+        args: --features unstable-curve448 --all --target ${{ matrix.target }}
+
     - name: tests
       uses: actions-rs/cargo@v1
       with:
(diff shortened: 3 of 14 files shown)


dkg:
I recognize that this might need a semver-incompatible version bump (to 0.15.0), if adopted, since the `x448` feature won't be enabled by default any longer.  Even if it were enabled by default, it might need a semver-incompatible version bump.

hko-s:
Thanks for digging into this!

I haven't looked at the details of the PR yet, but wanted to put on the record that I am in favor of the general direction:
It seems reasonable to enable feature-gating x448 out, while it is in a difficult-to-package-reasonably state.

I hope that x448 and ed448 will gain well maintained support (ideally under the rust crypto umbrella). But until then, considering x448 in rPGP to be of an experimental nature - and even to disable it by default - seems ok to me.

dignifiedquire:
need to add a CI check testing with the new feature, and should probably add some docs somewhere about it

Cargo.toml
```diff
@@ -71,7 +71,7 @@ eax = "0.5.0"
 ocb3 = "0.1"
 aes-kw = { version = "0.2.1", features = ["std"] }
 derive_more = { version = "1.0.0-beta.6", features = ["debug"] }
-x448 = "0.6"
+x448 = { version = "0.6", optional = true }
```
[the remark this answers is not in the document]
dignifiedquire: maybe calling feature `unstable-x448` might make more sense, to indicate its unifinishedness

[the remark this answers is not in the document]
dkg: I'm not sure i understand what you gain from having an explicit feature -- for example, when i build, cargo understands that there is a `gperftools` feature even though it is not explicitly named.  i'm happy to make this change, but can you point me to documentation that would help me understand it?   (i understand why you'd need it to give it a name like `unstable-x448`)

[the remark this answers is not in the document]
dkg: also, how about naming the explicit feature `unstable-curve448`, which would let us use the same feature to turn on `ed448` (in addition to `x448`) when that becomes available?

dkg:
I've added an explicit feature `unstable-curve448`.

> need to add a CI check testing with the new feature, and should probably add some docs somewhere about it

There are already tests  that are marked with the `x448` feature, so i've added a CI test to `.github/workflows/ci.yml`, which i think should do this by testing feature `unstable-curve448`.

I've also added a subsection to `README.md` describing the decision.

Let me know if you need anything else!

Cargo.toml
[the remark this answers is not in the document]
dkg: i think i've sorted this out now.  I understand it better after having read https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies

dkg:
I've adjusted the branch to now only create a single explicit feature `unstable-curve448`, and there is no more `x448` feature.

dkg:
@dignifiedquire this is a gentle nudge -- if you could review this and let me know if you think it needs anything else, i'd be grateful.

dkg:
I've just rebased and cleaned this up, please consider it for a future release!

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 All notable changes to rpgp will be documented in this file.
 
+## 0.15.0 - Unreleased
+
+### 🚜 Refactor
+
+- Move x448 work behind an `unstable-curve448` feature, off by default - ([#426](https://github.com/rpgp/rpgp/pull/426))
```
dignifiedquire, line 9: pleae remove this, changelog is auto generated when running the release process, based on the commit message

README.md
```diff
@@ -105,6 +105,15 @@ let public_key = SignedPublicKey::from_string(&key_string).unwrap().0;
 public_key.verify_signature(HashAlgorithm::default(), &data[..], &new_signature).unwrap();
 ```
 
+## Experimental Cryptography
```
dignifiedquire, line 108: I would prefer a section called `Features`, listing the availble features (with explanations), but no need for you to write them, I can do that as a follow up

CHANGELOG.md
dkg: I removed this section, though i don't know how to structure my commits so that they trigger the right kind of changelog update in the release process.  if you handle that, i'll look at your commits and try to emulate them in the future.

README.md
dkg: I've re-scoped this proposed section in `README.md`, but of course i don't mind if you want to modify it however you think is useful.

dignifiedquire (approved):

dkg:
thanks for the feedback, @dignifiedquire !  let me know if there's any other cleanup that you want.

README.md
dignifiedquire: thanks

dignifiedquire:
just need ci to become greeener, then this is good to go, sorry for the delay

dkg:
do you want me to take any action to make the CI "become greener"?  I'm not sure how to handle the cross-compilation or format "red" indicators, since i don't believe they are associated with the changes in this MR.  Let me know what i can do to move this along.  And, happy new year! :tada:

hko-s:
I think rebasing this branch on main might resolve (some of?) the CI issues

dignifiedquire:
I‘ll take it over tomorrow, don’t worry about it

dignifiedquire:
Finishing in https://github.com/rpgp/rpgp/pull/454

Issue #424, opened 2024-10-21
Problem verifying puppet release files


cyr:
I've been using this in an application to verify PGP signatures in debian repositories, which has been working well for the most part. But for some reason it does not verify Puppets repository release files, and I can't exactly tell why.

public key:
https://apt.puppetlabs.com/DEB-GPG-KEY-puppet-20250406

inline-signed cleartext message
https://apt.puppetlabs.com/dists/bookworm/InRelease

The id and fingerprints matches, but the verification fails here:

```rust
ensure_eq!(
    &self.signed_hash_value,
    &hash[0..2],
    "signature: invalid signed hash value"
);
```

Any idea why this would be? The gpg utility gives the signature an all clear.

hko-s:
Hello, and thanks for getting in touch!

I wonder if this might be a case where the "2-octet field holding the left 16 bits of the signed hash value" field ([see](https://www.rfc-editor.org/rfc/rfc9580.html#name-versions-4-and-6-signature-)) contains incorrect data, but the cryptographic signature is otherwise correct.

If this is the case, then commenting out the `ensure_eq` in rPGP (the one that you quote) should make the signature validation succeed.

(Assuming this fixes validation, we'd need to figure out if and how to adjust rPGP to allow callers to ignore the 2 bytes of hash digest in the signature packet.)

cyr:
I just tried commenting out that part and let it keep going, but it still stumbles and fails while verifying.

hko-s:
That is intriguing and somewhat troubling :sweat_smile: 

How does the failure look without the ensure?

cyr:
It SEEMS to reach this part:

`rust
crypto::rsa::verify(n.as_bytes(), e.as_bytes(), hash, hashed, sig[0].as_bytes())
`

and fails the verifications there.

hko-s:
Right. That sounds like rPGP's CSF code might be buggy, for this case.
Thanks again for the report, we'll need to dig deeper into this. I will try to do that, but can't promise super quick results.

Either way, this should be resolvable, and we should totally fix it.

Note (mostly) to self:

`cat InRelease | rsop inline-verify DEB-GPG-KEY-puppet-20250406`

reproduces the issue nicely.

wiktor-k:
Yup, I'm getting `error: pgp::errors::Error - "failed reading: cleartext body Error(Error { input: [79, 114, 105,` with this sample code:

```rust
use pgp::composed::cleartext::CleartextSignedMessage;
[...]

PR #427, opened 2024-10-23
fix CSF: strip trailing spaces/tabs during normalization for hashing

hko-s:
Added a test to debug #424


wiktor-k:
Hmm... maybe it'd be good to also attach a test case *without* the space to be trimmed? :thinking: 

Looks good to me. You could add "Fixes #424" in the commit message.

hko-s:
> Hmm... maybe it'd be good to also attach a test case _without_ the space to be trimmed? 🤔

We already have other CSF test messages that get tested. I think/hope those cover the base cases.
However, this case was totally missing.

Now that you mention it, though, I could add another test that does a sign/verify roundtrip on this message (the current test only does the "verify" side).

> Looks good to me. You could add "Fixes #424" in the commit message.

Right. Thanks, done!

src/composed/cleartext.rs
```diff
@@ -305,18 +305,35 @@ fn dash_escape(text: &str) -> String {
     out
 }
 
-/// Undo dash escaping of `text`.
+/// Undo dash escaping of `text`, and trim space/tabs at the end of lines.
 ///
-/// This implementation is implicitly agnostic between "\n" and "\r\n" line endings.
-fn dash_unescape(text: &str) -> String {
+/// This implementation can handle both "\n" and "\r\n" line endings.
+fn dash_unescape_and_trim(text: &str) -> String {
```
dignifiedquire, line 311: it would be good to have unit test for this behaviour, just testing this function

hko-s: good point. i've added a unit test for just the unescaper.

dignifiedquire (approved):

PR #428, opened 2024-10-23
test for possible buffer redux problem

hko-s:
In this new test `test_load_big_csf`, a (valid) CSF message can't be loaded with `CleartextSignedMessage::from_armor`.
However, the same message can be loaded fine with `Any::from_string`.

I wonder if the reason is that `CleartextSignedMessage::from_armor` adds a second layer of `buffer_redux` (the problem seems to happen suspiciously close to the 8k bytes mark, which seems to be a default buffering amount in `buffer_redux`)

This PR is a reminder to debug what's going wrong in the `test_load_big_csf` test, and fix it.

### DIFF
--- a/src/composed/cleartext.rs
+++ b/src/composed/cleartext.rs
@@ -617,6 +617,14 @@ mod tests {
     }
 
     #[test]
+    fn test_load_big_csf() {
+        let msg_data = std::fs::read_to_string("./tests/unit-tests/csf-puppet/InRelease").unwrap();
+
+        // FIXME: this fails to read -> buffer_redux problem!?
+        let (_msg, _) = CleartextSignedMessage::from_armor(msg_data.as_bytes()).unwrap();
+    }
+
+    #[test]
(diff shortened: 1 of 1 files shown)


dignifiedquire:
the issue with this testcase is two fold (a) there is a bug in the cleartext parser around handling incomplete inputs (b) we currently can't parse armor messages larger than the max buffer size, which is annyoing, and more complex to fix

hko-s:
Thanks for looking into this!

I'm worried that your enumeration doesn't seem to cover some of the buffered reader weirdnesses that we've seen, such as https://codeberg.org/heiko/rsop/issues/16

So unfortunately it seems the unit test in this PR doesn't cover all of the existing problems with the current buffered reader setup, after all :-/

dignifiedquire:
@hko-s I believe this can be closed now

hko-s:
> @hko-s I believe this can be closed now

Yes! Very happy that this is sorted out in main, now!

Issue #430, opened 2024-10-30
`pgp::composed::message::types::tests::test_*_encryption_seipdv2::{{closure}}` fails when the source tree is read-only


dkg:
Debian's autopkgtests run `cargo test` against a read-only copy of the crate.  for rpgp 0.14.0, [those tests fail](https://ci.debian.net/packages/r/rust-pgp/testing/amd64/).  this is similar to #403 but it doesn't appear to be the same code.  Here are the relevant errors:

```
1924s 
1924s failures:
1924s 
1924s ---- composed::message::types::tests::test_rsa_encryption_seipdv2 stdout ----
1924s thread 'composed::message::types::tests::test_rsa_encryption_seipdv2' panicked at src/composed/message/types.rs:1017:50:
1924s called `Result::unwrap()` on an `Err` value: Os { code: 13, kind: PermissionDenied, message: "Permission denied" }
1924s stack backtrace:
1924s    0: rust_begin_unwind
1924s              at /usr/src/rustc-1.81.0/library/std/src/panicking.rs:665:5
1924s    1: core::panicking::panic_fmt
1924s              at /usr/src/rustc-1.81.0/library/core/src/panicking.rs:74:14
1924s    2: core::result::unwrap_failed
1924s              at /usr/src/rustc-1.81.0/library/core/src/result.rs:1679:5
1924s    3: core::result::Result<T,E>::unwrap
1924s              at /usr/src/rustc-1.81.0/library/core/src/result.rs:1102:23
1924s    4: pgp::composed::message::types::tests::test_rsa_encryption_seipdv2
1924s              at ./src/composed/message/types.rs:1017:9
1924s    5: pgp::composed::message::types::tests::test_rsa_encryption_seipdv2::{{closure}}
1924s              at ./src/composed/message/types.rs:980:37
1924s    6: core::ops::function::FnOnce::call_once
1924s              at /usr/src/rustc-1.81.0/library/core/src/ops/function.rs:250:5
1924s    7: core::ops::function::FnOnce::call_once
1924s              at /usr/src/rustc-1.81.0/library/core/src/ops/function.rs:250:5
1924s note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
1924s 
1924s ---- composed::message::types::tests::test_x25519_encryption_seipdv2 stdout ----
1924s thread 'composed::message::types::tests::test_x25519_encryption_seipdv2' panicked at src/composed/message/types.rs:1081:65:
[...]

PR #431, opened 2024-10-30
Avoid writing to the crate source unnecessarily during tests

dkg:
See https://github.com/rpgp/rpgp/issues/430

Closes: #430


hko-s:
This PR looks good to me (and indeed in line with the changes in #403).

Sorry, I copied this writing pattern from #403, when I made those new tests. A naive grep makes me hopeful that those three occurrences are the only new writes I added inadvertently :crossed_fingers:

dignifiedquire (approved):

PR #433, opened 2024-11-15
Adds [`signature::Signer`] compatibility wrappers

baloo:
This adds rsa and ecdsa wrappers to provide pgp operations with a key kept in an external storage.


baloo:
This is a very rough first cut, but I'd love to know if this is something we could add here.

Thanks!

src/helper/ecdsa.rs
```diff
@@ -0,0 +1,223 @@
...
+        // TODO: check digest size
+        let signature = self.inner.sign_prehash(prehash)?;
+        let (r, s) = signature.split_bytes();
+        Ok(vec![r.to_vec(), s.to_vec()])
+    }
+}
+
+impl<C, T> fmt::Debug for EcdsaSigner<T, C> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("EcdsaSigner").finish()
```
wiktor-k, line 117: Just out of curiosity why not simply derive the Debug? Is this about hiding the `inner` which may contain private bits?

src/helper/ecdsa.rs
```diff
@@ -0,0 +1,223 @@
...
+    fn sign_prehash(&self, hash: HashAlgorithm, prehash: &[u8]) -> Result<Vec<Vec<u8>>> {
+        if C::Digest::HASH_ALGORITHM != hash {
+            bail!(
+                "Signer only support {expected:?}, found {found:?}",
+                expected = C::Digest::HASH_ALGORITHM,
+                found = hash
+            );
+        }
+        // TODO: check digest size
+        let signature = self.inner.sign_prehash(prehash)?;
```
wiktor-k, line 109: There's one edge-case that various crypto libraries treat differently: in PGP you can use bigger digests with smaller curves (say, SHA-512 with P-256) and the digest needs to be truncated before being passed to the underlying crypto primitive in some libs.

Maybe it'd be good to add tests for these scenarios?

src/helper/ecdsa.rs
```diff
@@ -0,0 +1,223 @@
...
+        let sig = self.sign_prehash(hash, prehashed_data)?;
+
+        // MPI format:
+        // strip leading zeros, to match parse results from MPIs
+        let mpis = sig
+            .iter()
+            .map(|v| Mpi::from_slice(&v[..]))
+            .collect::<Vec<_>>();
+
+        Ok(SignatureBytes::Mpis(mpis))
```
wiktor-k, line 157: @hko-s can you comment on the (lack of) native case here? (for ed25519)

src/helper/ecdsa.rs
```diff
@@ -0,0 +1,223 @@
...
+        F: FnOnce() -> String,
+    {
+        let sig = self.sign_prehash(hash, prehashed_data)?;
+
+        // MPI format:
+        // strip leading zeros, to match parse results from MPIs
+        let mpis = sig
+            .iter()
+            .map(|v| Mpi::from_slice(&v[..]))
+            .collect::<Vec<_>>();
```
wiktor-k, line 155: I wonder if you need the explicit type here since `SignatureBytes::Mpis` takes a `Vec`:

```suggestion
            .collect();
```

src/helper/ecdsa.rs
```diff
@@ -0,0 +1,223 @@
...
+
+        Ok(SignatureBytes::Mpis(mpis))
+    }
+
+    fn public_key(&self) -> Self::PublicKey {
```
wiktor-k, line 160: Hmm... this is kind of problematic in the interface that it wants an owned thing... I guess nothing we could do here about it :thinking:

src/helper/rsa.rs
```diff
@@ -0,0 +1,205 @@
...
+impl<T, D> RsaSigner<T, D>
+where
+    D: Digest,
+    T: Keypair<VerifyingKey = VerifyingKey<D>>,
+{
+    /// Create a new signer with a given public key
+    pub fn new(inner: T, created_at: DateTime<Utc>) -> Result<Self> {
+        let public_key = PublicKey::new(
+            Version::New,
+            KeyVersion::V4,
```
wiktor-k, line 59: Too bad that this supports only the V4 version. I'll ping our V6 evangelist here: @hko-s ;)

src/helper/rsa.rs
```diff
@@ -0,0 +1,205 @@
...
+    fn create_signature<F>(
+        &self,
+        _key_pw: F,
+        hash: HashAlgorithm,
+        prehashed_data: &[u8],
+    ) -> Result<SignatureBytes>
+    where
+        F: FnOnce() -> String,
+    {
+        let sig = self.sign_prehash(hash, prehashed_data)?;
```
wiktor-k, line 130: There's quite a lot of repetition with this and the ECDSA case... I wonder if it's possible to somehow move it to a common function or something :thinking:

wiktor-k (commented):
I think this already looks good. It's quite limited to signing only but I think it's a step in the right direction.

One thing I'd be happy to see extended here is tests which use the `Signer` interface but plug in PGP keys.

Thanks for you work. Highly appreciated! :bow: 

(and sorry for the extreme delay :pray: )

src/helper/ecdsa.rs
baloo: inner then needs to `impl Debug` which might not be feasible yeah.
I'll add the `public_key`

src/helper/ecdsa.rs
baloo: I am not sure how that works here though?
The `SecretKey::hash_alg` returns the hash algorithm this signer supports, and only this one is accepted.

baloo: I removed the TODO for now.

src/helper/rsa.rs
baloo: You know a whole lot more about PGP than I do.

src/helper/rsa.rs
baloo: I don't think I can do that, maybe a macro, but even then there are subtle differences and I worry about readability.

dignifiedquire:
@hko-s @wiktor-k would this allow to simplify the code in https://codeberg.org/openpgp-card/rpgp ?

hko-s:
> @hko-s @wiktor-k would this allow to simplify the code in https://codeberg.org/openpgp-card/rpgp ?

I'd also like to know this :sweat_smile:

The amount of code in openpgp-card-rpgp that handles signing is only ~100 lines, so ... probably not by a whole lot.
But that said, even if it would enable just a minor simplification, I'd be all for it.

baloo:
I think I messed up my last rebase and the `Cargo.lock` was from another branch. Could you run the tests again?

baloo:
> > would this allow to simplify the code in https://codeberg.org/openpgp-card/rpgp ?
> 
> I could try porting [signstar](https://gitlab.archlinux.org/archlinux/signstar) to this API and reporting potential issues. Note that we use quite a bit of esoteric features, such as creating OpenPGP signatures from partial file hash and appending notations to the signatures (to capture the original file digest)...
> 
> Yeah... maybe openpgp-card would be a better fit 🤔 or [oct-git](https://codeberg.org/openpgp-card/oct-git) (which does signing only).

Well, I've deployed this to sign RPM with an HSM, and that works for us.

src/helper/ecdsa.rs
hko-s: This code is for ECDSA.

Were you are thinking of EdDSALegacy and Ed25519 (see https://www.rfc-editor.org/rfc/rfc9580.html#name-public-key-algorithms)?

This PR currently doesn't support EdDSA, so that case doesn't apply to the code here.

src/helper/rsa.rs
hko-s: I think the constructor could take a `KeyVersion` parameter and accept either V4 or V6, so Signers could act as both generations of OpenPGP signing key.

I don't think that's urgent though, especially if we don't mind semver breaking changes in this API every now and then.

hko-s (commented):
This is great, especially hearing that it's tested and working with an HSM!

I'd love to see this PR landed after we finish the current round of rPGP changes (which will hopefully soon become a 0.16 release). This PR will probably need to be reshuffled a bit, then, though.

src/helper/ecdsa.rs
baloo: No. We’re targeting signature of rpm packages with gpg. The hsm we have are running in fips-validated mode and this hides Eddsa support from us.
We very explicitly wanted to add support for the two algorithms we have: rsa and ecdsa.

I did not test other algorithms.

hko-s: @baloo right, I assumed that you implemented support based on the hardware backend you wanted to support. That seems good to me.

However, @wiktor-k asked a question, which I think applies to EdDSA (unless I misunderstood him).
So I think all is well and we can close this thread (pending confirmation by Wiktor)

hko-s:
Actually ... maybe we can get this PR into the next release?
I'll try to rebase and port it to main.

baloo:
> Actually ... maybe we can get this PR into the next release? I'll try to rebase and port it to main.

I can do a rebase later tonight

src/helper/ecdsa.rs
baloo: ha, right, I've read the email notification which did not carry the context >< My bad.

wiktor-k: Hmm... Maybe I missed the fact that this is E*C*DSA signer. Let's close this thread for now 😅

hko-s: > ha, right, I've read the email notification which did not carry the context >< My bad.

no problem of course, i did wonder how to make that reply email-safe, but clearly failed :smile: 

thanks, @wiktor-k - closing this thread for now

hko-s:
> > Actually ... maybe we can get this PR into the next release? I'll try to rebase and port it to main.
> 
> I can do a rebase later tonight EDIT: there is a significant amount of API breakage, I will need some time to work through it.

I've done a first round of cleanup in https://github.com/hko-s/rpgp/tree/baloo/signer-helper

One thing that's still missing is that rPGP now uses Rust Crypto types to represent keys internally.

hko-s:
My branch now compiles, I opened #554 to run it in CI

hko-s:
Short FYI update on this (and/or #554): I now think we won't manage to get this into the next release, but I remain very interested in helping to make this land soon. We'll need to figure out how to bring rPGP's trait interface and the traits needed in this PR into alignment, and that will take a bit more work (but that sounds like extremely useful work to me).
@dignifiedquire wanted to work on the private key handling aspect of rPGP anyway (and there's a WIP PR #538 in flight about that, which will also not get into the upcoming release).

src/helper/ecdsa.rs
```diff
@@ -0,0 +1,202 @@
...
+    errors::{bail, Result},
+    helper::PgpPublicKey,
+    packet::{PubKeyInner, PublicKey},
+    types::{
+        EcdsaPublicParams, Fingerprint, KeyDetails, KeyId, KeyVersion, Mpi, Password,
+        PublicKeyTrait, PublicParams, SecretKeyTrait, SignatureBytes,
+    },
+};
+
+impl<C> PgpPublicKey for ecdsa::VerifyingKey<C>
```
dignifiedquire: given the crates name is `pgp`, I don' think we need all the `Pgp` prefixes

src/helper/ecdsa.rs
```diff
@@ -0,0 +1,202 @@
...
+
+impl<C, T> EcdsaSigner<T, C>
+where
+    C: PrimeCurve + DigestPrimitive,
+    SignatureSize<C>: ArrayLength<u8>,
+    T: PrehashSigner<ecdsa::Signature<C>>,
+    C::Digest: KnownDigest,
+{
+    fn sign_prehash(&self, hash: HashAlgorithm, prehash: &[u8]) -> Result<Vec<Vec<u8>>> {
+        if C::Digest::HASH_ALGORITHM != hash {
```
dignifiedquire: maybe use `ensure_eq`?

src/helper/ecdsa.rs
```diff
@@ -0,0 +1,202 @@
...
+            );
+        }
+
+        let signature = self.inner.sign_prehash(prehash)?;
+        let (r, s) = signature.split_bytes();
+        Ok(vec![r.to_vec(), s.to_vec()])
+    }
+}
+
+impl<C, T> fmt::Debug for EcdsaSigner<T, C> {
```
dignifiedquire: we now have `derive_more` in the deps, so please use that for custom debug impls

dignifiedquire:
left some additional comments, otherwise I think this is in a shape that we can get it in

src/helper/rsa.rs
dignifiedquire: I think it would be nice to add the key version param

baloo:
I think I fixed all the comments missing, anything else that would be required?

dignifiedquire:
I think the last thing is the name `helper`, that's a bit unfortunate/not saying much. Maybe `adapter` or sth like that

dignifiedquire:
thanks

PR #434, opened 2024-11-19
fix: extend is_signing_key() and is_encryption_key() for RFC 9580 algorithms

link2xt:
X25519 and X448 are encryption keys,
while Ed25519 and Ed448 are signing keys.


hko-s (approved):
Thanks for debugging and fixing this, and sorry I missed doing that when implementing RFC 9580!

PR #436, opened 2024-11-23
feat: implement TryFrom<PublicOrSecret> for Signed{Public,Secret}Key

link2xt:
This is less dangerous than checking
`is_public()` or `is_secret()`
and then using `into_secret()` and `into_public()` that may panic.

### DIFF
--- a/src/composed/signed_key/shared.rs
+++ b/src/composed/signed_key/shared.rs
@@ -1,3 +1,4 @@
+use std::fmt;
 use std::io;
 
 use chrono::Duration;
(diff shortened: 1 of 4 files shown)


src/composed/signed_key/shared.rs
```diff
@@ -246,6 +247,41 @@ impl PublicOrSecret {
...
+impl fmt::Display for TryFromPublicOrSecretError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Attempt to convert PublicOrSecret key to the wrong type")
+    }
+}
+
+impl std::error::Error for TryFromPublicOrSecretError {}
+
+impl TryFrom<PublicOrSecret> for SignedPublicKey {
+    type Error = TryFromPublicOrSecretError;
```
hko-s, line 264: Wouldn't it be easier and more ergonomic to use `pgp::errors::Error` here?

hko-s: Hmm. I never thought about how big `pgp::errors::Error` is, in terms of memory. I would have hoped it's minor.

Is there reason to consider this operation performance critical? The parsing that precedes this call should be more expensive by orders of magnitude.

I'd worry that a separate error type makes error handling more confusing for users of rPGP.

So my preference would be to reuse the regular rpgp error type, but this preference is not very strongly held.

hko-s:
Putting aside the error-type question in my comment, I think this PR is good.

It would be nice to deprecate/remove the panic-y functions at some point.
Maybe deprecating them already now would be good?

hko-s:
> It would be nice to deprecate/remove the panic-y functions at some point. Maybe deprecating them already now would be good?

Oh no. The deprecation is making CI very angry :scream_cat:

src/composed/signed_key/shared.rs
dignifiedquire: we probably need to split that error up into multiple suberrors anyway at some point, so this fine

src/composed/signed_key/shared.rs
```diff
@@ -246,6 +249,41 @@ impl PublicOrSecret {
     }
 }
 
+/// Error returned when trying to convert `PublicOrSecret` key
+/// into the wrong type.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct TryFromPublicOrSecretError;
```
dignifiedquire, line 255: we already have `thiserror` in the dependency tree, so please use that

dignifiedquire (approved):

dignifiedquire (approved):

PR #444, opened 2024-12-05
refactor: remove some dangerous slicing

link2xt:


src/packet/many.rs
```diff
@@ -83,7 +83,7 @@ impl<R: Read> Iterator for PacketParser<R> {
                             break;
                         }
                         Ok(r) => {
-                            body.extend_from_slice(&buf[..r]);
+                            body.extend_from_slice(buf.get(..r).unwrap_or(buf.as_slice()));
```
iequidoo, line 86: Shouldn't it be an error if `buf` isn't long enough?

src/packet/many.rs
```diff
@@ -258,7 +258,9 @@ fn read_fixed<R: Read>(
     }
 
     out.resize(out_len + len, 0u8);
-    reader.read_exact(&mut out[out_len..])?;
+    if let Some(ref mut out) = out.get_mut(out_len..) {
```
iequidoo, line 261: This isn't `if` actually, maybe `let ... else return Err(...);` is better? Some dead code is added anyway, be it `if` or `return Err(...)`, but the latter protects from wrong code modifications at least

hko-s:
As a high-level comment/context, the code that this PR changes has been made more robus very recently, in:

https://github.com/rpgp/rpgp/commit/a69992ca38625f671770a6c1b0cc5b852d3aedc0#diff-0c2121684730f47b19eac98dd7042f60a1376f48e9d5a285e5719fd500d9c562

So the crashes that Delta observed should hopefully not be possible anymore with the current code.

(But making the code easier to read, or more solid yet, is of course still good. I haven't tired to fully understand the diff yet.)

src/packet/many.rs
hko-s: Like @iequidoo, I also find it odd to quietly ignore an unexpectedly short `buf`.

The Reader should never report to have read more data (`r`) than fits into `buf`, I think? So throwing an Error in that case seems best to me.

hko-s:
Besides the minor nitpicks in the comments above, I think this change is good!

hko-s:
> I added errors in two places. But generally I think it does not matter much what happens in the case which must never happen as long as it does not potentially panic.

I kind of agree. But if such invariants ever fail, debugging a quiet failure sounds nightmarish, so I appreciate explicit erroring!

hko-s (approved):
Looks good to me!

hko-s:
> Returning errors works if the function already returns `Result`, but if the function is supposed to never fail, the only options are (roughly) `unwrap()` and `unwrap_or_default()`, in which case I prefer `unwrap_or_default()`.

:+1: 

In some such cases, I think it's also an option to change the signature to return `Result`.
Although probably not in every case.

iequidoo (approved):

hko-s:
> > In some such cases, I think it's also an option to change the signature to return `Result`. Although probably not in every case.
> 
> We already have this in https://docs.rs/pgp/0.14.2/pgp/composed/message/enum.Message.html#method.to_armored_bytes and https://docs.rs/pgp/0.14.2/pgp/composed/message/enum.Message.html#method.to_armored_string These functions never fail but return Result which is a bad API IMO.

Seems to me that these fail if the underlying writer fails?

hko-s:
> > Seems to me that these fail if the underlying writer fails?
> 
> `write` implementation for `Vec` never fails, it has no code path that returns errors.

Sure, but some callers will write to different targets?

hko-s:
> There is no way `to_armored_bytes` can write to anything other than Vec.

Oh! Now I see what you're saying, which took decidedly too long.

iequidoo:
> We already have this in https://docs.rs/pgp/0.14.2/pgp/composed/message/enum.Message.html#method.to_armored_bytes and https://docs.rs/pgp/0.14.2/pgp/composed/message/enum.Message.html#method.to_armored_string
> These functions never fail but return Result which is a bad API IMO.

But the only alternative is `panic!`. I think returning Result is an acceptable API, e.g. in C++ `std::bad_alloc` and other exceptions may be thrown almost everywhere, sometimes even instead of assertions to not cause the whole program to crash, but they live with this somehow. Rust's `Result` is better anyway

dignifiedquire:
going to close this for now, will revisit this later

Issue #158, opened 2022-01-14
Cannot decrypt messages larger than 1 GiB


OliverEvans96:
Hello, thanks for the great package.

There is a seemingly [arbitrary restriction](https://github.com/rpgp/rpgp/blob/master/src/packet/many.rs#L10) preventing messages > 1 GiB from being parsed.

Do you have a suggestion for decrypting large files? Could this variable simply be removed, along with the corresponding [check](https://github.com/rpgp/rpgp/blob/master/src/packet/many.rs#L120)?

Thanks,
Oliver

dignifiedquire:
it could probably be made configurable, though there should be a limit to protect against too large files

ChuanDou2021:
Any suggestion to process large file in stream style ? Such as demo or tests, thanks

kellpossible:
Also interested in this

dignifiedquire:
this fixed on main

PR #446, opened 2024-12-13
Remove MAX_CAPACITY limit

ntrinquier:
Fixes https://github.com/rpgp/rpgp/issues/158.

I believe `MAX_CAPACITY` was useful in previous implementations, e.g. when we needed to grow a buffer:

https://github.com/rpgp/rpgp/blob/bbd38f679a1cd3a6cddb4dbe9d4e9cd2eb4ad742/src/packet/mod.rs#L180-L186

However in the `Iterator` implementation, there _shouldn't_ be any reason to enforce a limit anymore?

It might be slow to parse > 1GiB, however I'd argue it is on users to decide if they are willing to accept the slowness (as opposed to have `rpgg` enforce an arbitrary limit).

### DIFF
--- a/src/packet/many.rs
+++ b/src/packet/many.rs
@@ -8,8 +8,6 @@ use crate::packet::packet_sum::Packet;
 use crate::packet::single;
 use crate::types::{PacketLength, Tag};
 
-const MAX_CAPACITY: usize = 1024 * 1024 * 1024;
-
 const DEFAULT_CAPACITY: usize = 1024 * 16;
 const READER_POLICY: MinBuffered = MinBuffered(1024);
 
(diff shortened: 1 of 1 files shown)


hko-s:
The purpose of these "max size" checks is to protect applications from (potentially maliciously caused by a sender) allocations that could cause their application to be OOM-killed.

That said, size limits are of course generally undesirable to have.

Ideally, I'd want to see streaming support in most parts of rPGP, to eliminate the possibility of excessive allocations within rPGP. In cases where this is not possible, caller-configurable size limitations might be the next best thing.

Do you have a specific need/use case, where the current size limit is a problem for you? If so, can you share some background?

dignifiedquire:
`mmap` is one way of doing this, but still brings risk in regards to memory exhaustion. The only real fix is to implement proper streaming support, for which these limits are not needed anymore, everything else can still pose a security risk when using the library on untrusted input

dignifiedquire:
I am working on a solution to stream messages here https://github.com/rpgp/rpgp/pull/453

Issue #367, opened 2024-07-06
Maybe rename PublicKey/PublicSubkey::sign() to be more expressive


hko-s:
`PublicKey::sign()` and `PublicSubkey::sign()` produce a "Primary Key Binding Signature (type ID 0x19)" and "Subkey Binding Signature (type ID 0x18)", respectively.

I think it would be nice to give these functions more expressive names.
One possibility would be to rename both to `bind`.

Alternatively, they could have longer form names to roughly reflect the RFC terms, such as `bind_primary_key` and `bind_subkey`.

dignifiedquire:
I like the longer forms, `bind` is very opaque

PR #450, opened 2024-12-20
rename specialized sign functions

hko-s:
Explore handling of #367: give specialized `sign` functions more specific names.


dignifiedquire:
closing as we won't do this likely

PR #453, opened 2024-12-30
feat: implement streaming message reading and writing

dignifiedquire:
- reduce allocations when doing symmetric encryption & decryption
- fullly streaming message reading
- fully streaming message writing
- bzip2 support


src/composed/message/builder.rs
```diff
@@ -0,0 +1,153 @@
...
+    compression: CompressionAlgorithm,
+    encryption: Encryption,
+}
+
+enum Source {
+    Bytes(Bytes),
+    File(PathBuf),
+}
+
+enum Encryption {
```
hko-s, line 56: Splitting encryption between password-based and public key-based seems unfortunate to me: One possible way to use OpenPGP encryption is to mix encryption to a set of public keys and additionally to a set of passwords.

I think something like this would be a good shape:

```
#[allow(dead_code)]
enum Encryption {
    SeipdV1 {
        session_key: Zeroizing<Vec<u8>>,
        sym_alg: SymmetricKeyAlgorithm,
        pkesk: Vec<PublicKeyEncryptedSessionKey>,
        skesk: Vec<SymKeyEncryptedSessionKey>,
    },
}
```

It might be nice to have an enum `EncryptedSessionKey` that combines the two? (Then we'd just need one `esk: Vec<EncryptedSessionKey>`, here)

hko-s: On second thought, maybe we should enforce [alignment between ESK-version and Seipd-version](https://www.rfc-editor.org/rfc/rfc9580.html#name-openpgp-encrypted-message-p) on the type level, here:

- SeipdV2 may only be combined with v6 ESK
- SeipdV1 may only be combined with v3/4 ESK

dignifiedquire: >  One possible way to use OpenPGP encryption is to mix encryption to a set of public keys and additionally to a set of passwords

using the same session key? I have never seen this in the wild, and am not sure that would be a good idea security wise 🤷

hko-s: I'm sure this is a relatively exotic use case, but it's clearly what the RFC has in mind. See the [Message grammar](https://www.rfc-editor.org/rfc/rfc9580.html#name-openpgp-messages).

I believe the idea is to have a single session key, and encrypt it to a set of recipients. This can take the form of a mix of `Public Key Encrypted Session Key Packet | Symmetric Key Encrypted Session Key Packet`.

This can have weird security properties, I guess? But then, so does encrypting a single session key to a mix of PKESK with very different algorithm strength.

dignifiedquire: it is now implemented

src/composed/message/builder.rs
```diff
@@ -0,0 +1,1705 @@
...
+                    &mut rng,
+                    Seipd::V1 {
+                        sym_alg: SymmetricKeyAlgorithm::AES128,
+                    },
+                    &pkey,
+                )
+                .expect("encryption")
+                .encrypt_with_password(
+                    &mut rng,
+                    Seipd::V1 {
```
hko-s, line 1220: It seems wrong to me that `encrypt_to_key` and `encrypt_with_password` take separate `Seipd` parameters. An encrypted message can only contain one symmetric encryption container, so the symmetric container should only be configurable once, in this type of call.

Maybe the symmetric container should be configured in the top `Builder::from_reader` call?

dignifiedquire: > Maybe the symmetric container should be configured in the top Builder::from_reader call?

that would not be a nice api, as you don't always encrypt

hko-s: Alternatively, we could combine the two `encrypt_to_*` into one call?

dignifiedquire: using type states now, just pushed

hko-s: uhh, that looks very nice!

wiktor-k (commented):
I've read it but most of my remarks are rather questions than issues :sweat_smile:

src/composed/message/decrypt.rs
```diff
@@ -5,17 +5,17 @@ use crate::crypto::sym::SymmetricKeyAlgorithm;
 use crate::errors::Result;
 use crate::packet::SymKeyEncryptedSessionKey;
 use crate::types::{EskType, PkeskBytes, PlainSecretParams, SecretKeyTrait, SkeskVersion};
+use crate::SignedSecretKey;
 
 /// Decrypts session key using secret key.
 pub fn decrypt_session_key<F, L>(
```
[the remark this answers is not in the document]
hko-s: Err, it occurs to me that I'm confused how this interacts with "external" private key material. I'll report back once I've (mostly) ported `openpgp-card-rpgp` to this PR.

hko-s:
I think we lost support for brainpool, e.g. in `EcdsaPublicParams`, somewhere in this PR. If true, that means we apparently lack tests for such keys.

hko-s:
While reading `message/builder.rs`, I feel like the ordering inside the file could use some re-ordering.
Trying to look at the file  from top to bottom is a bit confusing.

E.g.:
- `Builder::from_reader` is rather far away from `from_file`/`from_bytes`
-  `fn prepare` is defined far away from its use (and it could use a more expressive name, maybe `prepare_signature_configs`).

src/composed/message/builder.rs
```diff
@@ -0,0 +1,1962 @@
...
+
+            std::io::copy(&mut encrypted, &mut out)?;
+        }
+    }
+
+    Ok(())
+}
+
+struct SignGenerator<'a, R: std::io::Read> {
+    total_len: Option<u32>,
```
hko-s, line 922: Should this be an `Option<u64>`?

src/packet/literal_data.rs
```diff
@@ -119,57 +174,473 @@ impl AsRef<[u8]> for LiteralData {
...
+pub(crate) enum LiteralDataGenerator<R: io::Read> {
+    Fixed(LiteralDataFixedGenerator<MaybeNormalizedReader<R>>),
+    Partial(LiteralDataPartialGenerator<MaybeNormalizedReader<R>>),
+}
+
+impl<R: io::Read> LiteralDataGenerator<R> {
+    pub(crate) fn new(
+        header: LiteralDataHeader,
+        source: R,
+        source_len: Option<u32>,
```
hko-s, line 251: Should be an `Option<u64>`, I think?

src/packet/compressed_data.rs
```diff
@@ -86,14 +111,408 @@ impl Serialize for CompressedData {
...
+pub(crate) enum CompressedDataGenerator<R: io::Read> {
+    Fixed(CompressedDataFixedGenerator<R>),
+    Partial(CompressedDataPartialGenerator<R>),
+}
+
+impl<R: io::Read> CompressedDataGenerator<R> {
+    pub(crate) fn new(
+        alg: CompressionAlgorithm,
+        source: R,
+        source_len: Option<u32>,
```
hko-s, line 194: Should this be an `Option<u64>`?

src/composed/message/builder.rs
```diff
@@ -0,0 +1,1962 @@
...
+                    self.chunk_size,
+                    self.compression,
+                    self.encryption,
+                    out,
+                )?;
+            }
+            Source::File(ref path) => {
+                let in_file = std::fs::File::open(path)?;
+                let in_meta = in_file.metadata()?;
+                let in_file_size: u32 = in_meta.len().try_into()?;
```
hko-s, line 487: I think either the `source_len` parameters need to be `Option<u64>`, or this preparation for calling `to_writer_inner` needs to map "too large" to `None`. Currently I think large inputs will error out?

(It seems preferable to pass in an `Option<u64>` length hint instead of encoding "large file length" as `None`)

src/types/mpi.rs
```diff
@@ -1,209 +1,164 @@
...
+pub struct MpiBytes(#[debug("{}", hex::encode(_0))] Bytes);
 
-    /// Strips leading zeros.
-    pub fn from_slice(raw: &[u8]) -> Self {
-        Mpi(strip_leading_zeros(raw).to_vec())
+impl MpiBytes {
+    /// Wraps the given bytes as an MPI, must be normalized before
+    /// Avoid if possible.
+    // TODO: remove
+    pub fn from_raw(bytes: Bytes) -> Self {
```
hko-s, line 26: I fear that we might still need this internally, for handling half-broken legacy public keys.

But putting that to the side:

I think we should make this fn `pub(crate)`. It should really not be needed from outside of rpgp.

src/packet/signature/config.rs
```diff
@@ -374,7 +374,7 @@ impl SignatureConfig {
     }
 
     /// Calculate the serialized version of this packet, but only the part relevant for hashing.
-    pub fn hash_signature_data(&self, hasher: &mut dyn std::io::Write) -> Result<usize> {
+    pub fn hash_signature_data(&self, hasher: &mut Box<dyn DynDigest>) -> Result<usize> {
```
wiktor-k, line 377: This effectively reverts https://github.com/rpgp/rpgp/pull/393 :(

We're using that in Signstar for creating OpenPGP signatures from already open hashers and due to us using different version of sha2 crate we can't really plug it in here (since it implements a different `DynDigest` from Rust's perspective).

Would it be possible to revert this small change? If so I'd add a test case so that it doesn't break in the future... (but later, after the dust settles from all these PRs).

README.md
```diff
@@ -19,7 +19,7 @@
   <!-- msrv -->
   <a href="https://img.shields.io/badge/rustc-1.75+-blue.svg?style=flat-square">
     <img src="https://img.shields.io/badge/rustc-1.75+-blue.svg?style=flat-square"
```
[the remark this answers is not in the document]
hko-s: Thanks, put this in #502

src/crypto/hash.rs
```diff
@@ -10,183 +10,138 @@ use crate::errors::{Error, Result};
...
-        match self {
-            Self::SHA2_224 => Some(16),
-            Self::SHA2_256 => Some(16),
-            Self::SHA2_384 => Some(24),
-            Self::SHA2_512 => Some(32),
-            Self::SHA3_256 => Some(16),
-            Self::SHA3_512 => Some(32),
-            _ => None,
-        }
+        Self::Sha256
```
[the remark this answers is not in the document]
hko-s: Deriving default for `HashAlgorithm` isn't compatible with using `#[num_enum(catch_all)]`, apparently. 

```
error: Attribute `catch_all` is mutually exclusive with `default`
  --> src/crypto/hash.rs:55:16
   |
55 |     #[num_enum(catch_all)]
   |                ^^^^^^^^^
```

Slightly annoying, but not a huge deal, I guess :shrug:

src/crypto/x25519.rs
```diff
@@ -1,25 +1,56 @@
...
-    type KeyParams = ();
+impl Eq for SecretKey {}
+
+impl SecretKey {
+    /// Generate an X25519 `SecretKey`.
+    pub fn generate<R: Rng + CryptoRng>(mut rng: R) -> Self {
+        let mut secret_key_bytes = Zeroizing::new([0u8; 32]);
+        rng.fill_bytes(&mut *secret_key_bytes);
+
+        let secret = StaticSecret::from(*secret_key_bytes);
```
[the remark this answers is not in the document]
hko-s: That allows callers to choose what kind of random they want to use.

And it enables rPGP using fully deterministic tests, based on Rngs that are initialized to a constant value, like e.g. here: https://github.com/rpgp/rpgp/blob/main/src/packet/sym_encrypted_protected_data.rs#L392

src/crypto/sym.rs
```diff
@@ -414,6 +415,161 @@ impl SymmetricKeyAlgorithm {
         Ok(ciphertext)
     }
 
+    pub fn encrypted_protected_len(&self, plaintext_len: usize) -> usize {
+        // MDC is 1 byte packet tag, 1 byte length prefix and 20 bytes SHA1 hash.
+        let mdc_len = 22;
+        self.block_size() + 2 + plaintext_len + mdc_len
```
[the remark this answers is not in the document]
hko-s: suggested a patch in #502

src/types/params/public.rs
```diff
@@ -51,130 +51,245 @@ pub enum PublicParams {
...
     }
 }
 
 #[derive(Debug, PartialEq, Eq, Clone)]
 #[cfg_attr(test, derive(proptest_derive::Arbitrary))]
 pub enum EcdhPublicParams {
     /// ECDH public parameters for a curve that we know uses Mpi representation
-    Known {
-        curve: ECCCurve,
-        p: Mpi,
```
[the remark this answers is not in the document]
hko-s: I opened #503 to track that I might one day want to dig into finding examples for this concern. For now, I'll assume it's super exotic and we shouldn't care.

src/packet/header.rs
```diff
@@ -114,10 +291,22 @@ pub struct NewPacketHeader {
     tag: u8,
 }
 
+fn old_fixed_type(len: usize) -> u8 {
```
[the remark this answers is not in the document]
hko-s: Suggested a doc comment in #502

src/packet/secret_key_parser.rs
```diff
@@ -1,126 +1,99 @@
...
     pub_len: Option<usize>,
-) -> impl Fn(&[u8]) -> IResult<&[u8], (PublicParams, SecretParams)> {
-    move |i| {
-        map_res(
-            tuple((PublicParams::try_from_slice(typ, pub_len), rest)),
-            |(pub_params, v)| {
-                if let Some(pub_len) = pub_len {
-                    // if we received a public key material length (from a v6 secret key packet),
-                    // make sure that we consumed the expected number of bytes
-                    if i.len() - v.len() != pub_len {
```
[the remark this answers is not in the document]
hko-s: I added a commit to re-add this logic in #502, but it's a little awkward.

src/crypto/eddsa.rs
```diff
@@ -64,6 +62,17 @@ impl From<&SecretKey> for EddsaLegacyPublicParams {
 }
 
 impl SecretKey {
+    /// Generate an EdDSA `SecretKey`.
+    ///
+    /// `mode` picks between supported EdDSA key formats and curves
```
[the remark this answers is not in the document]
hko-s: added a commit for this to #502

src/types/mpi.rs
hko-s: done in #502

src/packet/signature/config.rs
dignifiedquire: Changing back to `std::io::Write` for the hashing interface is not gonna be possible in the current state, as `DynDigest` simply does not implement it. So I would suggest for the moment to write a small wrapper proxying things for using another type.

hko-s (approved):
Yay for streaming support! After quite a bit of reading and testing, I think we're good to merge!

hko-s:
> @dignifiedquire @hko-s thanks for this! Do you have an ETA for a release? I saw there are two alpha prereleases (which work well), but ideally I'd like to avoid pinning a version to keep receiving updates.

@ntrinquier: In case you haven't seen, there is now a stable release

PR #455, opened 2025-01-04
refactor: remove unused feature flags and improve CI runtime

dignifiedquire:
- `nightly` wasn't having any real benefit anymore
- `gperftools` is not really used anymore to profile
- ignore some slow tests to speed up CI
- improve over all CI times

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,6 @@ jobs:
         os: [ubuntu-latest, macOS-latest]
         rust:
           - nightly-2024-12-04
-          - beta
           - stable
 
     steps:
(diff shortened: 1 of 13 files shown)


hko-s (approved):
I don't fully grasp what exactly this PR changes about what github runs (I can't easily spot where the duplications happened), but to my superficial understanding this all looks reasonable.

In particular, it's of course very sensible to move more number-crunchy tests to run with `--release`.

Additionally, I've just built the `rsop` stack on this branch, and that also looks good.

wiktor-k:
Sorry to comment so late but one thing in mind with `nextest` is that it [does *not* run doctests](https://github.com/nextest-rs/nextest/issues/16). [In signstar](https://gitlab.archlinux.org/archlinux/signstar/-/blob/main/justfile?ref_type=heads#L267) we run `cargo nextest run --all` and then `cargo test --doc` to cover everything.

hko-s:
Good point, I've made https://github.com/rpgp/rpgp/pull/487 to try and address this

PR #457, opened 2025-01-04
chore: add Cargo.lock

### LINKED ISSUE #407: Add `Cargo.lock` to the repository and un-gitignore it
Hi,

Given the guidance in https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html I think it'd be good to add `Cargo.lock` to repository and un-gitignore it.

I had a case of using old repository and a really stale Cargo.lock which resulted in compilation issues (w.r.t `Duration::seconds`), removing `Cargo.lock` file helped.

If the file was committed then every contributor would have the same exact set of dependencies.

(Of course this doesn't affect users of rpgp, only local developers).

dignifiedquire:
This should stabilize development, and has no effect on library usage.

A CI check was added, for ensuring minimal version selection works, without `Cargo.lock`, and to ensure baseline compatability works for everyone.

Closes #407


hko-s (approved):
Looks good to me. Great to have a minimal versions check in CI!

PR #461, opened 2025-02-06
add test for reading the carol sample key; fix handling ElgamalEncrypt secret key packet

hko-s:
The [OpenPGP interoperability test suite](https://sequoia-pgp.gitlab.io/openpgp-interoperability-test-suite/results.html) uses a private key "Carol", which it describes as coming from ["draft-bre-openpgp-samples-00"](https://datatracker.ietf.org/doc/draft-bre-openpgp-samples/).

However, that draft doesn't seem to contain that key. So it's currently unclear to me where that key is actually from.

Independent of the provenance of that sample key, rPGP currently fails to load it with `Message(\"unexpected packet data: InvalidPacketContent(PacketError(MapRes))\")")`.

That seems bad. For now, this PR just has a unit test for the problem. But it should evolve to also contain a fix :sweat_smile:


hko-s:
Update: the Carol key comes from https://gitlab.com/openpgp-wg/openpgp-samples/-/merge_requests/3/

hko-s:
Update: fixed by implementing secret key packet loading for both elgamal packet variants.

PR #463, opened 2025-02-07
WIP: subpacket length must be hashed in its original encoding for signatures

hko-s:
The signatures in the `evn.cert` test case use subpacket length encodings in the ["5 byte format"](https://www.rfc-editor.org/rfc/rfc9580.html#name-signature-subpacket-specifi), even though the lengths are small.

This is legal, and when hashing such subpackets, the original encoding must be hashed.

This PR implements a (somewhat horrifying) fix for handling signatures that contain such subpackets.


hko-s:
Fixed in #453

PR #466, opened 2025-02-11
xx: make seipd Config public

hko-s:
This is just a note to say that I need to access the SEIPD Config from rpgpie.

I expect you want to re-export it from `src/packet/mod.rs`. In that case, I think it needs a more expressive name. Maybe `SeipdConfig`?


hko-s:
This is now resolved in #453, closing

Issue #413, opened 2024-09-20
KeyFlags should not be encoded as a single u8


hko-s:
[Key Flags](https://www.rfc-editor.org/rfc/rfc9580.html#key-flags) have historically acted as a `u8`, in practice. However, they have already been defined as an arbitrary-length octet string in RFC 4880.

We should re-visit the [encoding](https://github.com/rpgp/rpgp/blob/0439dd5d23e5340115388cb023ded8ec1b6562cc/src/packet/signature/types.rs#L975), as well as more generally the [API](https://github.com/rpgp/rpgp/blob/0439dd5d23e5340115388cb023ded8ec1b6562cc/src/packet/signature/types.rs#L973-L1001) for Key Flags.

hko-s:
We (read: I) should add a test that parses and interprets a certificate with key flags that take more than one `u8` to encode.

And also a test that produces a subkey binding with a "bigger than one `u8`" key flag.

hko-s:
This is resolved in #453

PR #467, opened 2025-02-17
WIP: test artifacts and unit test for "long key flags"

hko-s:
Adding a test for #413


hko-s:
this is fixed in #453

PR #468, opened 2025-02-18
doc: Include crate README in docs for automatic example validation

wiktor-k:
This change makes the README be part of the root crate docs.

The main benefit is that all Rust code examples will be included by `cargo test` (I've marked them `no_run` since they use non-existent files, but I think they could be adjusted to read from sample keys).

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ Additionally, it fully supports all functionality required by the [Autocrypt 1.1
 
 ### Load a public key and verify an inline-signed message
 
-```rust
+```rust no_run
 use std::fs;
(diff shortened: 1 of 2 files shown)


dignifiedquire (approved):
thanks

hko-s:
I like this!

It's a bit sad that the md-links from the README seem to produce dead links in the rustdoc view. I'm unsure if there's a reasonable way to deal with this. Maybe we'd want to break the markdown files up in some smarter way, in a next step?

wiktor-k:
> I'm unsure if there's a reasonable way to deal with this.

I wonder if using direct links to the GitHub repo for these files would be a reasonable trade-off here? :thinking:

PR #469, opened 2025-02-18
build: Make `cargo` not pull the tests submodule when referencing rpgp via git

wiktor-k:
Currently, using a pre-release git dependency in Cargo.toml makes cargo pull the tests directory as well which results in gigantic, unnecessary traffic.

```toml
[dependencies]
pgp = { git = "https://github.com/rpgp/rpgp" }
```

This patch disables pulling the tests by default and makes cargo skip the submodule:

```sh
$ cargo build
    Updating git repository `https://github.com/rpgp/rpgp`
    Skipping git submodule `https://github.com/rpgp/tests.git` due to update strategy in .gitmodules
```

See: https://github.com/rust-lang/cargo/issues/4247#issuecomment-1149178736

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,9 @@ jobs:
       with:
         submodules: recursive
 
+    - name: Checkout submodules
+      run: git submodule update --checkout
+
     - name: Install ${{ matrix.rust }}
       uses: actions-rs/toolchain@v1
       with:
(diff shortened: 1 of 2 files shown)


wiktor-k:
I actually think this should be the default of `actions/checkout` but we'll see what they say: https://github.com/actions/checkout/pull/2086

hko-s:
Thanks, this would be nice to have!

(The default behavior of pulling the tests submodule is also currently bugging me, while testing other projects against an rpgp PR.)

PR #471, opened 2025-02-21
mpi minor cleanup

hko-s:
I found the naming a bit confusing, and the casting seems needlessly complicated too (?).

Also, i think the comparison with MAX_EXTERN_MPI_BITS was wrong (it compared bytes against bits), so the limit was ineffective.

### DIFF
--- a/src/types/mpi.rs
+++ b/src/types/mpi.rs
@@ -35,7 +35,9 @@ impl MpiBytes {
         self.0.is_empty()
     }
 
-    /// Parses the given buffer for an MPI.
+    /// Parses the given buffer as an MPI.
+    ///
+    /// The buffer is expected to be length-prefixed.
     pub fn from_buf<B: bytes::Buf>(mut i: B) -> Result<Self> {
         let len_bits = i.read_be_u16()?;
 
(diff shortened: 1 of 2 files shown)


src/types/mpi.rs
```diff
@@ -35,24 +35,28 @@ impl MpiBytes {
         self.0.is_empty()
     }
 
-    /// Parses the given buffer for an MPI.
+    /// Parses the given buffer as an MPI.
+    ///
+    /// The buffer is expected to be length-prefixed.
     pub fn from_buf<B: bytes::Buf>(mut i: B) -> Result<Self> {
```
hko-s, line 41: I wonder if it would be good to rename the "from" functions to clearly reflect which are "parsing" (including a length header), and which just represent data as an MPI (where the input data is not length-prefixed).

hko-s: I'd suggest renaming `from_buf` to `parse_from_buf`, or something similar (and leaving the other names as they are)

dignifiedquire: > I wonder if it would be good to rename the "from" functions to clearly reflect which are "parsing" (including a length header), and which just represent data as an MPI (where the input data is not length-prefixed).

I am planning to rename them all to `try_from_buf` or sth similar, at one point...

hko-s: My concern is that there are two types of "make me an Mpi" functions: some expect a length-prefixed input, and  some that don't. It's currently easy to get them confused.

PR #474, opened 2025-02-25
Add note on compatibility reasons to use seipdv1

DoumanAsh:
Sorry for small PR, but I think it would be useful to add explanation as to what is SEIPDv1 and why you might want to use it as after upgrading from 0.13 it took me some time to find out

### DIFF
--- a/src/composed/message/types.rs
+++ b/src/composed/message/types.rs
@@ -334,6 +334,20 @@ impl Message {
     }
 
     /// Encrypt the message in SEIPDv1 format to a list of public keys `pkeys`.
+    ///
+    /// ## Note
+    ///
+    /// SEIPDv1 is broadly supported, while SEIPDv2 is specified in the new OpenPGP
+    /// RFC 9580, and not supported by all implementations yet.
+    /// Therefore, SEIPDv1 encryption may sometimes be preferable for compatibility reasons.
+    ///
+    /// Recipients can signal support for SEIPD versions via the
+    /// [Features](https://www.rfc-editor.org/rfc/rfc9580.html#name-features) subpacket in their certificates.
+    ///
+    /// Senders should select the most robust encrypted container format that the intended recipients can handle.
+    /// See [RFC 9580](https://www.rfc-editor.org/rfc/rfc9580.html#section-13.7-8).
+    ///
+    /// Also see ["OpenPGP for application developers"](https://openpgp.dev/book/migration.html#seipd-v2).
     pub fn encrypt_to_keys_seipdv1<R: CryptoRng + Rng>(
         &self,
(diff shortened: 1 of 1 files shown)


hko-s:
Thanks for sending that PR! I totally agree with your thought of helping users of the rPGP API to understand the compatibility situation.

I think we need to tweak the exact text, though: SEIPDv2 is not bound to "v6" keys. Formally, it's bound to "support for RFC 9580", and we should definitely write that. It's unclear to me if/what additional would be good, in that doc-comment.

Maybe something like

```
/// Note that SEIPDv1 is broadly supported, while SEIPDv2 is specified in the new OpenPGP RFC 9580, and not supported by all implementations yet. Therefore, SEIPDv1 encryption may sometimes be preferable for compatibility reasons.
```

DoumanAsh:
Sounds good
I will also add note to opengpg documentation on migration

src/composed/message/types.rs
```diff
@@ -334,6 +334,14 @@ impl Message {
...
 
     /// Encrypt the message in SEIPDv1 format to a list of public keys `pkeys`.
+    ///
+    /// ## Note
+    ///
+    /// SEIPDv1 is broadly supported, while SEIPDv2 is specified in the new OpenPGP
+    /// RFC 9580, and not supported by all implementations yet.
+    /// Therefore, SEIPDv1 encryption may sometimes be preferable for compatibility reasons.
+    ///
+    /// For more details refer to openpgp [documentation](https://openpgp.dev/book/migration.html#seipd-v2)
```
hko-s, line 344:
```suggestion
        /// Recipients can signal support for SEIPD versions via the
        /// [Features](https://www.rfc-editor.org/rfc/rfc9580.html#name-features) subpacket in their certificates.
        ///
        /// Senders should select the most robust encrypted container format that the intended recipients can handle.
        /// See [RFC 9580](https://www.rfc-editor.org/rfc/rfc9580.html#section-13.7-8).
        ///
        /// Also see ["OpenPGP for application developers"](https://openpgp.dev/book/migration.html#seipd-v2).
```

hko-s: After thinking some more, I came up with some more text suggestions :relaxed:

hko-s:
Ah, sorry. I messed up the indentation in my github edit suggestion. Can you `cargo fmt` (or manually fix the comment indentation)?

DoumanAsh:
Sure give me a sec

hko-s (approved):
Thanks again for bringing this clarification idea, and forthe back and forth! I find this  text a massive improvement.

I think we should merge it (and probably add similar text to the upcoming `Builder::seipd_v1/2` functions).

PR #478, opened 2025-03-01
CSF message that starts with a newline

hko-s:
CSF messages that start with a newline aren't currently parsed correctly (as reported in https://codeberg.org/heiko/rsop/issues/43)

It seems the reason for this is that `header_parser` is trying to be flexible about stray newlines.
Which is probably a good thing, for all headers outside of CSF messages.

Maybe `header_parser` should get a `csf: bool` parameter to enable/disable lax processing of newlines?


hko-s:
The windows tests fail kind of unrelated to this test. However, I think it's a mistake that we're changing the line-ending encoding in Literal packets (and I was going to suggest changing it, separately).

https://www.rfc-editor.org/rfc/rfc9580.html#name-text-signature-type-id-0x01 says:

"The signature is calculated over the text data with its line endings converted to <CR><LF>". However it does *NOT* say that we should normalize the line endings in the literal.

There are silly edge cases where such normalization would break otherwise harmless-ish operations: making a "text mode signature" over binary data is weird, but benign. We'd just sign a digest over a weird variation of the actual data, but ... why not.

However, if we normalize `<LF>` bytes to `<CR><LF>` in the literal data - just because we're making a type 0x01 signature - this breaks the payload, if it's actually some binary data.

hko-s:
After digging some more, I am now entirely confused why the `load_csf_starts_with_newline` test case seems to end up with platform-specific line ending normalization.

hko-s:
> After digging some more, I am now entirely confused why the `load_csf_starts_with_newline` test case seems to end up with platform-specific line ending normalization.

I changed the test to compare against normalized line endings. Which seems to be in line with what other tests do.

tests/any_test.rs
```diff
@@ -15,3 +15,24 @@ fn load_empty_csf() {
...
     assert!(matches![any, pgp::Any::Cleartext(_)]);
 }
+
+#[test]
+fn load_csf_starts_with_newline() {
+    // Load a CSF message with the message '\ntest\n'
+    let empty_csf_msg =
+        std::fs::read_to_string("tests/unit-tests/csf/starts-with-newline.csf").unwrap();
+
+    let (any, _) = pgp::Any::from_armor(empty_csf_msg.as_bytes()).expect("from_armor");
```
dignifiedquire, line 25: can you check the header information is correct as well

hko-s: Hmm, good point. But the CSF hash headers ("Hash: SHA512", here) should go into `CleartextSignedMessage::hashes`, which I don't think tests have access to?

At least the test can easily assert that the `Any::from_armor` header result (which comes from the SIGNATURE block) is empty. I just pushed that.

tests/any_test.rs
```diff
@@ -15,3 +15,27 @@ fn load_empty_csf() {
...
+    assert!(headers.is_empty());
+
+    match any {
+        pgp::Any::Cleartext(csm) => {
+            // Message payload with line endings normalized (CI on windows produces CRLF line endings)
+            let payload = csm.text().replace("\r\n", "\n");
+
+            assert_eq!(payload, "\ntest\n");
+
+            // TODO: verify signature?
```
dignifiedquire, line 37: lets also verify the signature? otherwise it is hard to know everything is actually parsed correct?

hko-s: alright, done. no harm in doing that :)

PR #479, opened 2025-03-01
don't store significant information in literal metadata

hko-s:
The filename and creation-time fields in the Literal Data Packet are not cryptographically protected, and are considered a misfeature of sorts. The guidance in RFC 9580 is to "avoid storing any significant data in these fields" (see https://www.rfc-editor.org/rfc/rfc9580.html#section-5.9-5).

This PR implements this guidance in the message builder.


dignifiedquire:
I would understand this guidance differently, and keep storing the filename, it is still useful, even if not verified.

dignifiedquire:
I guess

>  It SHOULD set the filename to the empty string (encoded as a single zero octet) and the timestamp to zero (encoded as four zero octets).
 does imply this..

src/packet/literal_data.rs
```diff
@@ -41,8 +40,18 @@ pub struct LiteralData {
 pub struct LiteralDataHeader {
     pub mode: DataMode,
     /// The filename, may contain non utf-8 bytes
-    pub file_name: Bytes,
-    pub created: DateTime<Utc>,
+    file_name: Bytes,
```
dignifiedquire, line 43: If you make them private, add at least accessor methods. Also make all fields private in that case

dignifiedquire:
ci is very sad

PR #480, opened 2025-03-01
builder tweaks

hko-s:


src/composed/message/builder.rs
```diff
@@ -408,7 +408,7 @@ impl<'a, R: Read, E: Encryption> Builder<'a, R, E> {
     ///
     /// Defaults to `DataMode::Binary`
     ///
-    /// If the mode is set to `DataMode::Utf8`, line endings will be normalized.
+    /// If the mode is set to `DataMode::Utf8` (or `DataMode::Text`), the [SignatureType] will be `Text`, and line endings will be hashed in normalized form.
```
dignifiedquire, line 411: text must not be generated in the latest rfc asfaik

hko-s: Agreed. However, I think the builder doesn't currently stop users from producing `DataMode::Text`.

So the comment should probably say that they shouldn't - but I think the code should still handle such setting gracefully. Or error out for DataModes that are illegal?

dignifiedquire: fair, lets do the right thing, we can always add an error

PR #481, opened 2025-03-03
make algorithm enums non-exhaustive

### LINKED ISSUE #465: Make enums non-exhaustive, where appropriate
To allow addition of e.g. a `PublicKeyAlgorithm` without semver breakage.

hko-s:
fixes #465

### DIFF
--- a/src/crypto/aead.rs
+++ b/src/crypto/aead.rs
@@ -27,6 +27,7 @@ pub type Aes192Gcm = AesGcm<Aes192, U12>;
 #[derive(Debug, PartialEq, Eq, Copy, Clone, FromPrimitive, IntoPrimitive)]
 #[cfg_attr(test, derive(proptest_derive::Arbitrary))]
 #[repr(u8)]
+#[non_exhaustive]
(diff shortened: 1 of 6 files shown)


hko-s:
this seems straightforward enough

dignifiedquire:
I have thought about this some more, and I actually don't think I want to do this, there are not going to be that many changes to these that this needs to be shipped in patch releases imo

hko-s:
> I have thought about this some more, and I actually don't think I want to do this, there are not going to be that many changes to these that this needs to be shipped in patch releases imo

Reasons to do this might include:

- Ed448
- PQC
- "packet type 20" read support

dignifiedquire:
Yep, but looking at the current state, I don't think the api will fully stabilize anyway before we had those.

hko-s:
I was hoping that it's reasonably possible to keep the rPGP API semver stable for a while (months). In that case, having non-exhaustive could be useful. If we're not trying to stay semver stable, then non-exhaustive of course doesn't add much value.

PR #486, opened 2025-03-06
Readers

hko-s:
Opening a PR to add comments onto :)


src/lib.rs
```diff
@@ -7,7 +7,6 @@
 //! [signing and verifying with external hashing]: crate::composed::signed_key
 //! [packet based signing and verifying]: crate::packet
 
-#![cfg_attr(not(feature = "mmap"), forbid(unsafe_code))]
```
hko-s, line 10: historically, there was a

`#![forbid(unsafe_code)]`

here.

we should probably put that back in here?

src/composed/message/decrypt.rs
```diff
@@ -37,6 +37,22 @@ pub enum PlainSessionKey {
         #[debug("..")]
         key: Vec<u8>,
     },
+
+    /// If the version is unknown, it will be matched to the packets
+    Unknown {
+        sym_alg: SymmetricKeyAlgorithm,
```
hko-s, line 43: I'm confused and will need to read up (so this is a note-to-self):

I wonder if `sym_alg` should be an `Option`.

src/lib.rs
dignifiedquire: ups, yeah

src/composed/message/decrypt.rs
dignifiedquire: if is `None` you would already know it is a v6

hko-s: I have no strong feelings about this. But we could choose to support handling seipdv1 session keys in the absence of an algorithm id.

src/composed/message/types.rs
```diff
@@ -1,54 +1,295 @@
...
-    Compressed(CompressedData),
+/// Encrypted Message | Signed Message | Compressed Message | Literal Message.
+#[derive(Debug)]
+#[allow(clippy::large_enum_variant)]
+pub enum Message<'a> {
+    /// Literal Message: Literal Data Packet.
+    Literal {
+        reader: LiteralDataReader<Box<dyn DebugBufRead + 'a>>,
+        /// is this a nested message?
+        is_nested: bool,
```
hko-s, line 37: As I understand the format, only the `SignedOnePass` variant should have an `is_nested` flag.

Only the one-pass signature packet contains this flag (which is a weird flag, and might be a misfeature. It stems from https://www.rfc-editor.org/rfc/rfc2440.html#section-5.4).

To add context for readers: The idea of the "nested" flag is to configure the hasher for that one-pass signature between two modes of operation:

- hashing just the unwrapped literal data (this is normal)
- hashing the unwrapped literal data, as well as the packets that the inner message contains (i don't think we should implement this. it's unclear to me if there has ever been any real world use of this feature - i think we might even want to error if a caller tries to validate such signatures)

PR #487, opened 2025-03-06
ci: run doctests

hko-s:
nextest doesn't currently run doctests, also see https://github.com/rpgp/rpgp/pull/455#issuecomment-2703215294


dignifiedquire (approved):
thanks

PR #497, opened 2025-03-24
fix: check that fixed part is as long as it claims to be

hko-s:
This is tested in the interop suite for the final fixed chunk in partial encoded bodies.

### DIFF
--- a/src/composed/message/reader/limited.rs
+++ b/src/composed/message/reader/limited.rs
@@ -2,11 +2,7 @@ use std::io::{self, BufRead, Read};
 
 #[derive(Debug)]
 pub enum LimitedReader<R: BufRead> {
-    Fixed {
-        // Number of bytes we (still) expect to read from this Fixed
-        expect_data: u32,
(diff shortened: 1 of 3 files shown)


src/composed/message/reader/limited.rs
```diff
@@ -2,7 +2,11 @@ use std::io::{self, BufRead, Read};
 
 #[derive(Debug)]
 pub enum LimitedReader<R: BufRead> {
-    Fixed { reader: io::Take<R> },
+    Fixed {
+        // Number of bytes we (still) expect to read from this Fixed
+        expect_data: u32,
+        reader: io::Take<R>,
```
dignifiedquire, line 8: you don't need to track this externally, you only need to check if the reader consumed everything, as `io::Take` already tracks how many bytes are still left

just check if https://doc.rust-lang.org/std/io/struct.Take.html#method.limit is `0` at the end

hko-s: Right. I cleaned this up.

PR #499, opened 2025-03-25
guardrails against sed packet decryption

hko-s:
We should probably not decrypt SED packets by default.

But I'm not sure how best to expose this in the API. Thoughts welcome.

### DIFF
--- a/src/composed/message/types.rs
+++ b/src/composed/message/types.rs
@@ -490,9 +490,32 @@ impl<'a> Edata<'a> {
         }
     }
 
-    pub fn decrypt(&mut self, key: &PlainSessionKey) -> Result<()> {
+    /// Decrypts only SEIPD (v1 or v2), errors for SED packets
+    /// (this avoids decrypting malleable ciphertext)
+    pub fn decrypt_protected(&mut self, key: &PlainSessionKey) -> Result<()> {
         let protected = self.tag() == Tag::SymEncryptedProtectedData;
-        debug!("decrypting protected = {:?}", protected);
+        debug!("decrypt_protected: protected = {:?}", protected);
+
+        match self {
+            Self::SymEncryptedProtectedData { reader } => {
+                reader.decrypt(key)?;
+            }
+            Self::SymEncryptedData { .. } => {
+                // SED packets are malleable, decrypting them should only be necessary for historical data
+                bail!("Decryption of SymEncryptedData is discouraged")
+            }
+        }
+
+        Ok(())
+    }
+
+    /// HAZMAT: Decrypts SEIPD (v1 or v2) and SED packets.
+    ///
+    /// Decrypting (malleable) SED packets is not necessary for most use cases, except for
+    /// historical data.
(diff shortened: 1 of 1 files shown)


src/composed/message/types.rs
```diff
@@ -490,9 +490,32 @@ impl<'a> Edata<'a> {
         }
     }
 
-    pub fn decrypt(&mut self, key: &PlainSessionKey) -> Result<()> {
+    /// Decrypts only SEIPD (v1 or v2), errors for SED packets
+    /// (this avoids decrypting malleable ciphertext)
+    pub fn decrypt_protected(&mut self, key: &PlainSessionKey) -> Result<()> {
```
dignifiedquire, line 495: I would keep this as `decrypt`

src/composed/message/types.rs
```diff
@@ -490,9 +490,32 @@ impl<'a> Edata<'a> {
...
+        }
+
+        Ok(())
+    }
+
+    /// HAZMAT: Decrypts SEIPD (v1 or v2) and SED packets.
+    ///
+    /// Decrypting (malleable) SED packets is not necessary for most use cases, except for
+    /// historical data.
+    pub fn decrypt_any(&mut self, key: &PlainSessionKey) -> Result<()> {
```
dignifiedquire, line 516: maybe `decrypt_legacy`?

hko-s: right, that's probably reasonable. it just felt wrong because it *also* decrypts seipd.

renamed both.

src/composed/message/types.rs
```diff
@@ -784,7 +784,7 @@ impl<'a> Message<'a> {
                     return Err(Error::MissingKey);
                 };
 
-                edata.decrypt_protected(&session_key)?;
+                edata.decrypt(&session_key)?;
```
dignifiedquire, line 787: this is a problem, as this means you can't use the legacy decryption with any other packet combination currently

hko-s: the reader API currently doesn't expose a convenient way to decrypt legacy SED containers.
is that what you're saying, or are you thinking about a different question?

it's unclear to me if we want to expose this functionality, until someone asks for it (and it is already possible for consumers of rpgp to "manually" get the edata, and call the legacy decryption function on it.)

hko-s: For scale: SEIPDv1 was introduced in RFC 4880, November 2007. It says

> An implementation MUST support decrypting these packets and SHOULD prefer generating them to the older Symmetrically Encrypted Data packet when possible

So creating SED packets has been strongly discourages for 17+ years.

I think in practice they are only relevant for ancient backups or ancient email history.
It's unclear to me what kind of API software that handles such data would even need to handle the data in a reasonable and safe way.

So I'd tend to suggest that we postpone designing an API for it.

(Just like we'll need to think about more an advanced ESK decryption API some day - which might involve callbacks, or some other flexible interaction between application and rpgp).

dignifiedquire: fair enough

PR #500, opened 2025-03-25
message parsing - discard excess compressed packet content

hko-s:


src/composed/message/types.rs
```diff
@@ -915,8 +915,18 @@ impl<'a> Message<'a> {
 
         match inner {
             MessageReader::Compressed(r) => {
-                let inner_reader = r.get_mut().get_mut();
-                let parser = crate::packet::PacketParser::new(inner_reader);
+                let compressed_body_reader = r.get_mut();
+
+                // discard excess data in the compressed packet
+                let mut sink = vec![];
+                let excess = compressed_body_reader.read_to_end(&mut sink)?;
```
dignifiedquire, line 922: this allocates, please don't do that, we have `drain` for this: https://github.com/rpgp/rpgp/blob/feat-large-files/src/parsing_reader.rs#L100

hko-s: thanks for the pointer!

hko-s: fixed

PR #502, opened 2025-03-28
various

hko-s:


src/packet/secret_key_parser.rs
```diff
@@ -14,7 +15,23 @@ fn parse_pub_priv_fields<B: BufRead>(
     pub_len: Option<usize>,
     mut i: B,
 ) -> Result<(PublicParams, SecretParams)> {
-    let pub_params = PublicParams::try_from_reader(typ, pub_len, &mut i)?;
+    let pub_params = match pub_len {
+        Some(pub_len) => {
+            // Use the pub_len hint to make sure we consume no more or less
+            let public = i.take_bytes(pub_len)?;
```
dignifiedquire, line 21: this should be `read_take`, no need to buffer this in memory

Issue #507, opened 2025-03-29
Use anonymous recipient for generated packets


link2xt:
Currently rPGP supports receiving packets with "anonymous recipient" (all-zero wildcard keyid/fingerprint), but I have not found any way to generate them.

But key_id is always included in the generated packets:
https://github.com/rpgp/rpgp/blob/be1902930d08b1cd63d5e154844f6468c69b7aa6/src/packet/public_key_encrypted_session_key.rs#L106

Same for the fingerprint in v6 PKESK:
https://github.com/rpgp/rpgp/blob/be1902930d08b1cd63d5e154844f6468c69b7aa6/src/packet/public_key_encrypted_session_key.rs#L132

See https://www.rfc-editor.org/rfc/rfc9580.html#section-5.1.8-1 for reference.

This will be needed for Delta Chat: https://support.delta.chat/t/does-delta-chat-use-hidden-recipient/3651

hko-s:
Right. The message builder could get an additional flag for encryption that tells it to produce PKESK without a recipient identifier. That shouldn't be hard. I think.

dignifiedquire:
this is now possible

PR #510, opened 2025-03-29
feat: add encrypt_to_key_anonymous functions

hko-s:
We might want to DRY this code, but I wanted to sketch an API to deal with #507, to discuss

### DIFF
--- a/src/composed/message/builder.rs
+++ b/src/composed/message/builder.rs
@@ -34,7 +34,8 @@ use crate::{
     ser::Serialize,
     types::{
         CompressionAlgorithm, Fingerprint, KeyId, KeyVersion, PacketHeaderVersion, PacketLength,
-        Password, SecretKeyTrait, StringToKey, Tag,},
+        Password, SecretKeyTrait, StringToKey, Tag,
+    },
     util::{fill_buffer, TeeWriter},
 };
 
@@ -1238,6 +1239,7 @@ impl<R: std::io::Read> std::io::Read for SignGenerator<'_, R> {
 
 #[cfg(test)]
 mod tests {
+    use nom::AsBytes;
     use rand::{Rng, SeedableRng};
     use rand_chacha::ChaCha20Rng;
     use testresult::TestResult;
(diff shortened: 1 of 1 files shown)


src/composed/message/builder.rs
```diff
@@ -313,6 +313,29 @@ impl<R: Read> Builder<'_, R, EncryptionSeipdV1> {
...
+        let mut pkes = PublicKeyEncryptedSessionKey::from_session_key_v3(
+            &mut rng,
+            &self.encryption.session_key,
+            self.encryption.sym_alg,
+            pkey,
+        )?;
+
+        // Blank out the recipient id
+        if let PublicKeyEncryptedSessionKey::V3 { id, .. } = &mut pkes {
+            *id = KeyId::new([0, 0, 0, 0, 0, 0, 0, 0]);
```
dignifiedquire, line 332: don't we have a constant or function for this?

src/composed/message/builder.rs
```diff
@@ -313,6 +313,29 @@ impl<R: Read> Builder<'_, R, EncryptionSeipdV1> {
         Ok(self)
     }
 
+    /// Encrypt to a public key, but leave the recipient field unset
+    pub fn encrypt_to_key_anonymous<RAND, K>(mut self, mut rng: RAND, pkey: &K) -> Result<Self>
```
dignifiedquire, line 317: why not make this a function on the builder, that modifies the state, rather than changing the finalizer method?

src/composed/message/builder.rs
hko-s: d'oh. we do!

src/composed/message/builder.rs
hko-s: I'm not sure what exactly you have in mind?

My idea was that some applications might want to anonymize only *some* of the recipients. So I sketched this API where the caller decides per recipient public key if that one should be anonymized.

(i.e.: a message could have `n` PKESK, but only one of them might want to be wildcard-ed)

dignifiedquire: sorry, I missread the code, ignore this comment

dignifiedquire:
api looks good, but a test would be nice

src/composed/message/builder.rs
```diff
@@ -1193,6 +1239,7 @@ impl<R: std::io::Read> std::io::Read for SignGenerator<'_, R> {
 
 #[cfg(test)]
 mod tests {
+    use nom::AsBytes;
```
dignifiedquire, line 1242: I don't think you want to use this

hko-s: Oh no ... O_o ... I don't. Fixed.

PR #516, opened 2025-04-01
happier cargo doc

hko-s:
partial fix of the complaints of `cargo doc`


dignifiedquire (approved):

r10s:
yeah, i am always making clippy happier. it's a good part of our job to keep the tools happy!

PR #521, opened 2025-04-08
Rename OpenPGP files from .gpg to .pgp

guillemj:
These files contain OpenPGP data, and are not vendor specific, so naming them with an extension after GnuPG creates a lock-in in detriment to the other multiple OpenPGP implementations, including the rPGP one.

Use the vendor neutral .pgp filename extension instead.


dignifiedquire:
thanks

PR #525, opened 2025-04-09
refactor: allow better error handling on the MessageBuilder

dignifiedquire:


hko-s (approved):
While this is making the `Builder` a bit more irregular to use, I appreciate that it enables recovering from errors (like trying to encrypt to a key which rPGP can't actually encrypt to) without losing the entire builder state :partying_face:

PR #526, opened 2025-04-09
refactor: cleanup deps in Cargo.toml

dignifiedquire:
- got to remove one unused dep
- also made sure that minimum versions still compiles on latest rust nightly


dkg:
looks like this drops `bytes-utils`, but it also explicitly adds a versioned dependency on `sha1-asm`.  is there a reason for this addition?

dignifiedquire:
> adds a versioned dependency on `sha1-asm`. is there a reason for this addition?

yes, because the default selection on minimal version doesn't compile with latest rust anymore

dignifiedquire:
> looks like this drops `bytes-utils`

yep, it turned out to be unused

hko-s:
This all looks reasonable enough, and rsop is running without complaint on this PR, fwiw

PR #539, opened 2025-04-17
add orderly support for V6 keygen

hko-s:
Some main concerns for v6 key generation include:

- "certificate metadata" should not be stored on user id self-signatures anymore. Certificate metadata (preferred algorithms, key expiration, key flags for the primary,etc) should live in direct key signatures on the primary
- relatedly, the "primary user id" concept is much less relevant in v6 keys
- user ids are overall optional, in v6 keys


hko-s:
After working through this, I'm now less sure that the "primary user id" needs so much special handling in the API.

This API seems ok to me, but I could now imagine an alternative (simpler) approach where the generation APIs don't have a specific "primary" user id parameter, but instead just one Vec, and we implicitly use the first User ID entry as "primary" during key generation.

hko-s:
> After working through this, I'm now less sure that the "primary user id" needs so much special handling in the API.
> 
> This API seems ok to me, but I could now imagine an alternative (simpler) approach where the generation APIs don't have a specific "primary" user id parameter, but instead just one Vec, and we implicitly use the first User ID entry as "primary" during key generation.

But then, the current changes seem reasonable enough, and might just cause the least disruption in the parts of the public API that applications use, so maybe that's a good tradeoff? I'm open to suggestions.

src/composed/key/builder.rs
```diff
@@ -19,25 +19,39 @@ use crate::{
     },
     errors::Result,
     packet::{self, KeyFlags, PubKeyInner, UserAttribute, UserId},
-    types::{
-        self, CompressionAlgorithm, PlainSecretParams, PublicParams, RevocationKey, S2kParams,
-    },
+    types::{self, CompressionAlgorithm, PlainSecretParams, PublicParams, S2kParams},
 };
 
+/// A SecretKeyParams
```
dignifiedquire, line 25: that’s an odd comment

src/composed/key/builder.rs
```diff
@@ -19,25 +19,39 @@ use crate::{
...
 #[derive(Debug, PartialEq, Eq, Builder)]
 #[builder(build_fn(validate = "Self::validate"))]
 pub struct SecretKeyParams {
+    // -- OpenPGP key version of primary
```
dignifiedquire, line 29: why not a doc comment?

src/composed/key/shared.rs
```diff
@@ -12,71 +12,47 @@ use crate::{
         UserAttribute, UserId,
     },
     ser::Serialize,
-    types::{
-        CompressionAlgorithm, KeyVersion, Password, PublicKeyTrait, RevocationKey, SecretKeyTrait,
-    },
+    types::{CompressionAlgorithm, KeyVersion, Password, PublicKeyTrait, SecretKeyTrait},
 };
 
+/// A KeyDetails specifies associated user id and attribute components, and some metadata for
```
dignifiedquire, line 18: please write This or sth like that, repeating the type is a horrible convention golang made popular for some unknown confused reason

src/composed/key/builder.rs
hko-s: Indeed. I probably wanted to write more, and then lost track :shrug: ... dropping it.

src/composed/key/builder.rs
hko-s: I extended the pre-existing `// --` comments that structured the fields, so that's where this pattern came from.

But agreed, this one (and the next) should definitely be a doc comment.

src/composed/key/builder.rs
```diff
@@ -47,109 +61,154 @@ pub struct SecretKeyParams {
...
                 }
             }
             Some(KeyType::ECDSA(curve)) => {
-                if let Some(can_encrypt) = self.can_encrypt {
-                    if can_encrypt {
-                        return Err("ECDSA can only be used for signing keys".into());
-                    }
-                };
+                if self.can_encrypt == Some(true) {
+                    return Err("ECDSA can only be used for signing keys".into());
```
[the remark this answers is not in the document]
hko-s: sounds good, done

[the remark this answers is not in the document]
hko-s: The function now also checks subkeys for keytype validity.

src/composed/key/builder.rs
```diff
@@ -197,22 +268,34 @@ impl SecretKeyParams {
...
+
+        let primary_user_id = match self.primary_user_id {
+            None => None,
+            Some(id) => Some(UserId::from_str(Default::default(), id)?),
+        };
+
+        let features = if !self.feature_seipd_v2 {
+            vec![0x01] // SEIPDv1
+        } else {
+            vec![0x01 | 0x08] // SEIPDv1 and SEIPDv2
```
dignifiedquire, line 281: can we make these constants/enums please?

src/composed/key/shared.rs
```diff
@@ -92,42 +68,67 @@ impl KeyDetails {
         K: SecretKeyTrait,
         P: PublicKeyTrait + Serialize,
     {
-        let keyflags = self.keyflags;
-        let preferred_symmetric_algorithms = self.preferred_symmetric_algorithms;
-        let preferred_hash_algorithms = self.preferred_hash_algorithms;
-        let preferred_compression_algorithms = self.preferred_compression_algorithms;
-        let preferred_aead_algorithms = self.preferred_aead_algorithms;
-        let revocation_key = self.revocation_key;
+        // Technically this should probably check for >= version 6, for at least
```
dignifiedquire, line 71: unfinished thought

src/composed/signed_key/secret.rs
```diff
@@ -190,6 +190,22 @@ impl SignedSecretKey {
         )
     }
 
+    pub fn signed_public_key(&self) -> SignedPublicKey {
```
dignifiedquire, line 193: doc comment?

src/composed/signed_key/secret.rs
```diff
@@ -285,6 +301,10 @@ impl SignedSecretSubKey {
         self.key.encrypt(rng, plain, typ)
     }
 
+    pub fn signed_public_key(&self) -> SignedPublicSubKey {
```
dignifiedquire, line 304: doc comment?

src/composed/signed_key/secret.rs
dignifiedquire: maybe `as_signed_public_key`or sth like that

src/composed/signed_key/shared.rs
```diff
@@ -133,8 +133,73 @@ impl SignedKeyDetails {
...
     }
 
+    /// Derive a `KeyDetails` from a `SignedKeyDetails`.
+    /// This is more of a heuristic than a surefire transformation.
+    ///
+    /// Note: this is also not checking the cryptographic validity of the signatures it's
+    /// evaluating, and e.g. not trying to find the newest signature of each type.
+    /// This whole functions is an optimistic heuristic, not a rigorous evaluation.
+    ///
+    /// Don't rely too hard on its outputs!
```
dignifiedquire, line 143: That's really unhelpful, I have no idea when I should use this, if I should use this now :/

src/composed/signed_key/shared.rs
```diff
@@ -133,8 +133,73 @@ impl SignedKeyDetails {
...
+                .iter()
+                .find(|u| u.is_primary())
+                .map_or_else(|| self.users.first(), Some);
+
+            KeyDetails::new(
+                primary_user.map(|su| su.id.clone()),
+                self.users
+                    .iter()
+                    .filter(|u| Some(&u.id) != primary_user.map(|pri_su| &pri_su.id)) // drop the primary
+                    .map(|u| u.id.clone())
```
dignifiedquire, line 189: `filter_map`would be a bit nicer

src/composed/signed_key/shared.rs
```diff
@@ -145,45 +210,52 @@ impl SignedKeyDetails {
...
-            KeyDetails::new_direct(
+            KeyDetails::new(
+                None,
                 self.users.iter().map(|u| u.id.clone()).collect(),
                 self.user_attributes
                     .iter()
                     .map(|a| a.attr.clone())
                     .collect(),
                 KeyFlags::default(),
+                vec![0x01], // SEIPDv1
```
dignifiedquire, line 254: constant

src/composed/key/shared.rs
```diff
@@ -12,71 +12,47 @@ use crate::{
...
+    primary_user_id: Option<UserId>,
+    non_primary_user_ids: Vec<UserId>,
     user_attributes: Vec<UserAttribute>,
     keyflags: KeyFlags,
+    features: Vec<u8>,
     preferred_symmetric_algorithms: SmallVec<[SymmetricKeyAlgorithm; 8]>,
     preferred_hash_algorithms: SmallVec<[HashAlgorithm; 8]>,
     preferred_compression_algorithms: SmallVec<[CompressionAlgorithm; 8]>,
     preferred_aead_algorithms: SmallVec<[(SymmetricKeyAlgorithm, AeadAlgorithm); 4]>,
-    revocation_key: Option<RevocationKey>,
```
dignifiedquire, line 29: so why is this gone entirely?

src/composed/key/builder.rs
hko-s: I made a bitfield-based `Features` type that can be `Into<Vec<u8>>`'d.

We could also use that `Features` type instead of the current `Vec<u8>` representation of the value of the features subpacket?

src/composed/key/public.rs
```diff
@@ -131,7 +129,13 @@ impl PublicSubkey {
         };
 
         config.hashed_subpackets = hashed_subpackets;
-        config.unhashed_subpackets = unhashed_subpackets;
+
+        // If the version of the issuer is greater than 4, this subpacket MUST NOT be included in
+        // the signature.
+        if u8::from(sec_key.version()) <= 4 {
```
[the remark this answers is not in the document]
hko-s: good point. that's much nicer indeed.

src/composed/key/shared.rs
hko-s: :chipmunk: ... cleaned up by dropping the whole thing.
it was only used in one place anymore, anyway.

src/composed/signed_key/shared.rs
hko-s: I tend to think we should try to get rid of this function. But I believe DC relies on it right now, so I didn't want to rush it.

For now I made a best-effort attempt to implement the same semantics for v6 as the pre-existing code had for v4 keys. However, I think there is no safe general way of implementing this function, for arbitrary certificates:

Going from a `KeyDetails` to a `SignedKeyDetails` is a bit of a trapdoor thing: We can safely move in the "forward" direction, because we get to pick one approach for how we want to model the data from the `KeyDetails`
But the "backwards" direction is much weirder, since OpenPGP certificates can contain model the metadata in many different ways. Including utterly unreasonable ones, especially with v4 keys. For example contradictory preferences on the primary user id vs. on a direct key signature.

And even if we just wanted to implement a "best effort" heuristic, we'd need to pull in a significant subset of the high level semantics code that currently lives in rpgpie.

So I think we should just not. I propose opening a ticket for removing this function to coordinate with DC about how we could replace this with a safer function that covers DC's needs.

src/composed/key/shared.rs
hko-s: My understanding is that the `RevocationKey` mechanism has been deemed a misfeature, and was never really used.

I see `KeyDetails` as a helper that exposes a curated subset of the wide range of things one could set during OpenPGP key generation. My suggestion would be to drop the `RevocationKey` mechanism from our curated set of things that rPGP users can construct easily.

src/composed/signed_key/shared.rs
hko-s: i changed this, but am not sure if my change is nicer.

maybe there's a better idea that i'm not coming up with?

src/composed/key/shared.rs
hko-s: Ah, and RFC 9580 says (https://www.rfc-editor.org/rfc/rfc9580.html#name-revocation-key-deprecated):

> This mechanism is deprecated. Applications MUST NOT generate such a subpacket.

(That said, while I think this change is a reasonable bit of cleanup, I don't feel super strongly about it.)

dignifiedquire: is it still possible to construct manually? I think so right? in that case I am okay with removing it from here

src/composed/key/builder.rs
dignifiedquire: very nice, yes lets use it where we can

src/composed/key/shared.rs
hko-s: I think it should be, yes.

As a general rule, I'd say if there is any blocker to making weird/arbitrary Transferable Secret Keys from scratch with rPGP, we should consider that a bug in the API, and fix it.

That rule should include the `RevocationKey` Subpacket.

src/composed/key/builder.rs
hko-s: Alright, sounds good to me.

Even though it also sounds a bit annoying: We will probably want to make the same contortions in `Features` that `KeyFlags` is making, to be able to roundtrip weird subpacket shapes without accidentally breaking signatures. I'll do that later today.

src/composed/signed_key/shared.rs
dignifiedquire: looks like I expected

hko-s: cool, thanks :)

src/composed/signed_key/shared.rs
hko-s: I opened #546 to track this

src/composed/key/builder.rs
hko-s: I wrote some code in https://github.com/rpgp/rpgp/pull/539/commits/eb25dcff01423226080af356c79e61641959e2ce, and hope it doesn't break too many things :scream_cat: 

Also, I'm idly wondering if `Features::default` and `Features::new` should by default set seipdv1 support to `true`. This is definitely an expected default, but I wonder if that's how we want the type to behave.

src/packet/signature/de.rs
```diff
@@ -589,9 +589,9 @@ fn rev_reason<B: BufRead>(mut i: B) -> Result<SubpacketData> {
 /// Parse a Features subpacket
 /// Ref: https://www.rfc-editor.org/rfc/rfc9580.html#name-features
 fn features<B: BufRead>(mut i: B) -> Result<SubpacketData> {
-    let features = SmallVec::from_slice(&i.rest()?);
+    let features = i.rest()?;
```
dignifiedquire, line 592: what if this is more than 1 byte?

src/packet/signature/types.rs
```diff
@@ -1199,13 +1197,106 @@ pub struct KnownKeyFlags {
     _padding2: u8,
 }
 
+#[derive(Clone, PartialEq, Eq, Debug)]
```
dignifiedquire, line 1200: docs please

tests/key_test.rs
```diff
@@ -656,7 +656,7 @@ fn test_parse_details() {
             Subpacket::regular(SubpacketData::PreferredSymmetricAlgorithms(p_sym_algs)).unwrap(),
             Subpacket::regular(SubpacketData::PreferredHashAlgorithms(p_hash_algs)).unwrap(),
             Subpacket::regular(SubpacketData::PreferredCompressionAlgorithms(p_com_algs)).unwrap(),
-            Subpacket::regular(SubpacketData::Features(smallvec![1])).unwrap(),
+            Subpacket::regular(SubpacketData::Features((&[0x01][..]).into())).unwrap(),
```
dignifiedquire, line 659:
```rust
KnownFeatures::from(0x01).into()
``` 

would be nicer

src/composed/key/builder.rs
```diff
@@ -47,107 +60,160 @@ pub struct SecretKeyParams {
...
+    version: types::KeyVersion,
+
+    // -- Asymmetric algorithm of this subkey
     key_type: KeyType,
 
+    // -- Keyflags for this subkey
     #[builder(default)]
     can_sign: bool,
     #[builder(default)]
-    can_certify: bool,
```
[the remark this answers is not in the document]
hko-s: I dropped it from the subkey builder.

My understanding is that certification-capable subkeys are not a thing, in the real world, as far as I understand, and I don't think we should make any.

In usual practice, certifications are only issued by the primary key

src/composed/key/builder.rs
```diff
@@ -47,107 +60,160 @@ pub struct SecretKeyParams {
...
+                    Some(&subkey.key_type),
+                    Some(subkey.can_sign),
+                    Some(subkey.can_encrypt),
+                    Some(subkey.can_authenticate),
+                )?;
+            }
+        }
+
+        if self.version == Some(types::KeyVersion::V4) && self.primary_user_id.is_none() {
+            return Err("V4 keys must have a primary User ID".into());
```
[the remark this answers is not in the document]
hko-s: This is one of those places in OpenPGP that is incredible flexible and incredibly underspecified.

The idea in v4 keys is that key metadata comes from self-signatures on the primary user id. What exactly that means, and how it should be implemented is left as an exercise for the reader.

Historically, primary-ness of user ids has also been inferred e.g. by "there's only one, so it's probably primary". I don't think there's any benefit in rPGP being able to conveniently generate this kind of Certificate, so this code rejects building this.

It's a moderately opinionated approach, but I think it's reasonable.

src/composed/key/builder.rs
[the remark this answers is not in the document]
hko-s: Ah, I think that particular proposal is actually mistaken:

I believe the "Certification" key flag does *not* allow issuing self-signatures to bind subkeys. It only allows issuing third party certifying signatures.

Either way, I think certification-capable subkeys would add complexity that is hard to justify. While it might enable some niche workflow, it would come at the cost of making the OpenPGP ecosystem more brittle. A poor tradeoff, I think.

src/packet/signature/types.rs
```diff
@@ -1199,6 +1197,153 @@ pub struct KnownKeyFlags {
...
+/// Ref <https://www.rfc-editor.org/rfc/rfc9580.html#name-features>
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub struct Features {
+    /// Handles the first byte.
+    /// Can be None to encode Features subpackets that are 0 byte long.
+    first: Option<KnownFeatures>,
+
+    /// Any additional features bytes.
+    /// (Must be empty if "known" is None.)
+    rest: Vec<u8>,
```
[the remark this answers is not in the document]
hko-s: The `rest` field could be needed and used to round-trip (and correctly hash) data from other implementations.

hko-s:
I adjusted a unit test to exercise the latest mini-change (in 2aa1b6524e3d868b78a5f0c15417d53c1c0e53a9), and think this wraps up the PR.

Once CI is green, I'd tend to squash/merge soon.

src/composed/key/shared.rs
```diff
@@ -8,75 +8,51 @@ use crate::{
...
-            preferred_compression_algorithms,
-            preferred_aead_algorithms,
-            revocation_key,
-        )
-    }
-
-    /// The primary UserId (if any) is expected to be contained in `user_ids`
-    #[allow(clippy::too_many_arguments)] // FIXME
-    pub(crate) fn new_direct(
+    pub(crate) fn new(
```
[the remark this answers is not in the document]
hko-s: Uhh, good question if that change of visibility is a good idea. I'm not sure that I made the change intentionally.

That said, the normal way to make a `KeyDetails` is via `SecretKeyParams::generate`, which still works using the public API.

I don't think there's a particularly strong reason to make `KeyDetails::new` private. However, it is a bit more of a footgun than using `SecretKeyParams::generate`.
For example, it's possible to directly construct a KeyDetails for a v4 key without a primary UserId, which is not a very interoperable format - while `SecretKeyParams::generate` does consistency checks that prevent that kind of thing. Or set `KeyFlags` for the primary key that are logically inconsistent with the algorithm used (like setting the encryption capability for an Ed25519 primary).

Do you have a use case where constructing a `KeyDetails` directly is useful?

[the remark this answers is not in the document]
baloo: Yes!
This is useful for use-cases like [externally held keys](https://github.com/rpgp/rpgp/pull/554) (TPMs/HSMs/...) where you generate the private key externally.

You will generate the private key, then make a SignedPublicKey from it providing the `KeyDetails` to sign (`pgp::composed::PublicKey::sign`).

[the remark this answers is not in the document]
hko-s: Right. Thanks for pointing that out. Your point feels plausible to me (even though I'll admit I can't fully reason through all of the finer points). 

I opened #557 to undo this regression in the public API. I'm quite sure I didn't make that change with a strongly held conviction, and suspect it was more of an editor mishap than anything else.

PR #540, opened 2025-04-19
draft: message builder ergonomics

hko-s:


hko-s:
This PR builds on top of #539, which I think we should sort out first

hko-s:
As an application that uses these interfaces, depending on the context, requiring `&Password` or `Password` are both inconvenient :confused:

... when I already have a set of `&Password`, then it's not nice to have to provide a `Password`, but when I need to create a `Password` on the fly, then it's annoying to make sure that it lives long enough when an rPGP API requires a `&Password`.

I'm starting to wonder if what I really want instead is that `&Password` implements `Clone`.

hko-s:
I'm retiring this idea, it seems to be the wrong idea.

PR #541, opened 2025-04-20
wip: "imprint" fn

hko-s:
This is just a sketch for discussion.

We could consider having an "imprint" function roughly like this, and implementing "fingerprint" in terms of it.
I think the cost of that would mainly be that the Hasher is Boxed in this more generic approach?

### DIFF
--- a/src/packet/key/public.rs
+++ b/src/packet/key/public.rs
@@ -611,6 +611,18 @@ impl crate::packet::PacketTrait for PublicSubkey {
 }
 
 impl PubKeyInner {
+    /// An imprint of a public key packet is a generalisation of a fingerprint.
+    ///
+    /// It is calculated in the same way as the fingerprint, except that it MAY use a
+    /// digest algorithm other than the one specified for the fingerprint.
+    ///
+    /// See https://www.ietf.org/archive/id/draft-ietf-openpgp-replacementkey-03.html#name-key-imprints
+    ///
+    /// Note that imprints are intended as a special purpose tool. For most use cases, the OpenPGP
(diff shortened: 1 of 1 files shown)


hko-s:
I've pushed a commit to see how `fingerprint` looks when based on `imprint`

hko-s:
It's a bit sad how much `expect`ing is currently going on in this PR. I think it would be possible to make `imprint` nicer by linking the hash algorithm parameter and returning an array with a corresponding output size.

But it seems that neither the rustcrypto types nor the current rpgp hashing mechanism have suitable facilities for this. We could make them for this PR?

hko-s:
> Is it going to use `generic-array`? We already have `generic-array` dependency and it does not need to appear in the public API unless we expose the `imprint` function with a generic hash algorithm.

I didn't have a concrete idea, but making this nicer on the type-level would seem potentially worthwhile.

Hoping for @dignifiedquire to weigh in with thoughts/preferences :)

src/packet/key/public.rs
```diff
@@ -612,29 +610,26 @@ impl crate::packet::PacketTrait for PublicSubkey {
     }
 }
 
-impl KeyDetails for PubKeyInner {
-    fn version(&self) -> KeyVersion {
-        self.version
-    }
+impl PubKeyInner {
+    fn imprint(&self, h: HashAlgorithm) -> Result<Vec<u8>> {
```
dignifiedquire, line 614: lets add some docs to this function

hko-s: done

hko-s:
Closing in favor of #549

Issue #542, opened 2025-04-22
Fuzzing on main


hko-s:

hko-s:
```
Failing input:

	artifacts/1_message_from_bytes/crash-82b02bbac39a10c7b98d020f78153ffb75c94607

Output of `std::fmt::Debug`:

	[145, 0, 11, 4, 9, 0, 0, 4, 89, 89, 169, 89, 0, 163, 141, 0, 11, 4, 9, 0, 0, 0, 89, 163, 4, 89, 169, 89, 167, 89]

Reproduce with:

	cargo fuzz run 1_message_from_bytes artifacts/1_message_from_bytes/crash-82b02bbac39a10c7b98d020f78153ffb75c94607

Minimize test case with:

	cargo fuzz tmin 1_message_from_bytes artifacts/1_message_from_bytes/crash-82b02bbac39a10c7b98d020f78153ffb75c94607
```

hko-s:
This crash is the only one i got after running each fuzz test for a few tens of minutes. I'll make a unit test out of it.

(Fuzz tests 4 and 6 don't currently run, so I excluded those)

hko-s:
#543 contains unit tests with the two problematic messages from this fuzz test run

PR #543, opened 2025-04-22
test: add message reader fuzz tests

hko-s:
Tests as produced in #542


hko-s:
On second glance, it looks to me like the underlying problem is that the message reader expects some degree of reasonableness in SED package contents?

hko-s:
closed in favor of #547

PR #549, opened 2025-05-07
feat: imprint function 

dignifiedquire:
Implement an Imprint trait that calculates fingerprint-like identifiers for keys, using arbitrary hash functions.

Based on #541


src/packet/key/public.rs
```diff
@@ -612,29 +618,36 @@ impl crate::packet::PacketTrait for PublicSubkey {
...
+    /// digest algorithm other than the one specified for the fingerprint.
+    ///
+    /// See https://www.ietf.org/archive/id/draft-ietf-openpgp-replacementkey-03.html#name-key-imprints
+    ///
+    /// Note that imprints are intended as a special purpose tool. For most use cases, the OpenPGP
+    /// fingerprint is the most appropriate identifier for a certificate or a component key.
+    ///
+    /// For [HashAlgorithm::Sha1], rPGP uses [sha1_checked](https://crates.io/crates/sha1-checked)
+    /// with collision detection.
+    fn imprint<D: KnownDigest>(&self) -> Result<GenericArray<u8, D::OutputSize>> {
```
hko-s, line 634: I tend to think that we could just make this fn public, and be done with this PR.

It seems a bit scary to encourage people to make all kinds of weird digests, but ... :shrug: it's probably fine?

hko-s: I think the code should live on `PubKeyInner`, but to expose the functionality publicly, we'll additionally need to implement an access fn [like this](https://github.com/rpgp/rpgp/blob/4c84a565c691ef51e2564ac313a00d0d5aff047d/src/packet/key/public.rs#L920) on PublicKey and PublicSubkey

hko-s:
I've added a unit test and public API for getting an imprint for a `SignedPublicKey`.

TODO:
- check if other implementations calculate the same digest for this example.
- not sure if this API is nice

src/packet/key/public.rs
```diff
@@ -615,29 +625,36 @@ impl crate::packet::PacketTrait for PublicSubkey {
     }
 }
 
-impl KeyDetails for PubKeyInner {
-    fn version(&self) -> KeyVersion {
-        self.version
-    }
+impl PubKeyInner {
```
dignifiedquire, line 628: if we make it part of the public API it should be on the trait `KeyDetails` imho

hko-s: I considered that, and shied away from it.

My preference would be to expose the `imprint` fn in the public API (mostly for DC), but not to advertise it too hard. Putting it on `KeyDetails` would give it a rather prominent position, and make it look more like a thing that everyone should use?

I'm open to having my mind changed, but ... very mixed feelings.

hko-s: Note (mostly) to self: We should also expose the `imprint` function for subkeys.

Putting the function in the `KeyDetails` trait would make it available in all places, which would be nice in some ways. However, I'm still very concerned that this would make the imprint concept look way more "official" than it should.

hko-s: Actually, we can't add the `imprint` function to `KeyDetails` because that would break `PublicKeyTrait`s dyn compatibility :confused: 

I just added the imprint fn to `PublicSubkey` and `SignedPublicSubKey`, which I think covers all of the public key locations. But it's indeed a bit annoying to document this API instead of a trait :disappointed:

PR #553, opened 2025-05-09
docs: Add doctest example for listing User IDs in a key

wiktor-k:
This PR adds a minimal example for listing User IDs in a given certificate ("public key").

I've been recently asked on how to do that with rPGP and if you don't mind I'd answer each such a question by progressively improving code examples in rPGP itself :sweat_smile: 

Happy to receive any wording/code improvements here. I'm mostly interested in having good examples for folks who just "want to get job done". :wave:


src/packet/user_id.rs
```diff
@@ -20,6 +20,22 @@ use crate::{
...
+///
+/// # Examples
+///
+/// The following example lists [`UserId`]s in a certificate ("public key"):
+///
+/// ```
+/// # fn main() -> testresult::TestResult {
+/// use pgp::composed::{Deserializable, SignedPublicKey};
+///
+/// let spk = SignedPublicKey::from_reader_single(std::fs::File::open("tests/carol.pub.asc")?)?.0;
```
hko-s, line 32: Not sure if that's just me, but I'd find something like 

```
/// let file = std::fs::File::open("tests/carol.pub.asc")?;
/// let (spk, _armor) = SignedPublicKey::from_reader_single(file)?;
```

a bit easier to read (untested, syntax may be slightly broken).

hko-s:
Nice! I like the idea of adding more such doctest examples, generally, and I like this specific example.

I left a suggestion/nit in the diff, to see if you agree :)

hko-s (approved):

wiktor-k:
Yep, agreed. Exactly the feedback I was looking for. I'll adjust ASAP :)

PR #554, opened 2025-05-11
Baloo/signer helper

hko-s:


src/helper/mod.rs
```diff
@@ -0,0 +1,46 @@
...
+pub trait PgpPublicKey {
+    /// Algorithm ID for a given signing algorithm
+    const PGP_ALGORITHM: PublicKeyAlgorithm;
+
+    /// Public key encoding for a public key
+    fn pgp_parameters(&self) -> PublicParams;
+}
+
+/// Equivalent PGP hash algorithm for a given digest
+pub trait PgpHash {
```
dignifiedquire, line 27: this is a duplication of the trait in the imprint PR

hko-s: Cool! We should get the imprint PR merged and rebase this on that :rocket:

src/helper/rsa.rs
```diff
@@ -0,0 +1,177 @@
...
+/// [`signature::Signer`] backed signer for PGP.
+pub struct RsaSigner<T, D> {
+    inner: T,
+    public_key: PublicKey,
+    _digest: PhantomData<D>,
+}
+
+impl<D, T> fmt::Debug for RsaSigner<T, D> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("RsaSigner").finish()
```
wiktor-k, line 40: I wonder if it makes sense to constrain T and D to implement Debug and then derive Debug.

Alternatively make the debug at least print the public key member. 🤔

baloo: No we can't because `trait SecretKeyTrait: KeyDetails + std::fmt::Debug {}`, but we can print the publickey

baloo:
I rebased it over in #433 and added the support for the KnownHash

PR #555, opened 2025-05-15
handle embedded "back" signatures for signing capable subkeys

hko-s:
Signing capable subkeys need an extra primary key binding signature (also see https://openpgp.dev/book/signing_components.html#special-case-binding-signing-subkeys)

### DIFF
--- a/src/composed/key/public.rs
+++ b/src/composed/key/public.rs
@@ -6,7 +6,7 @@ use crate::{
     composed::{KeyDetails, SignedPublicKey, SignedPublicSubKey},
     crypto::{hash::HashAlgorithm, public_key::PublicKeyAlgorithm},
     errors::Result,
-    packet::{self, KeyFlags},
+    packet::{self, KeyFlags, Signature},
     ser::Serialize,
     types::{
         EskType, Fingerprint, KeyId, Password, PkeskBytes, PublicKeyTrait, PublicParams,
@@ -26,6 +26,11 @@ pub struct PublicKey {
 pub struct PublicSubkey {
     pub key: packet::PublicSubkey,
     pub keyflags: KeyFlags,
+
+    /// Embedded primary key binding signature, required for signing-capable subkeys.
+    ///
+    /// See <https://www.rfc-editor.org/rfc/rfc9580.html#sigtype-primary-binding>
+    pub embedded: Option<Signature>,
 }
 
 impl PublicKey {
@@ -87,8 +92,12 @@ impl Deref for PublicKey {
 }
 
 impl PublicSubkey {
-    pub fn new(key: packet::PublicSubkey, keyflags: KeyFlags) -> Self {
-        PublicSubkey { key, keyflags }
+    pub fn new(key: packet::PublicSubkey, keyflags: KeyFlags, embedded: Option<Signature>) -> Self {
+        PublicSubkey {
+            key,
+            keyflags,
+            embedded,
+        }
     }
 
     /// Produce a Subkey Binding Signature (Type ID 0x18), to bind this subkey to a primary key
@@ -112,6 +121,7 @@ impl PublicSubkey {
             primary_pub_key,
             key_pw,
             self.keyflags,
+            self.embedded,
         )?];
 
         Ok(SignedPublicSubKey { key, signatures })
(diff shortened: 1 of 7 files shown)


hko-s:
I think this implements the right semantics now.

Testing with rsop (and gosop) also looks good.

hko-s:
I removed some signing functions, which I initially used and got very confused by:

There were a set of `sign` functions that use subkey-binding or primary-binding signature type, but actually do the hashing for a totally different signature type ... so they produced broken signatures.

Once I removed those, an additional internal helper fn could also be removed because it was only used in the inconsistent signing functions.

(As an aside: I think this cleanup resolves some of my confusion in https://github.com/rpgp/rpgp/pull/450/)

src/composed/key/public.rs
```diff
@@ -27,6 +27,11 @@ pub struct PublicKey {
 pub struct PublicSubkey {
     pub key: packet::PublicSubkey,
     pub keyflags: KeyFlags,
+
+    /// Embedded signature, required for signing-capable subkeys.
+    ///
+    /// See <https://openpgp.dev/book/signing_components.html#special-case-binding-signing-subkeys>
```
dignifiedquire, line 33: can we use rfc links as main ref please?

src/packet/key/public.rs
```diff
@@ -161,19 +149,6 @@ impl PublicSubkey {
         })
     }
 
-    pub fn sign<R: CryptoRng + Rng, K>(
```
dignifiedquire, line 164: why is this removed?

src/composed/key/public.rs
```diff
@@ -27,6 +27,11 @@ pub struct PublicKey {
...
+    pub embedded: Option<SubpacketData>,
```
dignifiedquire, line 34: can we make this more type safe?

src/composed/key/public.rs
hko-s: right, good point :)

src/composed/key/public.rs
hko-s: I'd like that, but didn't have a good idea for how. I guess we could store a signature packet instead?

src/packet/key/public.rs
dignifiedquire: that's fair, if they produce invalid sigs, we should remove them

dignifiedquire:
merge conflicts

hko-s:
> merge conflicts

should be sorted

src/packet/key/secret.rs
```diff
@@ -584,6 +624,9 @@ fn create_signature(
     }
 }
 
+/// Make a direct key or a revocation signature
+/// TODO: expose this functionality in the public API
+#[allow(dead_code)]
```
dignifiedquire, line 629: same here?

src/packet/signature/config.rs
```diff
@@ -303,6 +303,42 @@ impl SignatureConfig {
...
+    }
+
+    /// Sign a key binding.
+    fn sign_key_binding_internal<K, P, L>(
+        self,
+        signing_key: &K,
+        signing_pub_key: &P,
+        key_pw: &Password,
+        key: &L,
+        backsig: bool,
```
dignifiedquire, line 340: not a fan of this bool, maybe having two different functions makes more sense

dignifiedquire: or an enum or sth like that

tests/key_test.rs
```diff
@@ -154,10 +154,10 @@ parse_dumps!(
     (
         test_parse_dumps_0,
         0,
-        17_727,
+        17_724,
```
dignifiedquire, line 157: have you confirmed in some form that these are actually invalid backsigs?

src/packet/signature/config.rs
hko-s: How do you feel about fixing the naming scheme, while we're touching this?

I'd now prefer to align more closely with the [signature type names](https://www.rfc-editor.org/rfc/rfc9580.html#name-signature-types):

- `verify_key_binding` -> `verify_subkey_binding`
- `verify_backwards_key_binding` -> `verify_primary_key_binding`

(and the same for the signing functions)

dignifiedquire (approved):

PR #556, opened 2025-05-17
impl `KeyDetails` for SignedSecretKey

baloo:
This restore the implementation that got removed in 46d331a3578491616e171c42a0f11b73b55c1350

This is useful when implementing abstractions on private keys like an HSM-backed signer (https://github.com/rpgp/rpgp/pull/554)

### DIFF
--- a/src/composed/signed_key/secret.rs
+++ b/src/composed/signed_key/secret.rs
@@ -12,11 +12,14 @@ use crate::{
         signed_key::{SignedKeyDetails, SignedPublicSubKey},
         ArmorOptions, PlainSessionKey, SignedPublicKey,
     },
-    crypto::hash::KnownDigest,
+    crypto::{hash::KnownDigest, public_key::PublicKeyAlgorithm},
     errors::{ensure, Result},
     packet::{self, Packet, PacketTrait, SignatureType},
     ser::Serialize,
-    types::{EskType, Imprint, Password, PkeskBytes, PublicKeyTrait, Tag},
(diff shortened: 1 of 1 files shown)


baloo:
@hko-s if you're around, could you run the tests on this one for me?

hko-s:
> @hko-s if you're around, could you run the tests on this one for me?

Found the button. Running now :)

dignifiedquire:
I removed this on purpose, can you link the code that uses this?

baloo:
Actually I've got confused in my error report I guess.

We're using that for RPM signing with rpm-rs:
https://github.com/rpm-rs/rpm/blob/517e20c6632d6556504357e8868e6a78bb3d86d6/src/rpm/signature/pgp.rs#L100

There is a bound on the Signer::new for the inner to implement `PublicKeyTrait` because it needs it to determine the algorithm it's going to sign RPMs with.

dignifiedquire:
> There is a bound on the Signer::new for the inner to implement `PublicKeyTrait` because it needs it to determine the algorithm it's going to sign RPMs with.

looking at this code, it should now use the new trait [`KeyDetails`](https://docs.rs/pgp/0.16.0-alpha.1/pgp/types/trait.KeyDetails.html) which is already implemented for the relevant things

baloo:
I've made the upgrade to rpgp 0.16 over in https://github.com/rpm-rs/rpm/pull/271 but I think I still need

PR #557, opened 2025-05-21
fix: make KeyDetails::new public again

hko-s:
This fixes an unintended API regression from #539


baloo (approved):
That fixes my issue, thanks!

dignifiedquire (approved):

PR #569, opened 2025-05-30
feat: convenience converter from UserId to Result<String>

hko-s:
Not sure if this is the best way to do this, but would be nice if applications had a more convenient way to interact with UserIds that contain valid Utf8


hko-s:
Closing in favor of #570

PR #572, opened 2025-06-06
wip: feat: decrypt GnuPG-specific packet type 20 encryted data and v5 SKESK

hko-s:
This PR implements decryption support for two GnuPG-specific formats, which are described in https://www.ietf.org/archive/id/draft-koch-librepgp-03.html

- GnuPG's "OCB Encrypted Data Packet" (a precursor to v2 SEIPD), and


hko-s:
closing in favor of #580

PR #573, opened 2025-06-10
Adopt RustCrypto new releases

baloo:
This is probably a bit early, but this PR bumps the dependencies to use @RustCrypto next set of releases.

Missing releases:
 - [ ] argon2 0.6
 - [x] p521 0.14
 - [x] k256 0.14
 - [ ] rsa 0.10
 - [ ] eax 0.6
 - [ ] ocb3 0.2
 - [ ] ed448-goldilocks 0.14
 - [ ] x448 0.14
 - [ ] slh-dsa 0.2

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -128,7 +128,7 @@ testresult = "0.4.1"
 chacha20 = { version = "0.10", features = ["rng"] }
 p256 = { version = "0.14" }
 pkcs8 = { version = "0.11", default-features = false, features = ["std", "pem"] }
-rsa = { version = "0.10.0-rc.18", default-features = false, features = ["encoding"] }
+rsa = { version = "0.10.0-rc.18", default-features = false, features = ["encoding", "hazmat"] }
 
 [features]
 default = ["bzip2"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,73 +42,66 @@ flate2 = { version = "1.1.1", default-features = false, features = ["zlib-rs"] }
 bzip2 = { version = "0.6.0", optional = true }
 
 # Crypto
-aes = "0.8.4"
-aead = { version = "0.5", features = ["bytes"] }
-argon2 = { version = "0.5", features = ["zeroize", "std"] }
-block-padding = "0.3"
-blowfish = "0.9"
-camellia = "0.1"
-cast5 = "0.11"
-cfb-mode = "0.8.2"
-cipher = "0.4.2"
-const-oid = "0.9.6"
+aes = "0.9"
+aead = { version = "0.6", features = ["bytes"] }
+argon2 = { version = "0.6", features = ["alloc", "zeroize"] }
+block-padding = "0.4.0"
+blowfish = "0.10"
+camellia = "0.2"
+cast5 = "0.12"
+cfb-mode = "0.9"
+cipher = "0.5"
+const-oid = "0.10"
 crc24 = "0.1.6"
-curve25519-dalek = { version = "4.1.3", default-features = false, features = [
+crypto-bigint = { version = "0.7", default-features = false, features = ["alloc"] }
+curve25519-dalek = { version = "5", default-features = false, features = [
     "alloc",
     "precomputed-tables",
     "zeroize",
 ] }
-des = "0.8"
-digest = "0.10.7"
-dsa = "0.6.3"
-ecdsa = "0.16.9"
-ed25519-dalek = { version = "2.1.1", default-features = false, features = ["std", "zeroize", "fast", "rand_core"] }
-elliptic-curve = "0.13"
-generic-array = "0.14.6"
-idea = "0.5"
-md-5 = { version = "0.10.5", features = ["oid"] }
+des = "0.9"
+digest = "0.11"
+dsa = { version = "0.7", features = [ "hazmat" ] }
+ecdsa = "0.17"
+ed25519-dalek = { version = "3", default-features = false, features = ["alloc", "fast", "rand_core", "zeroize"] }
+elliptic-curve = "0.14.1"
+hybrid-array = "0.4.13"
+idea = "0.6.0"
+md-5 = { version = "0.11", features = ["oid"] }
 num_enum = { version = ">=0.5.7, < 0.8", default-features = false }
-num-traits = "0.2.19"
-p256 = { version = "0.13", features = ["ecdsa", "ecdh"] }
-p384 = { version = "0.13", features = ["ecdsa"] }
-p521 = { version = "0.13", features = ["ecdsa", "ecdh"] }
-k256 = { version = "0.13", features = ["ecdsa"] }
-rand = "0.8.6"
-ripemd = { version = "0.1.3", features = ["oid"] }
-rsa = { version = "0.9.10" }
-sha1 = { version = "0.10.6", features = ["oid"] }
-sha1-checked = { version = "0.10", features = ["zeroize"] }
-sha2 = { version = "0.10.6", features = ["oid"] }
-sha3 = { version = "0.10.8", features = ["oid"] }
-signature = "2.2"
+p256 = { version = "0.14", features = ["ecdsa", "ecdh"] }
+p384 = { version = "0.14", features = ["ecdsa"] }
+p521 = { version = "0.14", features = ["ecdsa", "ecdh"] }
+k256 = { version = "0.14", features = ["ecdsa"] }
+rand = "0.10"
+ripemd = { version = "0.2", features = ["oid"] }
+rsa = { version = "0.10.0-rc.18" }
+sha1 = { version = "0.11", features = ["oid"] }
+sha1-checked = { version = "0.11.0-rc.0", features = ["zeroize"] }
+sha2 = { version = "0.11", features = ["oid"] }
+sha3 = { version = "0.12", features = ["oid"] }
+signature = "3"
 subtle = "2.6.1"
-twofish = "0.7"
+twofish = "0.8"
 zeroize = { version = "1.8", features = ["zeroize_derive"] }
-getrandom = { version = "0.2", optional = true }
-hkdf = "0.12.4"
-aes-gcm = "0.10.3"
-eax = "0.5.0"
-ocb3 = "0.1"
-aes-kw = { version = "0.2.1", features = ["std"] }
-cx448 = { version = "0.1.1", features = ["zeroize"] }
-num-bigint = { version = "0.8.6", features = [
-    "rand",
-    "i128",
-    "u64_digit",
-    "prime",
-    "zeroize",
-], package = "num-bigint-dig" }
-x25519-dalek = { version = "2.0.1", default-features = false, features = [
-    "alloc",
+getrandom = { version = "0.4", optional = true }
+hkdf = "0.13"
+aes-gcm = "0.11"
+eax = "0.6"
+ocb3 = "0.2"
+aes-kw = { version = "0.3" }
+ed448-goldilocks = { version = "0.14.0-pre.15" }
+x448 = { version = "0.14.0-pre.12", features = ["static_secrets"] }
+x25519-dalek = { version = "3", default-features = false, features = [
     "precomputed-tables",
     "zeroize",
     "static_secrets",
 ] }
 
 # PQC
-ml-kem = { version = "0.2.1", features = ["zeroize", "deterministic"], optional = true }
-ml-dsa = { version = "0.0.4", features = ["zeroize"], optional = true }
-slh-dsa = { version = "0.0.3", optional = true }
+ml-kem = { version = "0.3", features = ["zeroize", "hazmat"], optional = true }
+ml-dsa = { version = "0.1", features = ["zeroize"], optional = true }
+slh-dsa = { version = "0.2.0-rc.5", optional = true }
 replace_with = "0.1.8"
 
 # WASM helpers
@@ -117,11 +110,10 @@ web-time = { version = "1.1", optional = true }
 
 [dev-dependencies]
 glob = "0.3"
-hex-literal = "0.4"
+hex-literal = "1"
 pretty_assertions = "1"
 pretty_env_logger = "0.5"
-rand_chacha = "0.3"
-rand_xorshift = "0.3"
+rand_xorshift = "0.5.0"
 rayon = "1.10"
 serde = { version = "1.0.217", features = ["derive"] }
 serde_json = "1.0.47"
@@ -132,14 +124,14 @@ proptest-derive = "0.8.0"
 escape_string = "0.1.2"
 regex = "1.12"
 testresult = "0.4.1"
+chacha20 = { version = "0.10", features = ["rng"] }
 
 [features]
 default = ["bzip2"]
 
 # See README.md for features documentation
 bzip2 = ["dep:bzip2"]
-asm = ["sha2/asm", "md-5/asm"]
-wasm = ["getrandom", "getrandom/js", "web-time"]
+wasm = ["getrandom", "web-time"]
 
 large-rsa = []
 malformed-artifact-compat = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,8 +43,8 @@ bzip2 = { version = "0.6.0", optional = true }
 
 # Crypto
 aes = "0.9"
-aead = { version = "0.6", features = ["bytes"] }
-argon2 = { version = "0.6", features = ["alloc", "zeroize"] }
+aead = { version = "0.6", default-features = false, features = ["bytes", "rand_core"] }
+argon2 = { version = "0.6", default-features = false, features = ["alloc", "zeroize"] }
 block-padding = "0.4.0"
 blowfish = "0.10"
 camellia = "0.2"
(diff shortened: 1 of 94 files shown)


src/types/mpi.rs
```diff
@@ -115,21 +116,96 @@ impl Serialize for Mpi {
...
 }
 
-impl From<BigUint> for Mpi {
-    fn from(other: BigUint) -> Self {
-        Mpi(other.to_bytes_be().into())
+impl From<BoxedUint> for Mpi {
+    fn from(other: BoxedUint) -> Self {
+        // TODO(baloo): are Mpi used for private keys?
+        //              This is only meant to be used for public keys
+        Mpi(other.to_be_bytes_trimmed_vartime().into())
```
baloo: One question that might require some larger rewrite?

Or maybe have Mpi be a newtype for BoxedUint?

dignifiedquire: We can't use `BoxedUint` internally because we need to ensure storage is bit exactly the way we need it. 
MPIs are in some cases unfortunately used for secret keys, yes..

src/composed/key/builder.rs
```diff
@@ -1756,7 +1727,7 @@ mod tests {
         }
     }
     #[cfg(feature = "draft-pqc")]
-    fn gen_ed448_ml_kem_x448<R: Rng + CryptoRng>(mut rng: R, version: KeyVersion) {
+    fn gen_ed448_ml_kem_x448<R: RngCore + CryptoRng + ?Sized>(rng: &mut R, version: KeyVersion) {
```
dignifiedquire: do we have to switch to `&mut` ? I am really not a big fan of forcing a reference

dignifiedquire:
the overall changes are a lot simpler than I expected, which is nice

src/composed/key/builder.rs
baloo: we don't have to. I think I can roll those back.

Just made the change by habit/consistency with rustcrypto.

src/types/mpi.rs
baloo: That's only the case for serialization of private keys right?

baloo:
I need to track down the regression about compatibility. From my understanding the verifier is tolerant to LF/CR variations.

hko-s:
> I need to track down the regression about compatibility. From my understanding the verifier is tolerant to LF/CR variations.

That is a surprising regression.

LF/CR tolerance is a subtle and somewhat annoying aspect of OpenPGP - it interacts both with the type of the literal packet, and with the SignatureType. And cleartext signatures add a few more wrinkles.

However, I would not expect any of this OpenPGP-subtlety to interact with rust crypto upgrades O_o

baloo:
Yeah, I definitely did not expect it either. But I haven't taken the time to root-cause it yet (not familiar enough with the code base, it's intimidating^^)

baloo:
Turns out I messed up the `Mpi::from(BoxedUint)`:

<details>

```
diff --git a/src/types/mpi.rs b/src/types/mpi.rs
index 8a412b2d71..1f3e898125 100644
--- a/src/types/mpi.rs
+++ b/src/types/mpi.rs
@@ -118,27 +118,27 @@

 impl From<BoxedUint> for Mpi {
     fn from(other: BoxedUint) -> Self {
+        Self::from(&other)
+    }
+}
+
+impl From<&BoxedUint> for Mpi {
+    fn from(other: &BoxedUint) -> Self {
         // TODO(baloo): are Mpi used for private keys?
         //              This is only meant to be used for public keys
         Mpi(other.to_be_bytes_trimmed_vartime().into())
     }
 }

-impl From<&BoxedUint> for Mpi {
-    fn from(other: &BoxedUint) -> Self {
-        Mpi(other.to_be_bytes().into())
-    }
-}
-
 impl From<NonZero<BoxedUint>> for Mpi {
     fn from(other: NonZero<BoxedUint>) -> Self {
-        Mpi(other.to_be_bytes().into())
+        Self::from(other.get())
     }
 }

 impl From<&NonZero<BoxedUint>> for Mpi {
     fn from(other: &NonZero<BoxedUint>) -> Self {
-        Mpi(other.to_be_bytes().into())
+        Self::from(other.as_ref())
     }
 }
```

</details>

Some of the implementation didn't use the trimmed version which broke the fingerprinting and then broke the match.

hko-s:
Awesome that you found it! (After 2h of debugging though? That sounds a bit brutal.)

The MPI format is a horrifying armada of footguns, but ... it is there, in the "old" formats.

(One of many nice things about the new OpenPGP RFC 9580 is that it uses plain fixed size encoding for modern types of key material and signatures.)

baloo:
> Awesome that you found it! (After 2h of debugging though? That sounds a bit brutal.)

nah, those 2 hours included giving dinner to the kids ^^

baloo:
I'm not sure I know how to debug the remaining failures on the parser. Who can I bug about it?

hko-s:
> I'm not sure I know how to debug the remaining failures on the parser. Who can I bug about it?

I can try to help sort through the CI failures. Not with super-low latency, but happy to do it at a modest pace :)

hko-s:
> I can try to help sort through the CI failures. Not with super-low latency, but happy to do it at a modest pace :)

Looking at the logs, I see:

```
 --- STDERR:              pgp packet::many::tests::test_many_parser ---

thread 'packet::many::tests::test_many_parser' panicked at src/packet/many.rs:271:13:
assertion failed: `(left == right)`: mismatch in packet "./tests/tests/sks-dump/0000.pgp" (# off=1344258 ctb=99 tag=6 hlen=3 plen=321)

Diff < left / right > :
<PublicKey
>UserId
```

Which is surprising - it seems like the parser is not finding the same series of packets in the `sks-dump/0000.pgp` public key dump on this branch as it does on main. Not a type of problem I would have anticipated :grimacing:

baloo:
Yeah, and the payload being a gigantic blob doesn't really help. My best guess is that I need to get the offset of the offending packet and compare against a reference implementation?

hko-s:
> Yeah, and the payload being a gigantic blob doesn't really help. My best guess is that I need to get the offset of the offending packet and compare against a reference implementation?

This is essentially a stress-test for the parser, on real-world data (public key dumps from the early days of rpgp, in their native format).

If you're not in a rush, I'd suggest you hand this question off to me, and I try to poke at it later this week.

baloo:
I'm not in a rush, I got end to end tests of our signature workflow using this branch. It doesn't seem to trigger for us.

dignifiedquire:
That failure means that some parsing logic has changed. One trick is to run the test with logging on the branch before and after and look at the diff, where it starts to change

baloo:
> That failure means that some parsing logic has changed. One trick is to run the test with logging on the branch before and after and look at the diff, where it starts to change

I've been able to get the logs with:
```
RUST_LOG=debug cargo test -- --ignored  test_many_parser
```

I haven't got the time to compare just yet, but that's a good hint, thank you!

src/types/mpi.rs
```diff
@@ -115,21 +116,96 @@ impl Serialize for Mpi {
...
+        Self::from(&other)
     }
 }
 
-impl From<&BigUint> for Mpi {
-    fn from(other: &BigUint) -> Self {
-        Mpi(other.to_bytes_be().into())
+impl From<&BoxedUint> for Mpi {
+    fn from(other: &BoxedUint) -> Self {
+        // TODO(baloo): are Mpi used for private keys?
```
hko-s, line 127: Yes, `Mpi` are also used for (some) private key material

baloo: That's probably not as bad as I initially thought. This should only be used when serializing / deserializing the private key, but crypto operations should be fine.

hko-s:
> Looking at the logs, I see:
> 
> ```
>  --- STDERR:              pgp packet::many::tests::test_many_parser ---
> 
> thread 'packet::many::tests::test_many_parser' panicked at src/packet/many.rs:271:13:
> assertion failed: `(left == right)`: mismatch in packet "./tests/tests/sks-dump/0000.pgp" (# off=1344258 ctb=99 tag=6 hlen=3 plen=321)
> 
> Diff < left / right > :
> <PublicKey
> >UserId
> ```
> 
> Which is surprising - it seems like the parser is not finding the same series of packets in the `sks-dump/0000.pgp` public key dump on this branch as it does on main. Not a type of problem I would have anticipated 😬

I've debugged this. The parser is now skipping a public key packet that it could previously parse.
It's a DSA public key with p/q bit sizes of 768/160.

This packet now gets an error from RustCrypto's DSA implementation, here:

https://github.com/RustCrypto/signatures/commit/8f770af1a29a390b63c4d92250e8553b192cd1ec#diff-417f24175761f2ad4f12633347b0a7c1dd9beb8f70bb9908d19ee3405a340d98R48

courtesy of `baloo authored on Apr 23` :grinning: 

(In rPGP, this error happens in `pgp::types::params::public::dsa::DsaPublicParams::try_from_mpi`)

I have no strong feelings about how this should be resolved. I see two possible approaches:

- RustCrypto could allow some more lengths, including this weird 768/160 size
- rPGP can change its expectation in this test, and just consider this packet malformed

If we go for the latter, I can adjust the test.

---

As a tangent: I notice that RustCrypto also doesn't support DSA keys with size 4096/256. I suspect such keys have been used in OpenPGP - but don't think it's crucial to support them.
(Added context: I randomly noticed just yesterday that RNP - which is used in Thunderbird - added support for DSA 4096 in their latest release: https://github.com/rnpgp/rnp/releases/tag/v0.18.0 in June.)

baloo:
Ha. I also spent some time on it yesterday night. I should have posted my findings! (Turns out they were the same!) 

There are a couple of keys that fail to parse, but I haven’t found the correct to skip them in the parser.
I could skip the first one, which was a sub key iirc, but the second one gave me trouble.
I’d appreciate your input on how to do that.

hko-s:
> Ha. I also spent some time on it yesterday night. I should have posted my findings! (Turns out they were the same!)
> 
> There are a couple of keys that fail to parse, but I haven’t found the correct to skip them in the parser. I could skip the first one, which was a sub key iirc, but the second one gave me trouble. I’d appreciate your input on how to do that.

I made a commit that skips all of the now unsupported DSA keys in

https://github.com/hko-s/rpgp/commits/baloo/rustcrypto-new-releases/

It's really a lot of (historical, and obviously not secure anymore) keys that are affected.

The large majority of them are 768/160 bit sized. Most of the other now bad keys are 512 bit, and at least one had an entirely odd length (912/160).

hko-s:
> The large majority of them are 768/160 bit sized. Most of the other now bad keys are 512 bit, and at least one had an entirely odd length (912/160).

Just for the record:

The test key set `sks-dump/0000.pgp` contains ~20k public keys. Of those, 49 DSA-based ones can't be handled anymore in this PR. So we're currently "losing" support for around 0.25% of that (many years old) set of public keys.

Not the end of the world (and of course those keys should definitely not be considered secure by today's standard).
Still, it's a not entirely trivial percentage of (historical and now unsafe) keys that we can't handle anymore right now.

dignifiedquire:
I think we should be able to parse them, so I think we need to add a dsa key variant of unsupported or similar if the dsa crate is that restrictive

baloo:
Latest commit makes use of `dsa::Components::from_components_unchecked` (https://github.com/RustCrypto/signatures/pull/1059) and should fix support for weird sized keys.

baloo:
At last, the tests are fixed. I got to ignore a couple parse errors in the last commit, let me know if you're concerned about that.

dignifiedquire:
> At last, the tests are fixed. I got to ignore a couple parse errors in the last commit, let me know if you're concerned about that.

great news, yeah those parsing tests variations should be fine

src/types/params/public/rsa.rs
```diff
@@ -66,7 +66,7 @@ mod tests {
     prop_compose! {
         pub fn rsa_pub_gen()(seed: u64) -> rsa::RsaPublicKey {
             let mut rng = rand_chacha::ChaCha8Rng::seed_from_u64(seed);
-            rsa::RsaPrivateKey::new(&mut rng, 512).unwrap().to_public_key()
```
baloo, line 70: There is an hazmat flag to allow the use of <1024 rsa keys, We can use that if needed, but I think the 1024 is acceptable.

hko-s: I guess that's small for speed. We can see if this change makes a noticeable difference in CI runtime? (I'd assume not)

baloo: actually, there is a `new_unchecked` function (that doesn't have the restriction) under `hazmat` which I can enable for tests.

overhacked:
@baloo, I see you've been pushing to this recently. I'm working on getting `ssh-agent-lib` ready for RustCrypto upgrades. The [`pgp` example][pgp-wrapper.rs] over there needs a lot of refactoring; before I start working on that in earnest, using this PR as the "new version", can I ask 1) are you already working on anything over at `ssh-agent-lib`, bringing it up to date for new RustCrypto, that isn't pushed to the public repo and 2) are you actively rebasing this PR so any work I did over at `ssh-agent-lib` that depends on this would be up-to-date when (🤞) this gets merged?

Also, I noticed when starting to play with getting [`ssh-agent-lib`'s PGP example][pgp-wrapper.rs] up to date that the `pgp::packet::PublicKey` constructor introduced back in #142 seems to have been removed some time after `pgp-0.15.0`. Being able to construct a `PublicKey` from the underlying RSA/etc. components is the key part of [pgp-wrapper.rs]. If I start bringing the example up to date, what are your thoughts on me either opening a rpgp/rpgp PR that is stacked on this one or re-introducting the #142 constructor in this PR?

[pgp-wrapper.rs]: https://github.com/wiktor-k/ssh-agent-lib/blob/main/examples/pgp-wrapper.rs

hko-s:
Hey @overhacked! A quick reaction to one narrow point in your message:

> Also, I noticed [..] that the `pgp::packet::PublicKey` constructor introduced back in #142 seems to have been removed [..]

That functionality should still be available, via:
https://docs.rs/pgp/latest/pgp/packet/struct.PublicKey.html#method.from_inner

The data structures have changed around a bit, but `PubKeyInner` should be usable just like `PublicKey` was, before. However, it seems that the type is not currently visible in the public API - at a quick glance, I don't understand why :thinking: .. either way, I think the way forward is to make it public.

overhacked:
> However, it seems that the type is not currently visible in the public API - at a quick glance, I don't understand why 🤔 .. either way, I think the way forward is to make it public.

IMO, having `packet::PublicKey::from_inner` exposed and needing to access `PubKeyInner` leads to a more difficult to understand API. I'd much rather `PubKeyInner` be invisible to the public API and a method on `packet::PublicKey` to construct it from key primitives.

baloo:
> @baloo, I see you've been pushing to this recently. 

Yeah, fairly consistently over the last 6-8 months ^^

Mostly because I need it for work, but also rpgp is using pretty all of rustcrypto, and it's been extremely useful to find bugs in the upcoming release.

> I'm working on getting `ssh-agent-lib` ready for RustCrypto upgrades. The [`pgp` example](https://github.com/wiktor-k/ssh-agent-lib/blob/main/examples/pgp-wrapper.rs) over there needs a lot of refactoring; before I start working on that in earnest, using this PR as the "new version", can I ask 1) are you already working on anything over at `ssh-agent-lib`, bringing it up to date for new RustCrypto, that isn't pushed to the public repo

No, I'm not using ssh-agent-lib at the moment. Mostly as a lack of spare time, not lack of interest.

> and 2) are you actively rebasing this PR so any work I did over at `ssh-agent-lib` that depends on this would be up-to-date when (🤞) this gets merged?

Yeah, I rebase this PR every week or so.

> 
> Also, I noticed when starting to play with getting [`ssh-agent-lib`'s PGP example](https://github.com/wiktor-k/ssh-agent-lib/blob/main/examples/pgp-wrapper.rs) up to date that the `pgp::packet::PublicKey` constructor introduced back in #142 seems to have been removed some time after `pgp-0.15.0`. Being able to construct a `PublicKey` from the underlying RSA/etc. components is the key part of [pgp-wrapper.rs](https://github.com/wiktor-k/ssh-agent-lib/blob/main/examples/pgp-wrapper.rs). If I start bringing the example up to date, what are your thoughts on me either opening a rpgp/rpgp PR that is stacked on this one or re-introducting the #142 constructor in this PR?

I'd rather not have to maintain your changes. What I would recommend would be to make those changes against main, then make a merge between those changes and this branch.

Those commits here are all made with jujutsu, and if you were to `jj bookmark track baloo/rustcrypto-new-releases@baloo`, your merge commit would get automatically rebased by jj, that way you can iterate with the authors of rpgp without having me in the loop.

overhacked:
Thanks for the reply! This is an awesome reason for me to learn `jj`, which I'd been meaning to for a while. Also thanks for the clear answer about not wanting to maintain the additional patch in this PR; makes it easy to decide which way to move forward.

> No, I'm not using ssh-agent-lib at the moment. Mostly as a lack of spare time, not lack of interest.

Since I've already got a working patch for upcoming RustCrypto 0.9+ for `ssh-agent-lib` except the PGP example, I'll go ahead and start a draft PR over there.

baloo:
> Thanks for the reply! This is an awesome reason for me to learn `jj`, which I'd been meaning to for a while. Also thanks for the clear answer about not wanting to maintain the additional patch in this PR; makes it easy to decide which way to move forward.

I mean, it's more that it's going to be a frustration for the both of us having to wait for each other to sync up, and I'd rather avoid that if possible.

> > No, I'm not using ssh-agent-lib at the moment. Mostly as a lack of spare time, not lack of interest.
> 
> Since I've already got a working patch for upcoming RustCrypto 0.9+ for `ssh-agent-lib` except the PGP example, I'll go ahead and start a draft PR over there.

Let's follow up there (I also have one, and rand_core changed its API, again)

src/crypto/rsa.rs
```diff
@@ -248,7 +248,7 @@ mod tests {
     prop_compose! {
         pub fn key_gen()(seed: u64) -> RsaPrivateKey {
             let mut rng = chacha20::ChaCha8Rng::seed_from_u64(seed);
-            RsaPrivateKey::new(&mut rng, 512).unwrap()
+            RsaPrivateKey::new_unchecked(&mut rng, 512).unwrap()
```
hko-s, line 251: Maybe we should bump this test to using 1024 bit keys, and lose the `hazmat` feature in the `rsa dependency`?

baloo: This costs an exponential amount of time. Which I'm not sure we should do in an propcheck.

Now: that tests an`rsa` API. I'm not sure I see the point of that test in the first place.

hko-s: Right. I'm not sure what is appropriate to do here. Just wondered if  we can avoid taking on the `hazmat` feature just for this test.

Maybe @dignifiedquire can weigh in sometime.

dignifiedquire: hazmat is fine to have for tests, that's why it exists

hko-s: Ahh, the feature is only taken in `dev-dependencies`. Nevermind me. I missed that.

wiktor-k:
Happy to see more and more stuff getting out of `rc` status into stable! :face_holding_back_tears: 

Thanks for your hard work @baloo :bow:

hko-s:
Happy birthday to this PR! Thanks, @baloo for your ongoing work on it.

(Here's hoping that the PR will be merged well before its second birthday :sweat_smile:)

Cargo.toml
```diff
@@ -132,14 +125,17 @@ proptest-derive = "0.8.0"
...
+p256 = { version = "0.14.0-rc.15" }
+pkcs8 = { version = "0.11", default-features = false, features = ["std", "pem"] }
+rsa = { version = "0.10.0-rc.18", default-features = false, features = ["encoding", "hazmat"] }
 
 [features]
 default = ["bzip2"]
 
 # See README.md for features documentation
 bzip2 = ["dep:bzip2"]
-asm = ["sha2/asm", "md-5/asm"]
```
baloo, line 141: @hko-s @link2xt heads up

There is no longer an asm feature in those crates. (And since we've dropped the direct dependency on `sha1-asm` in #814 ... the feature doesn't really make sense anymore).

PR #579, opened 2025-06-20
Performance improvements

hko-s:
This PR is a variant of the branch https://github.com/rpgp/rpgp/commits/perf-wip/, but without the GnuPG-AEAD functionality that is also present in `perf-wip`.

### DIFF
--- a/src/composed/message/reader/signed.rs
+++ b/src/composed/message/reader/signed.rs
@@ -11,6 +11,8 @@ use crate::{
     util::{fill_buffer_bytes, NormalizingHasher},
 };
 
+const BUFFER_SIZE: usize = 8 * 1024;
+
 #[derive(derive_more::Debug)]
 pub enum SignatureBodyReader<'a> {
     Init {
@@ -132,8 +134,8 @@ impl<'a> SignatureBodyReader<'a> {
                     signature,
                 } => {
                     debug!("SignatureReader init");
-                    let mut buffer = BytesMut::with_capacity(1024);
-                    let read = fill_buffer_bytes(&mut source, &mut buffer, 1024)?;
+                    let mut buffer = BytesMut::with_capacity(BUFFER_SIZE);
+                    let read = fill_buffer_bytes(&mut source, &mut buffer, BUFFER_SIZE)?;
 
                     if read == 0 {
                         return Err(io::Error::new(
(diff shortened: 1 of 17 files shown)


tests/decrypt_speed.rs
```diff
@@ -0,0 +1,79 @@
...
+    let elapsed_milli = elapsed.as_millis();
+    let mb_per_s = 900f64 / elapsed_milli as f64 * 1000f64;
+    println!("Elapsed: {elapsed_milli} ms, MByte/s: {mb_per_s:.2?}");
+    println!("{:?}", msg.last());
+}
+
+// temporary test, requires local copy of a large encrypted message
+#[ignore]
+#[test]
+fn speed_seipdv1() {
```
dignifiedquire: lets make these an example, that way we can run them, but it doesn't fail our test suite

hko-s: good idea. will do.

hko-s:
Note: This PR currently breaks semver, e.g. by changing the type of `SymEncryptedProtectedDataReader`.

tests/decrypt_speed.rs
hko-s: moved the speed tests aside, and after some other minor cleanup, CI is now at least trying to do its thing again :sweat_smile:

src/composed/message/reader/signed.rs
```diff
@@ -132,9 +132,8 @@ impl<'a> SignatureBodyReader<'a> {
                     signature,
                 } => {
                     debug!("SignatureReader init");
-                    let mut buffer = BytesMut::zeroed(1024);
-                    let read = fill_buffer(&mut source, &mut buffer, None)?;
-                    buffer.truncate(read);
+                    let mut buffer = BytesMut::with_capacity(1024);
```
hko-s: Is there a reason why this reader doesn't use the `const BUFFER_SIZE: usize = 8 * 1024;` pattern?

src/composed/message/reader/signed_one_pass.rs
```diff
@@ -124,9 +124,8 @@ impl<'a> SignatureOnePassReader<'a> {
                     mut source,
                 } => {
                     debug!("SignatureOnePassReader init");
-                    let mut buffer = BytesMut::zeroed(1024);
-                    let read = fill_buffer(&mut source, &mut buffer, None)?;
-                    buffer.truncate(read);
+                    let mut buffer = BytesMut::with_capacity(1024);
```
hko-s: Is there a reason why this reader doesn't use the `const BUFFER_SIZE: usize = 8 * 1024;` pattern?

dignifiedquire:
> Wondering about aligning the signed message reader buffer sizes with the others (and maybe the compressed one, as well?)

not really, other than me being lazy..

hko-s:
> not really, other than me being lazy..

I've normalized the buffers to 8k, all

dignifiedquire (approved):

PR #580, opened 2025-06-20
wip: feat: decrypt GnuPG-proprietary packet type 20 encryted data and v5 SKESK

hko-s:
This PR implements decryption support for two GnuPG-proprietary formats, which are described in https://www.ietf.org/archive/id/draft-koch-librepgp-03.html

- GnuPG's "OCB Encrypted Data Packet" (a precursor to v2 SEIPD), and


hko-s:
I think we'll need to make two decisions about how to expose this functionality:

1. Do we want to feature-gate the entire thing, so that users of rPGP need to opt into it (a feature like `enable-gnupg-proprietary`)?
2. Do we expose `SKESK v5` and `OCB` (packet type 20) decryption as opt-in functionality in the message decryptor API? (similar to how decryption of the `SED` packet [not integrity-protected] needs to be opted into by a caller?)

---

My current tendency is "no to 1, yes to 2.":

For 1., I have mixed feelings about feature-gating, but lean towards "no". While feature-gating feels orderly, it also seems a bit pointless. I don't expect it would save much space in the compiled output to disable, but adds cognitive overhead to the codebase and makes testing more combinatorically annoying.

For 2., as an application developer I'd like to be able to configure the message decryptor between "OpenPGP RFC compliant" behavior and "compatible with GnuPG proprietary formats" at runtime (e.g. to allow `rsop` users to choose between these decryption modes with a commandline parameter).

PR #581, opened 2025-06-22
fix: make signature equality more robust

hko-s:
I ran into issues with equality between signatures. When changing pre-existing Signatures in memory (e.g. changing the contents of the unhashed area), the length in the packet header in memory goes out of sync.

This currently breaks comparison with Signatures that have exactly the same contents, but where the length in the packet header is not equally out of sync.

I'm not sure that I'm happy with the approach in this PR, but it does resolve the concrete issue I'm seeing.

### DIFF
--- a/src/packet/signature/types.rs
+++ b/src/packet/signature/types.rs
@@ -33,10 +33,42 @@ use crate::{
 
 /// Signature Packet
 /// <https://www.rfc-editor.org/rfc/rfc9580.html#name-signature-packet-type-id-2>
-#[derive(Clone, PartialEq, Eq, derive_more::Debug)]
+#[derive(Clone, Eq, derive_more::Debug)]
 pub struct Signature {
     packet_header: PacketHeader,
     pub inner: InnerSignature,
+}
+
+impl PartialEq for Signature {
+    /// Semantical equality on signature packets.
+    ///
+    /// This equality implementation checks that:
+    /// - the packet header version is the same, and
+    /// - the content of "inner" is the same.
+    ///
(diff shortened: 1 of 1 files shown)


hko-s:
Hm. Independently, GnuPG of course insists on producing old-style packet headers, so comparing packets against GnuPG will always be "bitwise", and show differences if the packet framing differs.

This is a bit of a rabbithole.

dignifiedquire:
I don't think this is good to do. If you want a more relaxed version we could add a "is_similar_to" method or sth like that, but equality should really check for equality

hko-s:
> I don't think this is good to do. If you want a more relaxed version we could add a "is_similar_to" method or sth like that, but equality should really check for equality

Agreed.

I'd want such a relaxed comparison operation to be able to work recursively, e.g. all the way through a SignedPublicKey's parts. Will think some more and maybe suggest something in a while.

hko-s:
Actually ... while I agree that it's good to have `PartialEq` implement bitwise equality, I'm still unsure about the case I outlined above.

When I modify a `Signature` by doing something like `config.unhashed_subpackets.insert(foo)`, and then compare the `Signature` with a serialized, then deserialized copy of itself, it won't match. Because the original `Signature` still contains the (then wrong) length in its packet header.

The issue I ran into is that even the bitwise comparison is just wrong after modifying the contents of the `SignatureConfig` in `inner`. This is a bit of a footgun, and I wonder if we can improve the rPGP API to have this be less of a footgun

hko-s:
Closing in favor of #582

PR #582, opened 2025-06-24
refactor: make InnerSignature private to avoid footguns

dignifiedquire:
ref #581

### DIFF
--- a/src/packet/signature/types.rs
+++ b/src/packet/signature/types.rs
@@ -246,7 +246,7 @@ impl Signature {
         if let InnerSignature::Known { ref config, .. } = self.inner {
             self.insert_unhashed_subpacket(config.unhashed_subpackets.len(), subpacket)
         } else {
-            unimplemented!("error")
+            bail!("Unknown Signature type, can't add Subpacket");
         }
     }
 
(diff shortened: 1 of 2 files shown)


hko-s:
This PR is definitely the way forward. It removes the footgun I stepped into, and the current state of this PR covers all use cases that I need.

Happy to make more adjustments if you have any concerns.

src/packet/signature/types.rs
```diff
@@ -253,6 +325,7 @@ impl Signature {
         self.config().map(|c| c.hash_alg)
     }
 
+    /// Returns the actual byte level  signature.
```
wiktor-k:
```suggestion
    /// Returns the actual byte level signature.
```

dignifiedquire:
lgtm

hko-s:
I've made the unhashed subpacket fns a little more defensive, yet. Happy to see this merged, whenever.

dignifiedquire:
Can you add some tests for this? I just realized this has no coverage at all

hko-s:
> Can you add some tests for this? I just realized this has no coverage at all

D'oh. Good point, of course. Yes!

hko-s:
I've added a test (and renamed the unhashed area access functions to be more visibly grouped)

PR #583, opened 2025-07-07
ci: tests on windows

hko-s:


link2xt:
This does not actually test `x86_64-pc-windows-gnu`. If you go into the log for `x86_64-pc-windows-gnu` run, it says it already has installed `msvc`:
image

Need to somehow overwrite the target, otherwise I think it still uses `msvc` target and `gnu` target is installed on the side just in case you want to use it for "cross-compilation".

"override" just sets the "toolchain" to be the default, but toolchain has multiple targets and "native" msvc is selected by default.

I created https://github.com/rpgp/rpgp/pull/585 setting `--target` explicitly, let's see.

hko-s:
replaced by #585

PR #585, opened 2025-07-08
ci: run Windows tests with the specified target triple

link2xt:
Otherwise host triple is used, which is "msvc", not the "gnu" one that is installed. "override" option only sets the default toolchain, but not the target.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,12 +75,8 @@ jobs:
       matrix:
         os: [windows-latest]
         rust:
-          - nightly-2025-04-02
-          - stable
-        target:
-          - x86_64-pc-windows-gnu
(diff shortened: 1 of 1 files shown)


hko-s:
Great, thanks for moving the windows CI/debugging hopefully towards more insights.

It looks to me like this run still has the same issue. I see

```
installed toolchains
--------------------
stable-x86_64-pc-windows-msvc (active, default)
```

in the output for the `-gnu` job :confused:

link2xt:
I think it's fine that default toolchain is different because I passed `--toolchain` option explicitly. But now I pushed another commit, now just using rustup directly instead of archived https://github.com/actions-rs/toolchain

dignifiedquire:
thanks

PR #586, opened 2025-07-09
feat: add EcdsaPublicParams::curve and is_supported

hko-s:
This aligns the EcdsaPublicParams API with EcdhPublicParams

### DIFF
--- a/src/types/params/public/ecdsa.rs
+++ b/src/types/params/public/ecdsa.rs
@@ -40,10 +40,15 @@ pub enum EcdsaPublicParams {
 }
 
 impl EcdsaPublicParams {
+    /// Is this key based on a curve that we know how to parse?
+    ///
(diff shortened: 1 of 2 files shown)


src/types/params/public/ecdsa.rs
```diff
@@ -40,6 +40,20 @@ pub enum EcdsaPublicParams {
 }
 
 impl EcdsaPublicParams {
+    pub fn is_supported(&self) -> bool {
+        !matches!(self, Self::Unsupported { .. })
+    }
+
+    pub fn curve(&self) -> ECCCurve {
```
dignifiedquire, line 52: docs please

hko-s: Good point. Added some docs for both `*PublicParams`

dignifiedquire (approved):

PR #592, opened 2025-07-12
fix: make user id self certifications as CertPositive

hko-s:


src/packet/user_attribute.rs
```diff
@@ -242,17 +248,29 @@ impl UserAttribute {
...
         signee: &K,
+        typ: SignatureType,
     ) -> Result<SignedUserAttribute>
     where
         R: CryptoRng + Rng,
         P: SecretKeyTrait,
         K: PublicKeyTrait + Serialize,
     {
+        ensure!(
+            [
```
dignifiedquire: maybe make this a constant?

src/composed/key/shared.rs
```diff
@@ -121,7 +121,7 @@ impl KeyDetails {
         let mut users = vec![];
 
         if let Some(primary_user_id) = self.primary_user_id {
-            let mut config = SignatureConfig::from_key(&mut rng, key, SignatureType::CertGeneric)?;
+            let mut config = SignatureConfig::from_key(&mut rng, key, SignatureType::CertPositive)?;
```
dignifiedquire, line 126: please document why this type is used

hko-s: It's one of those things that seemed to have emerged, but not been documented, historically. The RFC doesn't seem to give any guidance. However, all other implementations that I tried just now make this signature type for User ID self-signatures.

Plus, I added links to `draft-gallagher-openpgp-signatures` which is still in an early stage, but it at least mentions the connection between that signature type and self-signatures.

dignifiedquire (approved):

PR #595, opened 2025-07-13
feat: decrypt GnuPG-proprietary packet type 20 encryted data and v5 SKESK

hko-s:
This PR implements decryption support for two GnuPG-proprietary formats, which are described in https://www.ietf.org/archive/id/draft-koch-librepgp-03.html

- GnuPG's "OCB Encrypted Data Packet" (a precursor to the IETF-standardized v2 SEIPD), and

[this is a rebased reboot of #580]


dignifiedquire:
please update the `STATUS.md` doc as well

src/packet/gnupg_aead.rs
```diff
@@ -0,0 +1,105 @@
...
+            packet_header,
+            config,
+            data: data.freeze(),
+        })
+    }
+}
+
+impl Serialize for GnupgAeadData {
+    fn to_writer<W: Write>(&self, _: &mut W) -> Result<()> {
+        todo!()
```
dignifiedquire: this is not good, please implement these so we have roundtrip tests, even if we don't emit them anywhere

tests/libre.rs
```diff
@@ -0,0 +1,181 @@
+//! Tests for non-OpenPGP formats from
```
dignifiedquire: file should be renamed to sth like`gnupg.rs`

src/packet/gnupg_aead.rs
hko-s: I'm happy as long as there's no risk rPRG users will pollute the zone with these packets.

At first I was worried that users of rPGP might produce them if we implement `Serialize`. But on second thought `Serialize` is fine, as long as we don't have an implementation that encrypts to this format.

So, yes! I'll implement serialize.

src/composed/message/types.rs
```diff
@@ -576,16 +628,26 @@ impl<'a> Edata<'a> {
                 // SED packets are malleable, decrypting them should only be necessary for historical data
                 bail!("Decryption of SymEncryptedData is discouraged")
             }
+            Self::GnupgAeadData { .. } => {
+                bail!("GnuPG's AEAD format (packet type 20) is non-standard, and not supported by default")
```
dignifiedquire: maybe mention what API to use instead? same with the SED above?

src/composed/message/types.rs
```diff
@@ -576,16 +628,26 @@ impl<'a> Edata<'a> {
...
+    /// In addition, decryption of legacy SED packets and GnuPG AEAD (packet type 20) can be explicitly opted into.
     ///
-    /// HAZMAT: Decrypts SEIPD (v1 or v2) and SED packets.
-    pub fn decrypt_legacy(&mut self, key: &PlainSessionKey) -> Result<()> {
+    /// HAZMAT: Depending on the passed `enable` parameter, this decrypts malleable SED packets.
+    /// Also see <https://www.rfc-editor.org/rfc/rfc9580.html#name-avoiding-ciphertext-malleab>
+    pub fn decrypt_permissive(
+        &mut self,
+        key: &PlainSessionKey,
+        enable: &[OtherEdataType],
```
dignifiedquire: I think a custom builder would be nicer for this

```rust
let decrypt_opts = DecryptionOpts::default()
    .enable_legacy()
    .enable_gnpg_aead();
```

src/composed/message/types.rs
```diff
@@ -594,7 +656,18 @@ impl<'a> Edata<'a> {
                 reader.decrypt(key)?;
             }
             Self::SymEncryptedData { reader } => {
-                reader.decrypt(key)?;
+                if enable.contains(&OtherEdataType::Legacy) {
+                    reader.decrypt(key)?;
+                } else {
+                    bail!("Decryption of SymEncryptedData is not enabled")
```
dignifiedquire: maybe "disabled"?

src/composed/message/types.rs
```diff
@@ -1080,6 +1164,13 @@ pub struct TheRing<'a> {
     ///
     /// Defaults to `false`.
     pub allow_legacy: bool,
+    /// If this is `true` non-standard GnuPG AEAD packets (type 20) are also decrypted.
+    ///
+    /// This also enables decryption of the GnuPG-proprietary v5 SKESK format
+    /// (AEAD-based password encrypted session key).
+    ///
+    /// Defaults to `false`.
+    pub allow_gnupg_aead: bool,
```
dignifiedquire: I think if we have `DecryptionOpts` like I suggested above, we can replace these two boolean fields with that in here

hko-s: yes, that sounds good :+1:

src/packet/gnupg_aead.rs
```diff
@@ -0,0 +1,126 @@
...
+        sym::SymmetricKeyAlgorithm,
+    },
+    errors::{ensure_eq, InvalidInputSnafu, Result},
+    packet::{PacketHeader, PacketTrait},
+    parsing_reader::BufReadParsing,
+    ser::Serialize,
+    types::Tag,
+};
+
+#[derive(Clone, PartialEq, Eq, derive_more::Debug)]
```
dignifiedquire, line 21: docs

src/packet/gnupg_aead.rs
```diff
@@ -0,0 +1,126 @@
...
+        Ok(())
+    }
+
+    fn write_len(&self) -> usize {
+        4 + self.iv.len()
+    }
+}
+
+#[derive(Clone, PartialEq, Eq, derive_more::Debug)]
+pub struct GnupgAeadData {
```
dignifiedquire, line 92: docs

src/packet/gnupg_aead.rs
```diff
@@ -0,0 +1,126 @@
...
+        sum += self.data.len();
+        sum
+    }
+}
+
+impl PacketTrait for GnupgAeadData {
+    fn packet_header(&self) -> &PacketHeader {
+        &self.packet_header
+    }
+}
```
dignifiedquire, line 132: please add proptests for serialization and deserialization, like we have for the other packets

src/packet/gnupg_aead.rs
```diff
@@ -0,0 +1,126 @@
...
+            aead,
+            chunk_size,
+            iv: iv.into(),
+        })
+    }
+}
+
+impl Serialize for Config {
+    fn to_writer<W: Write>(&self, writer: &mut W) -> Result<()> {
+        writer.write_u8(0x01)?;
```
dignifiedquire: make this a constant

src/packet/gnupg_aead.rs
```diff
@@ -0,0 +1,126 @@
...
+    ser::Serialize,
+    types::Tag,
+};
+
+#[derive(Clone, PartialEq, Eq, derive_more::Debug)]
+pub struct Config {
+    pub sym_alg: SymmetricKeyAlgorithm,
+    pub aead: AeadAlgorithm,
+    pub chunk_size: ChunkSize,
+    pub iv: Vec<u8>,
```
dignifiedquire: this is fixed size, should be an array not a `Vec`

src/packet/sym_key_encrypted_session_key.rs
```diff
@@ -411,15 +406,31 @@ fn parse_v5<B: BufRead>(
...
     let sym_alg = i.read_u8().map(SymmetricKeyAlgorithm::from)?;
-    let aead = i.read_u8().map(AeadAlgorithm::from)?;
-    let s2k_len = i.read_u8()?;
-    let s2k_data = i.read_take(s2k_len.into());
-    let s2k = StringToKey::try_from_reader(s2k_data)?;
+
+    // A one-octet encryption mode number which MUST be 2 to indicate OCB.
+    let mode = i.read_u8()?;
+
+    // we only support OCB mode (it's unclear if 1 (EAX) was every produced)
```
dignifiedquire: typo "every"

src/packet/sym_key_encrypted_session_key.rs
```diff
@@ -411,15 +406,31 @@ fn parse_v5<B: BufRead>(
...
+    // A starting initialization vector of size specified by the mode.
     let iv = i.take_bytes(aead.iv_size())?;
+
+    // An authentication tag for the encryption mode.
     let aead_tag_size = aead.tag_size().unwrap_or_default();
-    let esk = i.rest()?;
+
+    // The encrypted session key itself, which is decrypted with the string-to-key object using
+    // the given cipher and encryption mode.
+    let esk = i.take_bytes(sym_alg.key_size() + aead_tag_size)?; // TODO: is this always the right length?
```
dignifiedquire: what about this TODO?

src/packet/gnupg_aead.rs
hko-s: It's not fixed between the two modes that [libre](https://www.ietf.org/archive/id/draft-koch-librepgp-03.html#name-ocb-encrypted-data-packet-t) specifies. The draft has both EAX and OCB, with differently sized IV. If I ever get a hold of a test vector for EAX, i might like to also implement that, for completeness.

src/packet/gnupg_aead.rs
hko-s: I'm failing to find serialize/deserialze proptests for SED or SEIPD. Where should I look? 😕

src/composed/message/types.rs
```diff
@@ -576,16 +628,26 @@ impl<'a> Edata<'a> {
                 // SED packets are malleable, decrypting them should only be necessary for historical data
                 bail!("Decryption of SymEncryptedData is discouraged")
```
hko-s: `decrypt` and `decrypt_permissive` currently emit different error messages. If the messages were unified, `decrypt` could be implemented in terms of `decrypt_permissive` with `DecryptionOptions` set to default.

src/packet/sym_key_encrypted_session_key.rs
hko-s: That was me being unsure what the draft is telling me. I now think the code is correct, and removed the TODO :sweat_smile:

src/packet/gnupg_aead.rs
hko-s: I generalized [the iv reading](https://github.com/rpgp/rpgp/pull/595/files#diff-76051256aa3d623990719cb9015595759ca3243733c6806ba2c5d488d37aa919R61) to clarify this.

dignifiedquire: then use `Bytes` please

hko-s: ah ok, i didn't think of that. changed the type.

dignifiedquire (approved):

PR #600, opened 2025-07-16
feat: Implement several standard library extension traits for `Fingerprint`

wiktor-k:
This PR implements `std::fmt::UpperHex` and `std::fmt::LowerHex` formatters. Additionally `AsRef<[u8]>` is implemented for better compatibility with other crates.


hko-s:
Thanks for the PR, I've missed such formatters in the past, this looks good to me!

Issue #575, opened 2025-06-14
dearmor: disable crc24 calculation by default


hko-s:
When dearmoring large data, the CRC24 calculation takes >33% of CPU time.

However, RFC 9580 mandates that the checksum should always be ignored:

> An implementation MUST NOT reject an OpenPGP object when the CRC24 footer is present, missing, malformed, or disagrees with the computed CRC24 sum.

(https://www.rfc-editor.org/rfc/rfc9580.html#name-optional-checksum)

We should disable CRC24 calculation by default, and allow optionally enabling it.

PR #604, opened 2025-07-22
Disable crc24 in dearmor by default

hko-s:
See #575, CRC24 checking is relatively expensive, and discouraged by RFC 9580.

We should disable it by default, but allow users to opt into CRC24 checking if they really want it (presumably for some kind of detail-oriented inspection of legacy artifacts).


src/armor/reader.rs
```diff
@@ -397,7 +398,25 @@ impl<R: BufRead> Dearmor<R> {
...
+    }
+
+    /// Creates a new `Dearmor` with CRC24 checking (and the provided maximum buffer size).
+    ///
+    /// Calculating and checking the CRC24 is heavily discouraged by RFC 9580:
+    /// <https://www.rfc-editor.org/rfc/rfc9580.html#name-optional-checksum>
+    ///
+    /// This function allows opting into the check for special-purpose use cases, on legacy OpenPGP
+    /// data.
+    pub fn with_crc24(input: R, limit: usize) -> Self {
```
dignifiedquire: is this enough to actually make use of it in the relevant APIs?

hko-s: I added a convenience "status" checker fn and enum in 5a9020d78226013648fb163a75b09320f69c1a0d, to see if that would be nice to have. Wdyt?

src/armor/reader.rs
```diff
@@ -903,6 +952,77 @@ y5Zgv9TWZlmW9FDTp4XVgn5zQTEN1LdL7vNXWV9aOvfrqPk5ClBkxhndgq7j6MFs
...
+        let mut res = Vec::new();
+        let _read = dec.read_to_end(&mut res).unwrap();
+
+        assert_eq!(dec.typ, Some(BlockType::Message));
+        assert_eq!(res.len(), 154);
+
+        assert!(dec.checksum.is_some());
+        assert_eq!(dec.crc24_status(), ArmorCrc24Status::Unchecked);
+
+        // A "with_crc24" Dearmor should accept the bad CRC24 without complaint
```
dignifiedquire: this comment seems wrong

src/armor/reader.rs
dignifiedquire: I think at a minimum we should mention this caveat at in the docs for the `_armor` functions, and have a doc example somewhere on how to do this

src/armor/reader.rs
hko-s: oops. indeed.

src/armor/reader.rs
hko-s: I've added a `DearmorOptions` type and unified the two `with_*` constructors into one shared `with_options` constructor.

I like the idea of exposing that in all the places (while exposing an explicit `with_crc24` seemed truly icky).

hko-s: I added a variant that takes a `DearmorOptions` to all of the public `from_armor` functions.

But I haven't added dearmoring options to the convenience wrappers, or to the armor-autodetecting functions.
Both would seem too much to me.

src/armor/reader.rs
```diff
@@ -356,19 +356,36 @@ fn armor_footer_line(i: &[u8]) -> IResult<&[u8], BlockType> {
...
+#[derive(Debug, PartialEq, Eq, Copy, Clone)]
+pub enum ArmorCrc24Status {
+    /// No CRC24 was present in the footer of the armored data
+    NoCrc24,
+
+    /// A CRC24 was present, and it matches the correct checksum of the armored data
+    CheckedOk,
+
+    /// A CRC24 was present, it does not match the correct checksum of the armored data
+    CheckedInvalid,
```
dignifiedquire: if we do the work of exposing this level of detail, we might as well expose the values?

src/armor/reader.rs
```diff
@@ -384,21 +401,79 @@ enum Part<R: BufRead> {
...
+}
+
+impl DearmorOptions {
+    /// A new DearmorOptions with default settings
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// The maximum size of armored data that this Dearmor can handle
+    pub fn get_limit(&self) -> usize {
```
dignifiedquire: is this getter actually needed? the current setup could just be done by making the fields `pub` and skipping the methods

src/composed/any.rs
```diff
@@ -33,11 +33,18 @@ impl<'a> Any<'a> {
         Self::from_armor_buf(input.as_bytes())
     }
 
-    /// Parse armored ascii data.
```
dignifiedquire: why delete this doc comment?

hko-s: :scream:

src/armor/reader.rs
hko-s: Haha. While we're going entirely over the top, yes, why not.

dignifiedquire (approved):

PR #606, opened 2025-07-22
test: impl proptest for encrypted data packets

hko-s:


src/packet/sym_encrypted_data.rs
```diff
@@ -50,3 +50,59 @@ impl PacketTrait for SymEncryptedData {
...
+#[cfg(test)]
+mod tests {
+    use prop::collection::vec;
+    use proptest::prelude::*;
+
+    use super::*;
+    use crate::types::{PacketHeaderVersion, PacketLength, Tag};
+
+    proptest::prop_compose! {
+        pub fn data_gen()(source: Vec<u8>) -> Bytes {
```
dignifiedquire: unused?

src/packet/sym_encrypted_data.rs
```diff
@@ -50,3 +50,52 @@ impl PacketTrait for SymEncryptedData {
...
+    use super::*;
+    use crate::types::{PacketHeaderVersion, PacketLength, Tag};
+
+    impl Arbitrary for SymEncryptedData {
+        type Parameters = ();
+        type Strategy = BoxedStrategy<Self>;
+
+        fn arbitrary_with(_args: Self::Parameters) -> Self::Strategy {
+            let len = 1u32; // unused
+            let packet_header = PacketHeader::from_parts(
```
dignifiedquire: ideally we have old and new packet headers, given this packet can have both

src/packet/sym_encrypted_protected_data.rs
```diff
@@ -431,4 +436,74 @@ mod tests {
...
+        type Parameters = ();
+        type Strategy = BoxedStrategy<Self>;
+
+        fn arbitrary_with(_args: Self::Parameters) -> Self::Strategy {
+            any::<Config>()
+                .prop_flat_map(move |config| (Just(config), vec(0u8..=255u8, 0..=2048)))
+                .prop_map(move |(config, data)| {
+                    let len = 1u32; // unused
+                    let packet_header = PacketHeader::from_parts(
+                        PacketHeaderVersion::New,
```
dignifiedquire, line 477: same here

src/packet/sym_encrypted_data.rs
hko-s: i'm producing both header versions for SED now

src/packet/sym_encrypted_protected_data.rs
hko-s: SEIPD (and gnupg_aead) packets can only be used with new header versions.

Old packet headers can only store packet type ids that fit into 4 bits, and these packets are type 18 and 20 respectively.

dignifiedquire (approved):

PR #609, opened 2025-07-29
fix: always produce IssuerFingerprint subpacket for signatures

### LINKED ISSUE #605: Consider producing IssuerFingerprint subpacket in signing functions (such as UserId::sign)

hko-s:
Fixes #605

Historically, signatures have signaled the issuing key with the (Key ID-based) "Issuer" subpacket.
More recently, the (newer and Fingerprint-based) "Issuer Fingerprint" subpacket has been added to OpenPGP.

In RFC 9580, generation of the "Issuer Fingerprint" subpacket is a "SHOULD:

> "The OpenPGP Key fingerprint of the key issuing the signature. This subpacket SHOULD be included in all signatures."

See: https://www.rfc-editor.org/rfc/rfc9580.html#issuer-fingerprint-subpacket

### DIFF
--- a/src/composed/cleartext.rs
+++ b/src/composed/cleartext.rs
@@ -67,10 +67,10 @@ where {
         R: rand::Rng + rand::CryptoRng,
     {
         let hashed_subpackets = vec![
-            Subpacket::regular(SubpacketData::IssuerFingerprint(key.fingerprint()))?,
             Subpacket::regular(SubpacketData::SignatureCreationTime(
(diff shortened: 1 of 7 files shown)


hko-s:
(I also re-ordered some hashed subpacket vectors. Not that it makes any semantical difference, it just seemed nice to have them all look the same :))

dignifiedquire:
can you add some notes why we should be doing this?

dignifiedquire (approved):

PR #610, opened 2025-07-30
feat: allow custom signature subpacket configuration in message builder

hko-s:


src/composed/message/builder.rs
```diff
@@ -123,6 +123,9 @@ struct SigningConfig<'a> {
     key_pw: Password,
     /// The hash algorithm to be used when signing.
     hash_algorithm: HashAlgorithm,
+
+    /// Hashed, Unhashed area subpackets
+    subpackets: Option<(Vec<Subpacket>, Vec<Subpacket>)>,
```
dignifiedquire: I would strongly prefer individual fields, rather than a tuple

dignifiedquire:
So last time I wanted to add this, the issue I saw, is that the you need some parts of the signature already, to build the subpackets, so you would want potentially more like a callback to generate them, unless they are static. Maybe this is not needed anymore, but at a minimum this should have a test, where this option is used to produce the default subpackets externally to verify this works well

src/composed/message/builder.rs
hko-s: I also hate this a bit. I wanted to encode that I never want one to be `None` while the other is `Some`, and had no better idea.

hko-s:
> So last time I wanted to add this, the issue I saw, is that the you need some parts of the signature already, to build the subpackets, so you would want potentially more like a callback to generate them, unless they are static.

I think the caller can always decide what they want in the subpackets, before calling the builder :thinking: 

> [..] at a minimum this should have a test, where this option is used to produce the default subpackets externally to verify this works well

Right, we definitely want that!

src/composed/message/builder.rs
hko-s: I've made an enum to model this more nicely

src/composed/message/builder.rs
```diff
@@ -529,6 +565,24 @@ impl<'a, R: Read, E: Encryption> Builder<'a, R, E> {
         self
     }
 
+    pub fn sign_with_subpackets(
+        &mut self,
+        key: &'a dyn SecretKeyTrait,
+        key_pw: Password,
+        hash_algorithm: HashAlgorithm,
+        hashed: Vec<Subpacket>,
```
dignifiedquire: maybe make this actually take the enum?

hko-s: It seemed a bit wasteful to add a type to the public API just for this (for now it's an internal implementation detail, and that seems reasonable to me).

dignifiedquire: I am just very worried that it is easy to mix up `hashed` and `unhashed` in this API

hko-s: Agreed that this is not ideal. But then again, this entire API invites users to shoot themselves in the foot. Mixing up the two areas is kind of the least of my worries :sweat_smile:

src/composed/message/builder.rs
```diff
@@ -114,6 +114,16 @@ pub trait Encryption: PartialEq {
     fn is_plaintext(&self) -> bool;
 }
 
+/// Subpacket configuration, per signing key
+#[derive(Debug)]
+enum SubpacketConfig {
+    Implicit,
```
dignifiedquire: Maybe we call this `Default` and `UserDefined`?
Also a great place to document the subpackets that get added by default

dignifiedquire: also `derive(Default)`  please

src/composed/message/builder.rs
```diff
@@ -204,10 +226,24 @@ where
...
+                ];
+                if config.key.version() <= KeyVersion::V4 {
+                    sig_config.unhashed_subpackets =
+                        vec![Subpacket::regular(SubpacketData::Issuer(key_id))?];
+                }
+            }
+            SubpacketConfig::Explicit { hashed, unhashed } => {
+                sig_config.hashed_subpackets = hashed.to_vec();
+                sig_config.unhashed_subpackets = unhashed.to_vec();
+            }
```
dignifiedquire: how about pushing this code into sth like this

```rust
impl SubpacketConfig {
  fn into_packets(self) -> (Vec<Subpacket>, Vec<Subpacket>) {
    // ..
  }
}
```

hko-s: I've done that, and overall I guess I like the result.

However, It's a bit messy, since this fn needs the signing key to calculate the default subpacket data.
As a side-effect of that, now the key id gets calculated twice instead of once (for <v6 signers)

src/composed/message/builder.rs
```diff
@@ -529,6 +591,22 @@ impl<'a, R: Read, E: Encryption> Builder<'a, R, E> {
         self
     }
 
+    pub fn sign_with_subpackets(
```
dignifiedquire, line 596: can I get a doc comment on this?

src/composed/message/builder.rs
```diff
@@ -114,6 +114,65 @@ pub trait Encryption: PartialEq {
     fn is_plaintext(&self) -> bool;
 }
 
+/// Subpacket configuration for a Signature packet.
+///
+/// Configuration may be implicit, relying on rPGP to set customary subpackets, or
+/// explicit, consisting of lists of hashed and unhashed subpackets.
+#[derive(Debug, Default)]
+pub enum SubpacketConfig {
```
dignifiedquire, line 122: we might need to reexport this in i `message.rs` or `composed.rs`, not sure

hko-s: right, added a reexport

dignifiedquire (approved):
I think this is a nice API we ended up with

hko-s:
> I think this is a nice API we ended up with

Agreed! Thanks as always for great ideas and pointers!

PR #611, opened 2025-07-31
WIP: user-provided session keys

hko-s:


hko-s:
Update: Maybe what we actually want is functionality to produce SEIPD packets from plaintext and a user-provided session key.

dignifiedquire (approved):

PR #612, opened 2025-08-02
handle utf-8 literal data in message builder

hko-s:
We've deferred handling text-mode [literal data packets](https://www.rfc-editor.org/rfc/rfc9580.html#name-literal-data-packet-type-id) during the streaming rewrite for 0.16.

This PR adds back support for producing text mode literals of type  `DataMode::Utf8`, in the message builder.
Data mode UTF-8 is the only somewhat well-defined data mode besides `Binary`.

RFC 9580 states:

> Text data MUST be encoded with UTF-8 [..] and stored with `<CR><LF>` text endings (that is, network-normal line endings).

For `DataMode::Utf8`, the message builder checks that the caller is providing input data that conforms to these constraints. If it doesn't, an error is returned during processing.

### DIFF
--- a/src/packet/literal_data.rs
+++ b/src/packet/literal_data.rs
@@ -354,7 +354,7 @@ impl<R: io::Read> io::Read for CrLfCheckReader<R> {
         }
 
         // Remember if the last character is a CR.
-        // If so, we'll 'check for a matching LF at the start of the next read.
+        // If so, we'll check for a matching LF at the start of the next read.
         self.last_was_cr = buf[len - 1] == b'\r';
 
         Ok(len)


Cargo.toml
```diff
@@ -40,6 +40,7 @@ smallvec = "1.10.0"
 derive_more = { version = "2.0.1", features = ["debug", "display"] }
 regex = "1.7"
 snafu = { version = "0.8.5", features = ["rust_1_81"] }
+utf8parse = "0.2.2"
```
hko-s: Oh ... now I see that this crate took a different turn, and morphed into https://crates.io/crates/vte (which doesn't focus on handling UTF-8 anymore).

I'll need to find a different way to check UTF-8.

hko-s: Switched to https://crates.io/crates/utf-8

src/lib.rs
```diff
@@ -14,6 +14,7 @@
 #[cfg(test)]
 #[macro_use]
 extern crate pretty_assertions;
+extern crate core;
```
dignifiedquire: we should not need this

src/packet/literal_data.rs
```diff
@@ -249,34 +249,167 @@ impl PacketTrait for LiteralData {
...
+
+/// Wrapping reader that checks that the input data is valid UTF-8.
+///
+/// Non-UTF-8 data in the input stream are rejected with an `io::Error`.
+pub(crate) struct Utf8CheckReader<R>
+where
+    R: io::Read,
+{
+    source: R,
+    rest: Vec<u8>,
```
dignifiedquire: should be possible to make this an array, given this is very limited length

src/lib.rs
hko-s: indeed. oops. fixed.

src/packet/literal_data.rs
hko-s: We could. But i don't think that would buy us any actual improvements for general UTF-8 data.

Note that I'm `std::mem::take`-ing the `rest` Vec in each iteration, and `extend_from_slice`-ing it with the next read.
I don't think we can (easily) avoid that work, so I don't think there are easy to obtain optimizations here.

We *could* optimize the case where there is no rest at all (which should always be the case for ascii7 data) by special-casing empty `rest`)?

hko-s: I added an optimization for streams with few multi-byte chars in 9dea818e3991b5e19bebd877fbe0e2dda5a7d53e.

I think that business logic should be as cheap as it gets, for ascii-7 streams.

dignifiedquire (approved):

PR #613, opened 2025-08-02
fix: handling of non-critical packets in messages

hko-s:


src/composed/message/parser.rs
```diff
@@ -176,6 +176,13 @@ pub(super) fn next(
                 packet.drain()?;
                 packets = crate::packet::PacketParser::new(packet.into_inner());
             }
+            Tag::Other(40..=63) => {
```
dignifiedquire: this is duplicated multiple times, I think we should expand the `Tag` honestly to reflect this

hko-s: Sounds good.

(Having multiple range-capable tags will mean we'll need to move away from `num_enum` though ... I think.)

hko-s: See 81ded3805bfd521d3b72168b855aadda0fa9f924

src/packet/single.rs
```diff
@@ -65,26 +65,37 @@ impl Packet {
...
-            Tag::Other(other) => Err(UnsupportedSnafu {
-                message: format!("Unknown packet type: {other}"),
+            // "Private or Experimental Use"
+            Tag::Experimental(id) => Err(UnsupportedSnafu {
+                message: format!("Experimental Packet type: {id}"),
             }
             .build()),
+
+            // This should only happen for "reserved number" packet types (0, 15, 16).
+            // All other cases should be covered above.
```
dignifiedquire: in this case actually match on that tag explicitly?

hko-s: good point. a5839b2a0df13e7de5ecab58c438fafff835abff

src/types/packet.rs
```diff
@@ -142,42 +142,58 @@ pub enum Tag {
     /// Padding Packet
     Padding = 21,
 
-    #[num_enum(catch_all)]
+    /// Unassigned Critical Packets [22-39]
+    #[cfg_attr(test, proptest(skip))]
+    UnassignedCritical(u8),
```
dignifiedquire: So you can now easily create invalid tags unfortunately, eg `Tag::UnassignedCritical(1)` is valid code. So I think we should do sth like this

```rust
#[repr(transparent)]
pub struct UnassignedCriticialTag(u8)

impl TryFrom<u8> for UnassignedCriticalTag {
  type Error = InvalidTagError;

  fn try_from(tag: u8) -> Result<Self, Self::Error> {
    if tag >= 22 && tag <= 39 {
       Ok(Self(tag))
    } else {
      Err(InvalidTagError)
    }
  }
}
```

src/packet/single.rs
dignifiedquire: that's not the match I was commenting on, but a good improvement in any case 😅

hko-s: Hah. After the enum reshuffling, I think I now fixed the right thing: a32bde5e333ad5842b5003fa88299cb21b161f50

src/types/packet.rs
hko-s: right, i think we've now collectively changed all of these range-y enum variants (i dropped the `Deref` impl, since it seemed noisy and not super-useful).

src/types/packet.rs
```diff
@@ -142,42 +142,142 @@ pub enum Tag {
...
+    }
+}
+
+/// Tag for Unassigned Non-Critical Packets, that can only be `40..=59`.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+#[repr(transparent)]
+pub struct UnassignedNonCriticalTag(u8);
+
+impl UnassignedNonCriticalTag {
+    /// Creates a new tag, returning `None` if it is not a valid `Other` tag.
```
dignifiedquire: docs

src/types/packet.rs
```diff
@@ -142,42 +142,142 @@ pub enum Tag {
...
-            Self::Other(i) => i,
-        };
-        0b1100_0000 | t
+/// Tag for Unassigned Critical Packets, that can only be `22..=39`.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+#[repr(transparent)]
+pub struct UnassignedCriticalTag(u8);
+
+impl UnassignedCriticalTag {
+    /// Creates a new tag, returning `None` if it is not a valid `Other` tag.
```
dignifiedquire: docs

hko-s: that problem took me a moment to see %)

PR #614, opened 2025-08-04
feat: use explicit type for RawSessionKey

dignifiedquire:
Based on #611


hko-s:
That's great! I pulled this commit into #611 and am closing this PR.

PR #616, opened 2025-08-07
fix: if one pass sig metadata doesn't match signature, don't validate

hko-s:
Refuse to validate one pass signed messages if the metadata in the One Pass Signature packet doesn't match the corresponding metadata in the Signature packet.

(Also see this thread for some more discussion of this topic: https://mailarchive.ietf.org/arch/msg/openpgp/RLMBugGhg_c9xT7zmDrkBklRZB8/)

### DIFF
--- a/src/composed/message/parser.rs
+++ b/src/composed/message/parser.rs
@@ -161,12 +161,8 @@ pub(super) fn next(
                 };
 
                 let reader =
-                    SignatureOnePassReader::new(&one_pass_signature, Box::new(inner_message))?;
-                let message = Message::SignedOnePass {
-                    one_pass_signature,
-                    reader,
-                    is_nested,
-                };
+                    SignatureOnePassReader::new(one_pass_signature, Box::new(inner_message))?;
+                let message = Message::SignedOnePass { reader, is_nested };
                 return Ok(Some(message));
             }
             Tag::CompressedData => {
--- a/src/composed/message/reader/signed_one_pass.rs
+++ b/src/composed/message/reader/signed_one_pass.rs
@@ -16,12 +16,16 @@ const BUFFER_SIZE: usize = 8 * 1024;
 #[derive(derive_more::Debug)]
 pub enum SignatureOnePassReader<'a> {
     Init {
+        /// One Pass Signature packet
+        ops: OnePassSignature,
         /// Running hasher
         norm_hasher: Option<NormalizingHasher>,
         /// Data source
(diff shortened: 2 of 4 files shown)


src/composed/message/reader/signed_one_pass.rs
```diff
@@ -64,6 +68,7 @@ impl<'a> SignatureOnePassReader<'a> {
         let norm_hasher = hasher.map(|hasher| NormalizingHasher::new(hasher, text_mode));
 
         Ok(Self::Init {
+            ops: ops.clone(),
```
dignifiedquire: can we avoid the clone?

hko-s: I've changed the APIs slightly and moved the clone up where it only happens once. That seems better.

src/composed/message/parser.rs
```diff
@@ -160,8 +160,10 @@ pub(super) fn next(
...
                 };
 
-                let reader =
-                    SignatureOnePassReader::new(&one_pass_signature, Box::new(inner_message))?;
+                let reader = SignatureOnePassReader::new(
+                    one_pass_signature.clone(),
+                    Box::new(inner_message),
+                )?;
                 let message = Message::SignedOnePass {
                     one_pass_signature,
```
dignifiedquire: we can avoid the clone, and maybe not pass it in here? but rather read it from the `reader`?

hko-s: oh yes ... looks like we didn't need the `one_pass_signature` in `Message::SignedOnePass` at all (anymore) :partying_face:

hko-s:
I added a test that constructs the following synthetic "evil" one-pass signed message:

The evil message consists of:
- a `SignatureType::Binary` signature over a text with CR+LF normalized line ending
- a literal data packet that contains an un-normalized variant of the signed message, with "LF" line ending
- a OPS that matches the signature packet, except that it shows `SignatureType::Text`

The test then checks how a message consisting of these packets is handled by the message reader. Specifically: if the reader considers the signature in the message as validating correctly.

---

With this PR, the message reader doesn't validate the signature as correct (which is the correct behavior).

Before this PR however, the signature validated as correct, even though it should not.
This was caused by the message reader "trusting" the signature type in the one-pass signature packet.
(If the signature were really a "SignatureType::Text" message - like the OnePassSignature claims - then the signature should validate as correct against the LF-only variant of the signed text.)

PR #617, opened 2025-08-07
feat: implement detached signature creation

### LINKED ISSUE #599: Add convenience functions to produce detached signatures

hko-s:
Fixes #599

### DIFF
--- a/src/composed/cleartext.rs
+++ b/src/composed/cleartext.rs
@@ -11,12 +11,16 @@ use log::debug;
 
 use crate::{
     armor::{self, header_parser, read_from_buf, BlockType, DearmorOptions, Headers},
-    composed::{ArmorOptions, Deserializable, DetachedSignature},
+    composed::ArmorOptions,
     crypto::hash::HashAlgorithm,
     errors::{bail, ensure, ensure_eq, format_err, InvalidInputSnafu, Result},
     line_writer::LineBreak,
     normalize_lines::{normalize_lines, NormalizedReader},
-    packet::{Signature, SignatureConfig, SignatureType, Subpacket, SubpacketData},
+    packet::{
+        Packet, PacketParser, PacketTrait, Signature, SignatureConfig, SignatureType, Subpacket,
+        SubpacketData,
+    },
+    ser::Serialize,
     types::{KeyVersion, Password, PublicKeyTrait, SecretKeyTrait},
 };
 
@@ -35,7 +39,7 @@ pub struct CleartextSignedMessage {
     hashes: Vec<HashAlgorithm>,
 
     /// The actual signature(s).
-    signatures: Vec<DetachedSignature>,
+    signatures: Vec<Signature>,
 }
 
 impl CleartextSignedMessage {
@@ -51,7 +55,6 @@ where {
         let signature_text = NormalizedReader::new(&mut bytes, LineBreak::Crlf);
         let hash = config.hash_alg;
         let signature = config.sign(key, key_pw, signature_text)?;
-        let signature = DetachedSignature::new(signature);
 
         Ok(Self {
             csf_encoded_text: dash_escape(text),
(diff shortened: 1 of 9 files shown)


hko-s:
✅ Location in the rPGP public API: `composed::signature`
✅ Add a variant that takes `SubpacketConfig`
❌ Is it worth implementing this optimized for handling multiple signing keys in one pass? (not doing this for now)

dignifiedquire:
API looks good, should have some basics tests to make sure it works as expected

hko-s:
> [..] should have some basics tests to make sure it works as expected

Yes! I added basic tests for all 4 functions (and still think that the API seems good :sweat_smile:)

src/composed/cleartext.rs
```diff
@@ -207,12 +207,18 @@ where {
...
-        let mut bytes = Vec::new();
-        dearmor.read_to_end(&mut bytes)?;
+        let mut signatures = vec![];
 
-        let signatures = StandaloneSignature::from_bytes_many(&bytes[..])?;
-        let signatures = signatures.collect::<Result<_>>()?;
+        let mut pp = PacketParser::new(BufReader::new(&mut dearmor));
+
+        loop {
+            match pp.next() {
```
dignifiedquire: `while let Some(packet) = pp.next()` is a bit nicer

src/composed/cleartext.rs
```diff
@@ -252,7 +258,12 @@ where {
         writer.write_all(b"\n")?;
 
         armor::write(
-            &self.signatures,
+            &self
+                .signatures
+                .iter()
+                .map(Clone::clone)
+                .map(DetachedSignature::new)
+                .collect::<Vec<_>>(),
```
dignifiedquire: not a big fan of allocating here, if you want to reuse the logic, maybe make the critical piece a standalone function?

src/composed/signature.rs
```diff
@@ -91,13 +216,177 @@ impl<I: Sized + Iterator<Item = Result<Packet>>> Iterator for SignatureParser<I>
...
+            Subpacket::regular(SubpacketData::IssuerFingerprint(alice.fingerprint())).unwrap(),
+            Subpacket::regular(SubpacketData::SignatureCreationTime(
+                chrono::Utc::now().trunc_subsecs(0),
+            ))
+            .unwrap(),
+            Subpacket::regular(SubpacketData::PolicyURI("foo".into())).unwrap(),
+        ];
+
+        let sig = DetachedSignature::sign_binary_data_with_subpackets(
+            thread_rng(),
```
dignifiedquire: please use deterministic rngs for all tests, eg chacha

hko-s: D'oh, of course. That was me being confused.

src/composed/cleartext.rs
hko-s: Oh yes! that's much nicer. Once I did that, clippy one-upped you, though: `help: try: for res in pp`.

src/composed/cleartext.rs
hko-s: The best idea i could come up with is 38fb3e8ad329373a5f1df9adcdfd75446ac5e036 - which isn't making me particularly happy.

Maybe you had a better idea and I didn't understand it? Would appreciate suggestions for improvement :sweat_smile:

dignifiedquire: this is a bit nicer

```rust
     /// A signature wrapper that serializes complete with packet header
        struct SerializableSignatures<'a>(&'a [Signature]);
        impl Serialize for SerializableSignatures<'_> {
            fn to_writer<W: std::io::Write>(&self, writer: &mut W) -> Result<()> {
                for sig in self.0 {
                    sig.to_writer_with_header(writer)?;
                }
                Ok(())
            }

            fn write_len(&self) -> usize {
                let mut sum = 0;
                for sig in self.0 {
                    sum += sig.write_len_with_header();
                }
                sum
            }
        }

        armor::write(
            &SerializableSignatures(&self.signatures),
            armor::BlockType::Signature,
            writer,
            opts.headers,
            opts.include_checksum,
        )?;

```

dignifiedquire (approved):

PR #618, opened 2025-08-07
refactor: use static dispatch in message verify functions

hko-s:


hko-s:
(Especially for `verify_nested` it's unclear if this change is an improvement, though - static dispatch doesn't allow a list that mixes concrete types.)

hko-s:
Closing this, since the change is probably not a useful tradeoff.

PR #619, opened 2025-08-07
feat: more uniform usage of primary and subkey packets

hko-s:


hko-s:
Closing this for now, we can revisit later

PR #622, opened 2025-08-19
message reader tests from the fuzzer

hko-s:
Two test cases that revolve around messages that consist of a series of `OnePassSignature` packets, followed by something malformed.

These currently cause stack overflows during message parsing.


hko-s:
A first step towards a fix is happening in #625

hko-s:
Closing in favor of #625, which incorporates these tests

PR #625, opened 2025-08-20
fix: recursion issues in deeply nested messages

dignifiedquire:
All signature packets are now read using the `SignedManyReader`, and can be accessed through that. 

# TODOS

- [x] verify regular signatures work as expected
- [x] verify performance is not regressing
- [x] fix TODO about mixed OPS and signatures

### DIFF
--- a/src/packet/marker.rs
+++ b/src/packet/marker.rs
@@ -1,7 +1,7 @@
 use std::io::{self, BufRead};
 
 use crate::{
-    errors::{Result, ensure_eq},
+    errors::{ensure_eq, Result},
     packet::{PacketHeader, PacketTrait},
     parsing_reader::BufReadParsing,
     ser::Serialize,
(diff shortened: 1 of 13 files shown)


src/composed/message/parser.rs
```diff
@@ -4,202 +4,279 @@ use std::{
...
+    fn finish(self, message: Message<'a>, is_nested: bool) -> Result<Option<Message<'a>>> {
+        if self.messages.is_empty() {
+            return Ok(Some(message));
+        }
+
+        let reader = SignatureManyReader::new(self.messages, Box::new(message))?;
+        Ok(Some(Message::Signed { reader, is_nested }))
+    }
+
+    fn visit_esk(
```
hko-s, line 195: Naming nit/question:

This function seems to process an "Encrypted Message" (as defined in the grammar in https://www.rfc-editor.org/rfc/rfc9580#name-openpgp-messages), including the Edata.

If that's right, then i'd find `visit_encrypted` (or maybe even `visit_encrypted_message`) easier to understand as a reader.

PR #627, opened 2025-08-21
fix: limit length of DSA public parameters

hko-s:
Fuzz testing produced some test cases where excessively long DSA parameters caused slow runs.

This PR limits the length of DSA public parameters based on https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-4.pdf

### DIFF
--- a/src/types/params/public/dsa.rs
+++ b/src/types/params/public/dsa.rs
@@ -1,6 +1,10 @@
 use std::io::{self, BufRead};
 
-use crate::{errors::Result, ser::Serialize, types::Mpi};
+use crate::{
+    errors::{ensure, Result},
+    ser::Serialize,
+    types::Mpi,
+};
 
 #[derive(Debug, PartialEq, Clone)]
 #[cfg_attr(test, derive(proptest_derive::Arbitrary))]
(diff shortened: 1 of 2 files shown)


hko-s:
TODO: maybe extend to 4096 bit keys?

see https://github.com/rnpgp/rnp/releases/tag/v0.18.0

hko-s:
> TODO: maybe extend to 4096 bit keys?
> 
> see https://github.com/rnpgp/rnp/releases/tag/v0.18.0

Update: The code for upcoming RustCrypto DSA version currently doesn't currently support 4096 bit keys.

See https://github.com/RustCrypto/signatures/blob/2164b728af30b1f28bd705d264c0d877c91a6a03/dsa/src/components.rs#L48-L55

src/types/params/public/dsa.rs
```diff
@@ -13,12 +17,45 @@ pub struct DsaPublicParams {
...
         let g = Mpi::try_from_reader(&mut i)?;
         let y = Mpi::try_from_reader(&mut i)?;
 
+        // Cap the lengths of these Mpis to avoid inputs that could cause (slightly)
+        // expensive calculations.
+
+        // The prime `p` is limited to L, `q` to N
+        ensure!(
+            p.len() * 8 <= Self::MAX_L_BITS,
+            format!("p is too long ({} bytes)", p.len())
```
dignifiedquire: no need for the additional `format!`?

hko-s: Right, I cleaned those up

hko-s:
I extended the allowed key size to 4096.

Rejecting real world 4096 bit keys (even if it's just a few) doesn't seem super useful. After all, my intent for this PR was to guard against totally unreasonable sizes that act as a DoS vector. Not to lock DSA key shapes down with the heaviest of hands.

dignifiedquire (approved):

PR #628, opened 2025-08-26
refactor: pub access to SecretKey, SecretSubkey, KeyDetails

hko-s:


hko-s:
This PR aims to explore the options for producing custom subpackets during key generation with `composed::key::builder`.
The current flow via `SecretKey::sign` doesn't offer a way to explicitly influence the hashed and unhashed subpacket areas - and extending the API to explicitly set them doesn't seem practical.

One approach to custom subpacket settings is to enable users of rPGP to destructure `SecretKey` and `SecretSubkey` and implement signing for the components. This PR enables that flow.

This also opens access to the fields of `KeyDetails` so that custom signers can access the metadata settings stored in those.

hko-s:
After some discussion, I'll close this for now.

We'll explore a callback-shaped mechanism for fully configurable subpacket setting in key generation in a separate PR.

PR #630, opened 2025-09-01
docs: changes based on 2024 ROS input

hko-s:
In the [2024 ROS security evaluation](https://github.com/rpgp/rpgp/blob/main/docs/SECURITY_STATUS.md#2024-12), the "Non-Findings" sections suggest a number of improvements to the documentation.

This PR addresses these pointers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,10 +44,21 @@ rPGP implements OpenPGP as specified in [RFC9580], including the commonly used v
 All formats specified in the historical RFCs [RFC4880] and [RFC6637], such as v3 keys and signatures, are supported as well.
 
 
-See [`IMPL_STATUS.md`](docs/IMPL_STATUS.md) for more details on the implemented PGP features.
+See [`IMPL_STATUS.md`](docs/IMPL_STATUS.md) for more details on the implemented PGP features and
+[`openpgp.md`](docs/openpgp.md) for a rough taxonomy.
 
 rPGP offers a flexible low-level API and gives users the ability to build higher level PGP tooling in the most compatible way possible.
 Additionally, it fully supports all functionality required by the [Autocrypt 1.1 e-mail encryption specification].
+
+## Notable Users & Libraries built using rPGP
+
+- [Delta Chat]: Cross-platform messaging app that works over e-mail
+- [`rpm`]: A pure rust library for parsing and creating RPM files
+- [`rpgpie`]: An experimental high level OpenPGP API
+- [`rsop`]: A SOP CLI tool based on rPGP and rpgpie
+- [`debian-packaging`]: a library crate for dealing with Debian packages
+
+Don't see your project here? Please send a PR :)
 
 ## Usage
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -141,6 +141,35 @@ fn main() -> pgp::errors::Result<()> {
...
 
+## rPGP is a low-level OpenPGP library
+
+rPGP offers abstractions for handling the formats and mechanisms specified in RFC 9580.
+However, it offers them as relatively low-level building blocks, and doesn't attempt to ensure that users can not apply them unsafely.
+
+> rPGP tries to allow following most parts of the OpenPGP specification, but the APIs are low level building blocks and do not claim that using them is (a) secure or (b) following the OpenPGP specification
+
+Using the building blocks in rPGP correctly and safely requires a solid understanding of OpenPGP and at least a basic understanding of cryptography.
+In this sense, rPGP should be considered a "hazmat"-level cryptographic library.
```
dignifiedquire: I think the `hazmat` label is a bit too aggressive tbh

README.md
```diff
@@ -141,6 +141,35 @@ fn main() -> pgp::errors::Result<()> {
...
+However, it offers them as relatively low-level building blocks, and doesn't attempt to ensure that users can not apply them unsafely.
+
+> rPGP tries to allow following most parts of the OpenPGP specification, but the APIs are low level building blocks and do not claim that using them is (a) secure or (b) following the OpenPGP specification
+
+Using the building blocks in rPGP correctly and safely requires a solid understanding of OpenPGP and at least a basic understanding of cryptography.
+In this sense, rPGP should be considered a "hazmat"-level cryptographic library. 
+
+### Legacy mechanisms are available (but using them may be insecure)
+
+rPGP implements support for most mechanisms in OpenPGP, both modern and legacy.
```
dignifiedquire: "both modern and those considered legacy"

README.md
```diff
@@ -141,6 +141,35 @@ fn main() -> pgp::errors::Result<()> {
...
+
+It can handle a wide range of OpenPGP artifacts.
+This explicitly includes artifacts that use historical algorithms, which are now too weak to provide adequate safety.
```
dignifiedquire: "which are considered to be insecure given todays understanding"

README.md
```diff
@@ -141,6 +141,35 @@ fn main() -> pgp::errors::Result<()> {
...
+
+Users of rPGP must ensure that their software only relies on mechanisms that are appropriate for their use case.
```
dignifiedquire: not a fan of this tbh, it's like saying don't cut yourself with a knife

README.md
```diff
@@ -141,6 +141,35 @@ fn main() -> pgp::errors::Result<()> {
...
+
+### Higher level OpenPGP semantics
+
+rPGP does not offer higher level abstractions to deal with the status of OpenPGP certificates or their components, such as *expiration* or *revocation*.
```
dignifiedquire: if we add this, this should mention the higher level constructions in `composed` as those are not high high level, but considerably higher level than dealing with raw packets

dignifiedquire: maybe mention that most of these semantics are not specified in the RFCs?

README.md
```diff
@@ -141,6 +141,35 @@ fn main() -> pgp::errors::Result<()> {
 - Security Status: [STATUS_SECURITY.md](docs/SECURITY_STATUS.md)
 - Supported Platforms: [PLATFORMS.md](docs/PLATFORMS.md)
 
+## rPGP is a low-level OpenPGP library
+
+rPGP offers abstractions for handling the formats and mechanisms specified in RFC 9580.
+However, it offers them as relatively low-level building blocks, and doesn't attempt to ensure that users can not apply them unsafely.
+
+> rPGP tries to allow following most parts of the OpenPGP specification, but the APIs are low level building blocks and do not claim that using them is (a) secure or (b) following the OpenPGP specification
```
dignifiedquire: I don't think `most parts` is necessary, we pretty much follow everything atm?

hko-s: rPGP does implement a massive subset of the things that RFCs 1991-9580 talk about.

As far as I can tell, we are currently only missing brainpool and Elgamal (which is mostly deprecated, but 9580 does still allow decryption).

That said, I'm still happy to streamline the phrasing here.

README.md
```diff
@@ -49,6 +49,16 @@ See [`IMPL_STATUS.md`](docs/IMPL_STATUS.md) for more details on the implemented
 rPGP offers a flexible low-level API and gives users the ability to build higher level PGP tooling in the most compatible way possible.
 Additionally, it fully supports all functionality required by the [Autocrypt 1.1 e-mail encryption specification].
 
+## Users & Libraries built using rPGP
```
dignifiedquire: maybe `Notable Users`

README.md
```diff
@@ -141,30 +151,72 @@ fn main() -> pgp::errors::Result<()> {
...
+Checkout [FAQ.md](docs/FAQ.md).
 
-However, the OpenPGP format and its semantics are relatively complex, and there has historically not been a reader-friendly introduction to the domain.
+## OpenPGP Versions and Features
 
-The text ["OpenPGP for application developers"](https://openpgp.dev/) offers an introduction to OpenPGP, aimed mainly at application developers. We recommend it for initial orientation.
+Checkout [openpgp.md](docs/openpgp.md).
 
-Independently, we welcome questions in the rPGP issue tracker.
+## rPGP is a low-level OpenPGP library
```
dignifiedquire, line 172: it is very confusing that this comes after a bunch of links to docs, with such a long paragraph

README.md
```diff
@@ -141,30 +151,72 @@ fn main() -> pgp::errors::Result<()> {
...
-- [`rpgpie`]: An experimental high level OpenPGP API
-- [`rsop`]: A SOP CLI tool based on rPGP and rpgpie
-- [`debian-packaging`]: a library crate for dealing with Debian packages
+> rPGP allows following almost all parts of the OpenPGP specification, but the APIs are low level building blocks and do not claim that using them is (a) secure or (b) following the OpenPGP specification
 
-Don't see your project here? Please send a PR :)
+Using the building blocks in rPGP correctly and safely requires a solid understanding of OpenPGP and at least a basic understanding of cryptography.
 
-### FAQs
+### OpenPGP is a layered technology
```
dignifiedquire, line 181: This is a great paragraph, and if i had this when I started my work, it would have saved me so many hours of confusion and head banging..

README.md
```diff
@@ -141,30 +151,72 @@ fn main() -> pgp::errors::Result<()> {
...
+Of these layers, the OpenPGP RFC specifies 1-3, while 4 is not specified in detail.
+
+Like the RFC, rPGP handles layers 1-3, but explicitly does not deal with 4.
+Applications that need OpenPGP semantics must implement them manually, or rely on additional libraries to deal with that layer.
+
+Some work on formalizing OpenPGP semantics can be found in
+[draft-gallagher-openpgp-signatures](https://datatracker.ietf.org/doc/draft-gallagher-openpgp-signatures/) and
+[draft-dkg-openpgp-revocation](https://datatracker.ietf.org/doc/draft-dkg-openpgp-revocation/).
+
+### rPGP doesn't implement support for dealing with "OpenPGP semantics"
```
dignifiedquire: not sure this section is adding much benefit, given that you already say so above effectively?

README.md
```diff
@@ -141,30 +151,72 @@ fn main() -> pgp::errors::Result<()> {
...
+Applications in which these semantics are relevant must ensure that they implement them separately.
+
+The same applies for abstractions to determine the semantical validity of component keys for a particular operation
+(such as: "which component key, if any, is currently valid for the verification of data signatures?").
+
+NOTE: The [`rpgpie`] library implements some of these high level OpenPGP semantics.
+It may be useful either to incorporate in rPGP projects, or to study for reference.
+
+### Legacy mechanisms are available (but using them may be insecure)
+
```
dignifiedquire: I wonder if this could be folded better into the layering approach, by adding a time dimension and saying that in addition to the layers, there is the time dependent evolution of the spec

README.md
```diff
@@ -141,30 +151,72 @@ fn main() -> pgp::errors::Result<()> {
...
+rPGP can handle a wide range of OpenPGP artifacts.
+It implements support for almost all mechanisms in OpenPGP, both modern and those considered legacy.
```
dignifiedquire: I think it would be useful to specify what is legacy somewhere, in a sense of being like the these RFCs, these algorithms etc

README.md
```diff
@@ -141,30 +151,72 @@ fn main() -> pgp::errors::Result<()> {
...
 
+This explicitly includes artifacts that use historical algorithms, which are considered to be insecure given today's understanding.
+
+rPGP doesn't attempt to ensure that application developers use appropriate cryptographic building blocks for their purposes
+(even though it generally produces appropriately modern artifacts, by default).
+
+## rPGP for application developers
```
dignifiedquire: I think this paragraph should come much earlier potentially, but maybe we just keep the first sentence, and add  link to openepgp.dev somewhere else

README.md
dignifiedquire: just realized you wrote this in the other doc, probably should mention/link it here

README.md
hko-s: Good point. I've cut out the redundancy and only kept the "also see rpgpie" section.

README.md
hko-s: Good point. Linked.

README.md
hko-s: I quite like short sections, each with a heading that explains what work that section does. The 'layered technology' section is already pretty long, for my tastes :confused: 

I've adjusted this section heading, and moved it into a "sub-level" of the "layered technology" section. I like that a little better.

README.md
hko-s: I've moved the whole paragraph up, and edited it to be a lot shorter.

Maybe we can move out at least the "Independently, we welcome questions in the rPGP issue tracker.", but I don't know where that could go.

It's nice to have it in a very exposed place (and not hidden behind deeper links)

PR #639, opened 2025-09-11
chore!: fix unintended visibilities

hko-s:
I think the intent is that the types in these submodules are all re-exported directly in `packet` (as they are). This PR drops some accidental duplicate re-exports.

### DIFF
--- a/src/packet/signature.rs
+++ b/src/packet/signature.rs
@@ -1,7 +1,7 @@
-pub mod config;
-pub mod de;
-pub mod ser;
-pub mod subpacket;
-pub mod types;
+mod config;
+mod de;
+mod ser;
+pub(crate) mod subpacket;
+pub(crate) mod types;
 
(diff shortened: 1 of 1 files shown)


hko-s:
I have done another round of checking of this diff, and I'm quite convinced that this PR improves clarity and doesn't remove anything from the public API:

Without this diff, some types are exported in duplicate. Specifically, types under `pgp::packet::signature::config`, `pgp::packet::signature::subpacket` and `pgp::packet::signature::types` are re-exported as `pgp::packet::config`, etc.

This duplication happens because `packet.rs` re-exports `signature::*`:

```
signature::{
        subpacket::{Subpacket, SubpacketData, SubpacketLength, SubpacketType},
        *,
    },
```

However, all of these duplicate types are already re-exported publicly in `pgp::packet`, which I think is their canonical public location. They remain visible in that module with this PR applied.

PR #640, opened 2025-09-11
Add some more API docs

hko-s:


src/composed/key/builder.rs
```diff
@@ -47,6 +47,7 @@ impl EncryptionCaps {
     }
 }
 
+/// Specification for creation of a [`SignedSecretKey`]
```
dignifiedquire: Specification is a really odd word for this? or is it just me?

hko-s: I don't love the word here either.

Maybe "Parameters" is better? (It would be aligned with the struct name, which is also nice.)

```suggestion
/// Parameters for the creation of a [`SignedSecretKey`]
```

dignifiedquire: yeah parameters is probably better

hko-s: Cool. I adjusted the two occurrences of this wording.

src/composed/key/builder.rs
```diff
@@ -380,6 +382,7 @@ impl SecretKeyParams {
     }
 }
 
+/// Specifies the cipher of a key packet
 #[derive(Clone, Debug, PartialEq, Eq)]
```
dignifiedquire, line 386: still says specifiers

src/composed/key/builder.rs
```diff
@@ -423,6 +426,7 @@ pub enum KeyType {
     SlhDsaShake256s,
 }
 
+/// DSA key size specification
```
dignifiedquire: this is the size, not a specification

hko-s:
> some last nits, otherwise lgtm

great, thanks! i adjusted three occurrences of "specify" for different wording.

PR #642, opened 2025-09-18
docs: move information about features to README

hko-s:
Breaking out this side quest from https://github.com/rpgp/rpgp/pull/636#discussion_r2358052142

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,13 +50,6 @@ See [`IMPL_STATUS.md`](docs/IMPL_STATUS.md) for more details on the implemented 
 rPGP offers a flexible low-level API and gives users the ability to build higher level PGP tooling in the most compatible way possible.
 Additionally, it fully supports all functionality required by the [Autocrypt 1.1 e-mail encryption specification].
 
-### Cargo features
-
-- `bzip2`: Enables bzip2 support
-- `asm`: Enables assembly based optimizations
-- `wasm`: Allows building for wasm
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -50,6 +50,13 @@ See [`IMPL_STATUS.md`](docs/IMPL_STATUS.md) for more details on the implemented
 rPGP offers a flexible low-level API and gives users the ability to build higher level PGP tooling in the most compatible way possible.
 Additionally, it fully supports all functionality required by the [Autocrypt 1.1 e-mail encryption specification].
 
+Cargo features:
```
dignifiedquire: should probably be an `h3´

README.md
```diff
@@ -50,6 +50,13 @@ See [`IMPL_STATUS.md`](docs/IMPL_STATUS.md) for more details on the implemented
 rPGP offers a flexible low-level API and gives users the ability to build higher level PGP tooling in the most compatible way possible.
 Additionally, it fully supports all functionality required by the [Autocrypt 1.1 e-mail encryption specification].
 
+### Cargo features
```
dignifiedquire: maybe it should be under `Usage`?

hko-s: Right. Sounds like the best place.

I do feel that it's a bit hidden, so far down the page, there. But that's probably a task for a separate future step: streamlining the top level README some more.

PR #644, opened 2025-09-25
fix: use correct key size when using s2k derive_key

### LINKED ISSUE #643: Interoperability issues with encrypted RFC9580 keys.
Hi there.
I noticed that my program, which uses rpgp, was unable to open encrypted RFC9580 keys generated with SequoiaPGP. Having no idea where this came from, I opened an issue in the OpenPGP interoperability test repo:
https://gitlab.com/sequoia-pgp/sequoia/-/issues/1202#note_2771228917
The issue has moved around a bit, and the tests were extended. This new systematic data seems to point towards an issue with rpgp:
https://sequoia-pgp.gitlab.io/openpgp-interoperability-test-suite/results.html?profiles=1#Key_generation_with_password__sign-verify_inverse_roundtrip
I also created a minimal example in case that helps reproducing the issue:
https://github.com/SimonHeidrich/open_pgp_tests
But I guess the new interop test does a better job at that. :)
Could someone who knows their way around this here repo have a look?
Thanks a lot,
Simon

dignifiedquire:
This would only manifest with keys that have different sizes, so it wasn't caught earlier.

Fixes #643


hko-s (approved):
LGTM :+1:

hko-s:
(Also: oops! :sweat_smile:)

PR #648, opened 2025-10-14
feat!: improve handling of large RSA keys

hko-s:
The OpenPGP standard doesn't clearly specify what RSA key sizes should be supported.
In practice, sizes up to 4096 bit are widely used, while larger key sizes are exceedingly uncommon. However some implementations support (or have historically supported) larger key sizes (8192 bit, or even larger).

rPGP has so far not supported cryptographic operations with RSA keys >4096 bit (however, it has supported *parsing* larger keys from key packets).

With this PR, rPGP rejects all operations on RSA key packets >4096 bit by default (including parsing key packets).

On the other hand, when the feature `malformed-artifact-compat` is enabled, large RSA keys (up to 16384 bit) are now supported for both public and private key operations.


hko-s:
This test works with https://github.com/RustCrypto/RSA/pull/594

hko-s:
Note to self: maybe limit `pgp::crypto::rsa::MAX_KEY_SIZE` to 4096 by default? (... and RSA key creation?)

hko-s:
(Note: The test in this PR is broken until we use a release >0.9.8 of RSA, that includes https://github.com/RustCrypto/RSA/pull/594, so this PR needs to wait for an RSA release before merging)

I've feature-gated larger RSA keys on `malformed-artifact-compat`.
Not sure if I entirely like that exact approach, but it seems kind of preferable to reject large RSA keys by default.

An alternative option would be that we refuse to generate large RSA keys, but always allow use of existing RSA keys (of any size).

If we reject large public keys by default, we need to adjust the SKS tests.

src/types/params/public/rsa.rs
```diff
@@ -4,6 +4,12 @@ use rsa::traits::PublicKeyParts;
 
 use crate::{errors::Result, ser::Serialize, types::Mpi};
 
+/// MAX_KEY_SIZE limits rsa key size while parsing public key packets
+#[cfg(not(feature = "malformed-artifact-compat"))]
```
dignifiedquire: lets update the readme with this info

hko-s: oops, yes!

tests/key_test.rs
```diff
@@ -151,99 +151,99 @@ parse_dumps!(
     (
         test_parse_dumps_0,
         0,
-        17_724,
```
dignifiedquire, line 154: can you test that all these numbers don't change iff running with the compat flag?

hko-s: When the tests are run with the feature `malformed-artifact-compat`, then the test results diverge from the test expectations (and fail).

Specifically, the tests: `test_many_parser` and `test_parse_dumps_*` now fail when run with `--features malformed-artifact-compat`.

One way or another, the test expectations must (implicitly) encode which RSA key sizes we expect to be able to handle (and that impacts the SKS parsing test runs).
Previously, the mass tests expected to handle RSA keys up to 16k. My changed test expectations now assume rPGP will only handle RSA keys up to 4k.

Running the tests with `--features malformed-artifact-compat` now causes the test expectations from before the changes in 0277643780d6222d7bfcca978743e057e9b5e7af to be correct.

In principle, we could also encode different expectations with and without the feature `malformed-artifact-compat`. But ... ugh :grimacing: that seems like a road to combinatorical horrors.

PR #650, opened 2025-10-21
refactor: in text signatures, don't normalize single `\r` characters

### LINKED ISSUE #649: Align text-mode signature normalization
dkg points out that the OpenPGP ecosystem is not currently uniformly normalizing line-endings for text-mode signatures:

https://mailarchive.ietf.org/arch/msg/openpgp/NzVArGf7JiBVm-3BmMdkFH_1nvM/

Some implementations normalize `\r` into `\r\n`, while others don't. rPGP currently does.
I'm leaning strongly towards changing rPGP's behavior, and stopping our normalization of `\r`.

If we want to do this, I'll make a PR for it sometime.

hko-s:
Fixes #649

### DIFF
--- a/src/packet/literal_data.rs
+++ b/src/packet/literal_data.rs
@@ -304,13 +304,6 @@ impl<R: io::Read> io::Read for CrLfCheckReader<R> {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
         let len = self.source.read(buf)?;
 
-        // If the previous read ended with a CR, an LF must follow now
-        if self.last_was_cr && (len == 0 || buf[0] != b'\n') {
-            return Err(io::Error::other(
-                "Illegal line ending (CR without matching LF)",
-            ));
-        }
-
         // Reading is done, no more checks required
         if len == 0 {
             return Ok(0);
@@ -318,27 +311,25 @@ impl<R: io::Read> io::Read for CrLfCheckReader<R> {
 
         // Check the body of this read for any illegal linebreaks
 
-        // Skip the first byte if it was a matching LF
-        let mut pos = if self.last_was_cr { 1 } else { 0 };
+        // Skip the first byte if it is a matching LF
+        let mut pos = if self.last_was_cr && buf[0] == b'\n' {
+            1
+        } else {
+            0
+        };
 
         // Inspect data from the start until the second-to-last byte
         // (because we want to look ahead one byte, if "pos" is a CR)
         while pos < len - 1 {
-            // bare linefeed is not ok
+            // A standalone linefeed is not ok
             if buf[pos] == b'\n' {
                 return Err(io::Error::other(
                     "Illegal line ending (LF without preceding CR)",
                 ));
             }
 
-            // CR must be followed by LF
-            if buf[pos] == b'\r' {
-                if buf[pos + 1] != b'\n' {
-                    return Err(io::Error::other(
-                        "Illegal line ending (CR without matching LF)",
-                    ));
-                }
-
+            // Skip CR followed by LF in one go
+            if buf[pos] == b'\r' && buf[pos + 1] == b'\n' {
                 pos += 2;
             } else {
                 pos += 1;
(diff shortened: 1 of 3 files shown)


src/util.rs
```diff
@@ -234,14 +235,14 @@ impl NormalizingHasher {
...
-                                self.hasher.update(b"\r\n");
-
                                 // we are guaranteed to have at least two bytes
                                 if buf[1] == b'\n' {
                                     // there was a '\n' in the stream, we consume it as well
+                                    self.hasher.update(b"\r\n");
                                     buf = &buf[2..];
                                 } else {
-                                    // this was a lone '\r', we have normalized it
+                                    // this was a lone '\r', we don't normalized it
```
dkg:
```suggestion
                                    // this was a lone '\r', we don't normalize it
```

hko-s: Whoops. Right, thanks!

src/packet/literal_data.rs
```diff
@@ -940,19 +931,19 @@ mod tests {
 
             let string = random_string(&mut rng, len);
 
-            if !string.contains('\n') && !string.contains('\r') {
+            // We want to produce strings with "illegal" linebreaks.
+            // So we skip all strings that don't fit this goal.
+            //
+            // Test strings for this case need to contain a LF.
+            // And they must not normalize to legal CR+LF (like CR+CR+LF) does.
```
dkg: I don't understand this comment.  CR+CR+LF might be the most contentious part of normalization, as it distinguishes the implementers who treat `\r?\n` as the line ending from those who treat `\r*\n` as the line ending.

hko-s: This test is a "fuzzing" style test. It attempts to produce random messages that are guaranteed to have illegal line endings (and then verifies that `CrLfCheckReader` does indeed throw an error while consuming that message).

It does that by first filtering random strings as candidates, and then normalizing  CR+LF to just LF for all candidates that we expect will indeed have illegal line endings after this step (this "counter-"normalization happens in the next statement, which reads `let norm = normalize_lines(&string, LineBreak::Lf);`)

The check you're looking at is throwing away random strings that are not going to trigger an error, because they either:

- contain no LF at all (those can't possibly be illegal in terms of their line ending normalization)
- contain CR+CR+LF, which would get normalized to just CR+LF (which is legal, so it's not an input that we want to test in this particular test function - we want to randomly construct illegal inputs)

hko-s: And just as a tiny bit more context (even though it's entirely separate from what you asked for):

This test wraps the (pesudo-random) test messages in a `ChaosReader` before trying to process it with the `CrLfCheckReader` (which is the actual subject of this test). `ChaosReader` is a Reader that returns the message chopped up into (pseudo-)random slices with lengths between one byte and the full message length, each `read` call returns a (pseudo-)random number of bytes..

This exercises handling of "unfortunate" read lengths (such as right in the middle of a CR+LF pair) in the reader that gets tested (again, here: `CrLfCheckReader`).

By turning up the number of iterations of the test, this test method has proven pretty thorough in finding missed edge cases in code like this.

dkg: i guess the point i was trying to make is that while GnuPG might transform `\r\r\n` to a "legal line ending", it's not at all clear that this is actually what "normalizing" means in the context of an OpenPGP text message.  So i think the comment could be improved, unless you're trying to make some sort of assertion that OpenPGP line ending normalization actually is `\r*\n` → `\r\n`

I really do appreciate the ChaosReader fuzzing approach for this normalization silliness, i've seen more than i would like of that sort of breakage, so i'm glad you're checking for it.

hko-s: To clarify the test some more: The transformation that is used as part of preparing error-cases in this test is actually a "denormalization", which just replaces every occurrence of `\r\n` with `\n`.

This particular transformation is (I think) not used in any production code path within rPGP. It's just used in this test to produce error test cases. Those cases are then thrown at the `CrLfCheckReader`, to  make sure it rejects all manner of strings with illegal line-endings.

The comment you're looking at is effectively saying: We're trying to produce strings that will have unacompanied `\n` characters (that do not form proper `\r\n` pairs). I'll clarify the comment to say that.

---

Independently, I'm starting to see your point that different software might have more varies ideas of what it means to "normalize" line endings than I imagined. That is a very disturbing point.

hko-s: I've pushed a more verbose comment. I am still a bit unsure if I'm missing some point, but I am relatively confident that the test does what it should.

dignifiedquire (approved):

Issue #651, opened 2025-10-26
Can't encrypt MessageBuilder to multiple recipeint keys


alexng353:
Worked in v0.15.x, encrypt_to_keys_seipd_v1().

alexng353:
Nevermind, closing because this is more of a docs issue than a crate issue.

PR #652, opened 2025-10-27
alex/docs 01

alexng353:
- **docs: add example + references for encrypt_to_key on Builder+EncryptionSeipdV1**
- **docs: add example + references for Argon2 S2K and new_argon2**

Added docs for things I ran into issues with from #651

### DIFF
--- a/src/composed/message/builder.rs
+++ b/src/composed/message/builder.rs
@@ -397,12 +397,48 @@ impl<R: Read> Builder<'_, R, EncryptionSeipdV1> {
     }
 
     /// Encrypt to a public key
+    ///
+    /// Adds a [PK ESK] packet to the message for a given public key. Call this function multiple
+    /// times in order to add multiple ESKs.
+    ///
+    /// # Example
+    /// ```rust,no_run
+    /// use pgp::composed::MessageBuilder;
+    /// use pgp::types::PublicKeyTrait;
+    /// use pgp::crypto::sym::SymmetricKeyAlgorithm;
+    ///
+    /// # use pgp::types::{KeyDetails, SignatureBytes, PublicParams, KeyId, Fingerprint, KeyVersion};
+    /// # use pgp::crypto::hash::HashAlgorithm;
+    /// # use pgp::crypto::public_key::PublicKeyAlgorithm;
+    /// # use pgp::errors::Result;
+    /// fn get_encryption_key(_lookup: &str) -> pgp::packet::PublicSubkey {
+    ///     // [...]
+    /// #   panic!("dummy getter");
+    /// }
+    ///
+    /// let public_key_1 = get_encryption_key("alice");
+    /// let public_key_2 = get_encryption_key("bob");
+    ///
+    /// let message = "Hello, world!";
+    ///
+    /// let mut rng = rand::thread_rng();
+    /// let mut builder = MessageBuilder::from_bytes("", message.as_bytes())
+    ///    .seipd_v1(&mut rng, SymmetricKeyAlgorithm::AES256);
+    ///
+    /// builder.encrypt_to_key(&mut rng, &public_key_1);
+    /// // Optionally, you can also encrypt to multiple keys
+    /// builder.encrypt_to_key(&mut rng, &public_key_2);
+    /// ```
+    ///
+    /// [PK ESK]: https://www.rfc-editor.org/rfc/rfc9580#name-public-key-encrypted-sessio
+    /// # References
+    /// [RFC 9580 &sect; 5.1 - Public Key Encrypted Session Key Packet](https://www.rfc-editor.org/rfc/rfc9580#section-5.1)
     pub fn encrypt_to_key<RAND, K>(&mut self, mut rng: RAND, pkey: &K) -> Result<&mut Self>
     where
         RAND: CryptoRng + Rng,
         K: crate::types::PublicKeyTrait,
     {
-        // Encrypt (sym) the session key using the provided password.
+        // Encrypt (asym) the session key using the provided public key.
         let pkes = PublicKeyEncryptedSessionKey::from_session_key_v3(
             &mut rng,
             &self.encryption.session_key,
@@ -414,6 +450,8 @@ impl<R: Read> Builder<'_, R, EncryptionSeipdV1> {
     }
 
     /// Encrypt to a public key, but leave the recipient field unset
+    ///
+    /// See [encrypt_to_key][Self::encrypt_to_key] for more information
     pub fn encrypt_to_key_anonymous<RAND, K>(
         &mut self,
         mut rng: RAND,
@@ -423,7 +461,7 @@ impl<R: Read> Builder<'_, R, EncryptionSeipdV1> {
         RAND: CryptoRng + Rng,
         K: crate::types::PublicKeyTrait,
     {
-        // Encrypt (sym) the session key using the provided password.
+        // Encrypt (asym) the session key using the provided public key.
         let mut pkes = PublicKeyEncryptedSessionKey::from_session_key_v3(
             &mut rng,
             &self.encryption.session_key,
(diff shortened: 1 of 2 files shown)


alexng353:
working on reading through rpgp documentation + code (particularly on builder & string 2 key) and updating the docs

alexng353:
about the comments i made on [64c56b7](https://github.com/rpgp/rpgp/pull/652/commits/64c56b75a7f8d337e3369cc6ce664a32e4e2df1f), I'm not sure it's actually possible to do.

hko-s:
Hey, thanks for reaching out and working on docs! I'm also in the process of improving the API docs, and appreciate your additional eyeballs on this!

(I'll have a closer look soon, just wanted to send a quick sign of life now)

alexng353:
> Hey, thanks for reaching out and working on docs! I'm also in the process of improving the API docs

I also saw your PR #640. Should I change PR into your branch?

hko-s:
> I also saw your PR #640. Should I change PR into your branch?

I'm up for both approaches, but would suggest that we do your PR separately, to land it faster (mine will take some more time)

alexng353:
> > I also saw your PR #640. Should I change PR into your branch?
> 
> I'm up for both approaches, but would suggest that we do your PR separately, to land it faster (mine will take some more time)

Okay. I will be adding one or two more commits (documentation on seipd v2 for MessageBuilder). Let me quickly convert this to a draft, and when I finish with that stuff, I'll convert it back. Thanks for the sign of life!

hko-s:
Sounds great, looking forward, thanks again!

src/composed/message/builder.rs
```diff
@@ -534,6 +659,32 @@ impl<R: Read> Builder<'_, R, EncryptionSeipdV2> {
...
+    /// use pgp::composed::{ArmorOptions, Message, MessageBuilder};
+    /// use pgp::crypto::{aead::{AeadAlgorithm, ChunkSize}, sym::SymmetricKeyAlgorithm};
+    /// use pgp::types::{CompressionAlgorithm, StringToKey};
+    ///
+    /// let message = "Hello, world!";
+    /// let password = "password";
+    ///
+    /// let mut rng = rand::thread_rng();
+    /// let mut msg = MessageBuilder::from_bytes("", message.as_bytes())
+    ///     .seipd_v2(&mut rng, SymmetricKeyAlgorithm::AES256, AeadAlgorithm::Gcm, ChunkSize::default());
```
hko-s:
```suggestion
    ///     .seipd_v2(&mut rng, SymmetricKeyAlgorithm::AES256, AeadAlgorithm::Ocb, ChunkSize::default());
```

Super minor point of preference: I'd suggest using OCB here, as the most "default" mode. It is the one "mandatory to implement (MTI)" AEAD mode in the spec:

https://www.rfc-editor.org/rfc/rfc9580#appendix-B-2.3.1

src/composed/message/builder.rs
```diff
@@ -493,12 +549,79 @@ impl<R: Read> Builder<'_, R, EncryptionSeipdV2> {
...
+    /// # Example
+    /// ```rust,no_run
+    /// use pgp::composed::MessageBuilder;
+    /// use pgp::types::PublicKeyTrait;
+    /// use pgp::crypto::sym::SymmetricKeyAlgorithm;
+    ///
+    /// # use pgp::types::{KeyDetails, SignatureBytes, PublicParams, KeyId, Fingerprint, KeyVersion};
+    /// # use pgp::crypto::hash::HashAlgorithm;
+    /// # use pgp::crypto::public_key::PublicKeyAlgorithm;
+    /// # use pgp::errors::Result;
```
hko-s, line 578: How do you feel about this shorter method of obtaining "fake keys"?

```
    /// # use pgp::errors::Result;
    /// fn get_encryption_key(_lookup: &str) -> pgp::packet::PublicSubkey {
    ///     // [...]
    /// #     panic!("dummy getter");
    /// }
    /// #
    /// let public_key_1 = get_encryption_key("alice");
    /// let public_key_2 = get_encryption_key("bob");
```

hko-s: (The same approach would also make the example in line 410 buildable.)

src/composed/message/builder.rs
```diff
@@ -440,6 +469,33 @@ impl<R: Read> Builder<'_, R, EncryptionSeipdV1> {
     }
 
     /// Encrypt to a password.
+    ///
+    /// # Example
+    ///
+    /// ```rust
```
hko-s: I'm torn if I'd rather set this to `no_run` (to avoid paying the cost of running argon2 in the doc tests).

I tend to think that letting this just compile is enough, and running doesn't add *that* much additional value?

src/composed/message/builder.rs
```diff
@@ -397,12 +397,39 @@ impl<R: Read> Builder<'_, R, EncryptionSeipdV1> {
...
+    /// let mut rng = rand::thread_rng();
+    /// let mut builder = MessageBuilder::from_bytes("", message.as_bytes())
+    ///    .seipd_v1(&mut rng, SymmetricKeyAlgorithm::AES256);
+    ///
+    /// builder.encrypt_to_key(&mut rng, &public_key_1);
+    /// // Optionally, you can also encrypt to multiple keys
+    /// builder.encrypt_to_key(&mut rng, &public_key_2);
+    /// ```
+    ///
+    /// [PK ESK]: https://www.rfc-editor.org/rfc/rfc9580#name-version-3-public-key-encryp
```
hko-s: Is there a reason to link specifically to the v3 PKESK section here?

Maybe I'm missing something, but I think linking to the enclosing section (that encompasses both v3 and v6 PKESK) `https://www.rfc-editor.org/rfc/rfc9580#name-public-key-encrypted-sessio` would be better?

src/composed/message/builder.rs
```diff
@@ -493,12 +549,79 @@ impl<R: Read> Builder<'_, R, EncryptionSeipdV2> {
...
```
hko-s: Like above, I think `https://www.rfc-editor.org/rfc/rfc9580#name-public-key-encrypted-sessio` might be the better link here?

hko-s (approved):
Thanks for writing these docs, they are great! I left a few ideas/nits, and look forward to merging this soon :)

alexng353:
Hi @hko-s , sorry I didn't see your messages before. Let me fix up the things you've mentioned. Thanks!

hko-s:
> Hi @hko-s , sorry I didn't see your messages before. Let me fix up the things you've mentioned. Thanks!

No problem, thanks for having a look now!

src/composed/message/builder.rs
alexng353: I can set it to garbage argon2 settings to make it faster (it does use 2gb ram as it stands), but that's not good as an example if people copy it. Downside is that it currently takes 18 seconds to run the doctest.

hko-s: Not running the test here in the docs seems like the best tradeoff to me. We do run the functionality in unit tests, elsewhere, anyway.

alexng353:
What do you think about the argon2 stuff?

hko-s:
> What do you think about the argon2 stuff?

Do you mean the "slow" doc test? I still favor setting that to `no_run`.

Or is there another open question that I've missed?

hko-s:
Cool, looking good!

Are you up for doing a final rebase and making fmt happy?

alexng353:
> Do you mean the "slow" doc test? I still favor setting that to no_run.

yep

> Are you up for doing a final rebase and making fmt happy? 

rebase all into one commit?

hko-s:
> rebase all into one commit?

rebase on main, and squash into one commit, would be great, yes!

alexng353:
Ok, now I'm really done. Thanks for your guidance @hko-s !

hko-s:
> Ok, now I'm really done. Thanks for your guidance @hko-s !

Awesome, thanks for the very neat work! This looks great to me now.

PR #659, opened 2025-11-05
additional text signature normalization excitement

hko-s:
`NormalizedReader` in `LineBreak::Lf` mode doesn't handle `\r\n` correctly when a read is split between the two characters.

### DIFF
--- a/src/normalize_lines.rs
+++ b/src/normalize_lines.rs
@@ -2,7 +2,7 @@
 
 use std::sync::LazyLock;
 
-use bytes::{Buf, BytesMut};
+use bytes::{Buf, BufMut, BytesMut};
 
 use crate::{line_writer::LineBreak, util::fill_buffer};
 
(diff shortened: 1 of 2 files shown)


hko-s:
I've updated this PR with the fix from #660, and added comments until I understood *why* the fix is correct.

Unless @lubux disagrees with the contents of my additional comments, I'd propose to squash and merge this.

src/normalize_lines.rs
```diff
@@ -33,33 +33,66 @@ impl<R: std::io::Read> NormalizedReader<R> {
...
-        let in_buffer = if have_split_crlf && read > 0 {
-            // skip the first byte of the buffer, which is a `\n` as it was already handled before
-            &self.in_buffer[1..read]
-        } else {
-            &self.in_buffer[..read]
-        };
-
-        let res = RE.replace_all(in_buffer, self.line_break.as_ref());
+    fn cleanup_buffer(&mut self, read: usize, last_char: u8) {
+        const CR: u8 = b'\r';
```
dignifiedquire, line 61: pretty sure we have constants for these somewhere already

hko-s: We have similar constants in `line_writer.rs`, but those are arrays, not plain `u8`. Other than that, I can't find these constants elsewhere. I have no strong feelings about the constants. I'm happy to have them here, define them elsewhere, or to use `\r` literals instead of them.

src/normalize_lines.rs
```diff
@@ -33,33 +33,66 @@ impl<R: std::io::Read> NormalizedReader<R> {
...
     fn fill_buffer(&mut self) -> std::io::Result<()> {
-        // edge case, if the last byte of the previous buffer was `\r` and the first of the new is `\n`
-        // we need to make sure to correctly handle it.
-        let last_was_cr = self.in_buffer[self.in_buffer.len() - 1] == b'\r';
+        // edge case: if the last byte of the previous read was `\r` and the first of the new read
+        // is `\n` we need to make sure to handle it correctly.
+
+        // the previous read was guaranteed to have filled up the buffer (otherwise we would have
+        // switched to `self.is_done`, so this is the last byte of the previous read.
+        // if it is a CR, we haven't handled in during the last call.
```
lubux:
```suggestion
        // If this is a CR, it wasn’t handled in the previous call.
```

src/normalize_lines.rs
```diff
@@ -33,33 +33,66 @@ impl<R: std::io::Read> NormalizedReader<R> {
...
+
+        // Handle edge case where the last byte of the previous buffer was `\r`.
+        let edge_case = [last_char, self.in_buffer[0]];
+        match (edge_case, read > 0) {
+            ([CR, LF], true) => {
+                // Edge case, we need to normalize this pair of bytes separately.
+                let res = RE.replace_all(&edge_case, self.line_break.as_ref());
+                self.replaced.extend_from_slice(&res);
+
+                // We consumed the LF at the beginning of `self.in_buffer` and skip it in the following
```
lubux:
```suggestion
                // We already processed the leading LF in `self.in_buffer`,  so it’s omitted from the final normalization step.
```

PR #660, opened 2025-11-06
Fix edge case in NormalizedReader with LF mode

lubux:
This fixes a failing test where NormalizedReader in LineBreak::Lf mode did not normalize `\r` at the end of one buffer followed by `\n` at the start of the next. The issue was caused by a regex change from `r"(\r\n?|\n)" to r"(\r\n|\n)"`, which left `\r | \n` unhandled. This PR updates NormalizedReader to correctly replace `\r | \n` with `\n` across buffer boundaries.

(Ref https://github.com/rpgp/rpgp/pull/659)

### DIFF
--- a/src/normalize_lines.rs
+++ b/src/normalize_lines.rs
@@ -2,7 +2,7 @@
 
 use std::sync::LazyLock;
 
-use bytes::{Buf, BytesMut};
+use bytes::{Buf, BufMut, BytesMut};
 
 use crate::{line_writer::LineBreak, util::fill_buffer};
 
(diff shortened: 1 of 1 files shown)


src/normalize_lines.rs
```diff
@@ -37,29 +37,47 @@ impl<R: std::io::Read> NormalizedReader<R> {
     fn fill_buffer(&mut self) -> std::io::Result<()> {
         // edge case, if the last byte of the previous buffer was `\r` and the first of the new is `\n`
         // we need to make sure to correctly handle it.
-        let last_was_cr = self.in_buffer[self.in_buffer.len() - 1] == b'\r';
+        let last_char = self.in_buffer[self.in_buffer.len() - 1];
         let read = fill_buffer(&mut self.source, &mut self.in_buffer, None)?;
-        if read == 0 {
+        if read < self.in_buffer.len() {
```
hko-s, line 42: Hmm. I wonder if this comparison is always a solid signal that the underlying reader is done?

hko-s: Ah, nevermind. I see, `pgp::util::fill_buffer` reads aggressively.

hko-s:
Closing in favor of #659 (which adds some comments)

PR #662, opened 2025-11-13
feat!: by default, MDC-check first when decrypting SEIPDv1; allow optional streaming

hko-s:
## Context

When decrypting SEIPDv1 data packets, the integrity of the ciphertext can only be ascertained after processing the entire data packet.

## Change: Two different SEIPDv1 decryption modes

This PR introduces two distinct decryption modes for SEIPDv1 encrypted data:

- `Seipdv1ReadMode::CheckFirst` (with a caller-specified maximum allowed message size)
- `Seipdv1ReadMode::Streaming`

The new default behavior for decryption of SEIPDv1 data packets is changed to `CheckFirst` with a size limit of 1 GByte (before this PR, `Streaming` mode was implicitly used).

In `CheckFirst` mode, the entire data packet is read into memory and decrypted in one go, and the MDC integrity protection is checked at the same time. Only after this does any plaintext get released to the caller (or processed by an inner message parser).

This has the benefit that callers (and the message parser) will never process unauthenticated (and potentially manipulated) plaintext.
As a downside, `CheckFirst` decryption requires at least as much memory as the plaintext size, so it may be prohibitive for large messages.

The streaming decryption behavior can be configured by callers via [`DecryptionOptions::set_seipdv1_read_mode`].


src/crypto/sym/decryptor.rs
```diff
@@ -444,7 +590,7 @@ where
                         protected,
                         ..
                     } => {
-                        if let MaybeProtected::Protected { mut hasher } = protected {
+                        if let MaybeProtected::Protected { mut hasher, .. } = protected {
                             // last read
                             if buffer.remaining() < MDC_LEN {
                                 return Err(io::Error::new(
```
lubux: should we not change this to:

```
return Err(io::Error::new(
       io::ErrorKind::InvalidInput,
       "invalid MDC",
));
```

src/crypto/sym/decryptor.rs
```diff
@@ -338,47 +451,80 @@ where
...
+                                let to_read = buf_size - current_len;
+                                let read = fill_buffer_bytes(source, buffer, buf_size)?;
+
+                                read < to_read
+                            }
+                            DecryptionMode::Seipdv1CheckFirst => {
+                                // Try to read the entire input stream in one go
+                                // (we ask for MAX+1 bytes to see if the input stream is too long)
+                                let read =
+                                    fill_buffer_bytes(source, buffer, MAX_UNSTREAMED_MSG_SIZE + 1)?;
```
lubux: For me, this part was a bit confusing:

The fill_buffer_bytes function does not make it clear that the buffer will be reallocated if the input length exceeds the current buffer length. It might be worth adding a comment to clarify this behavior.

The logic itself is correct, since put_slice in fill_buffer_bytes appears to handle reallocation automatically, but that was not immediately obvious to me without reading deeper.

src/crypto/sym/decryptor.rs
```diff
@@ -23,6 +28,10 @@ use crate::{
 const MDC_LEN: usize = 22;
 const BUFFER_SIZE: usize = 1024 * 8;
 
+/// The maximum message length that we're willing to decrypt in non-streaming SEIPDv1 mode
+// TODO: maybe make this configurable?
```
lubux: Yes,  I think this should be configurable.
Maybe instead of passing a `allow_unauthenticated_streaming` bool around, a dedicated enum would allow to encode the max allocation as well.

src/crypto/sym/decryptor.rs
lubux, line 395: With the change this function now stretches over even more lines. Might be worth refactoring it a bit into separate helpers.

lubux: Hmm does the loop make sense if we anyways only execute max twice in the `Self::Prefix { .. } => (true, false)` case? i.e., it is essentially a goto in disguise.

lubux: -> What do you think about this: https://github.com/lubux/rpgp/commit/b63b448a721fc0c7b53a65680ead67bd36ac6788

hko-s:
I added a unit test that triggers a regression in this PR, in 45d8f31eaf2bf478ccd28458b6189ed3303779ac.

The issue was found (and formalized into this test) by Christian Reitter from Radically Open Security.

src/crypto/sym/decryptor.rs
hko-s: > -> What do you think about this: [lubux@b63b448](https://github.com/lubux/rpgp/commit/b63b448a721fc0c7b53a65680ead67bd36ac6788)

:star_struck: ... oh wow. Yes!

src/crypto/sym/decryptor.rs
hko-s: Good point (I see you've already done that in your refactor commit, which is now in this PR)!

(Independently, I now think it would be nicer to wrap an rpgp `Error::MdcError` with `std::io::Error::other`, instead of using string literals here)

src/crypto/sym/decryptor.rs
hko-s: Agreed.

I've changed the parameters, and think the new state is an improvement. It all feels clumsy and I'm not sure about the naming - but I guess there's no way to make these parameters truly smooth.

(If you have any ideas for additional improvements, I'd love to make the naming better though)

hko-s:
> I added a unit test that triggers a regression in this PR, in [45d8f31](https://github.com/rpgp/rpgp/commit/45d8f31eaf2bf478ccd28458b6189ed3303779ac).
> 
> The issue was found (and formalized into this test) by Christian Reitter from Radically Open Security.

This specific issue is fixed by 5af70f81e57158f37e68dbabec628651e39f0ae3

src/crypto/sym/decryptor.rs
```diff
@@ -328,161 +390,195 @@ where
...
+                        self.finalize_data()?;
+                    }
+                }
+                Err(e) => {
+                    *self = Self::Error;
+                    return Err(e);
+                }
+            },
+            Self::Done { .. } => {}
+            Self::Error => panic!("error state"),
```
lubux:
```suggestion
            Self::Error => return Err(io::Error::other("decryptor is in error state")),
```

tests/security.rs
```diff
@@ -357,3 +359,49 @@ fn oom_signature_1() {
...
+        48, 123, 142, 202, 99, 182, 207, 101, 106, 252, 173, 8, 233, 108, 32, 119, 163, 230, 7, 74,
+        118, 238, 167, 25, 242, 105, 210, 40, 1, 74, 92, 164, 57, 220, 200, 43, 232, 4, 30, 193,
+        225, 240, 5, 244, 112, 26, 194, 157, 76, 187, 216, 231, 67, 242, 16, 223, 91, 151, 160,
+        109, 31, 75, 249, 110, 219, 40, 53, 50, 130, 36, 23, 110,
+    ];
+
+    let key_input = include_str!("draft-bre-openpgp-samples-00/bob.sec.asc");
+    let (decrypt_key, _headers) = SignedSecretKey::from_string(key_input).unwrap();
+
+    let message = Message::from_bytes(Cursor::new(bad_input)).unwrap();
```
lubux: Nit: It is possible to do it without Cursor with `let bad_input: &[u8] = &[ ` as in the other tests above or `&bad_input[..]`

hko-s: Right, thanks!

hko-s:
Note (mostly) to self:

If `decryptor::BUFFER_SIZE` were caller-configurable, users could choose the max message size that gets MDC-checked in one shot, in streaming decryption mode (this setting could also live in `Seipdv1ReadMode`).

src/crypto/sym/decryptor.rs
```diff
@@ -328,161 +390,195 @@ where
...
+                    // MDC is 1 byte packet tag, 1 byte length prefix and 20 bytes SHA1 hash.
+                    let mdc = buffer.split_off(buffer.len() - MDC_LEN);
+
+                    hasher.update(&mdc[..2]);
+
+                    let sha1: [u8; 20] = hasher.finalize().into();
+
+                    if mdc[0] != 0xD3 || // Invalid MDC tag
+                        mdc[1] != 0x14 || // Invalid MDC length
+                        mdc[2..] != sha1[..]
```
lubux: Should we not use constant time comparison here?

hko-s: Good point! I made an attempt to improve this, in 55ec80cb7b60494cf06c6e458158b9f285647da4

src/crypto/sym/decryptor.rs
```diff
@@ -328,161 +393,201 @@ where
...
+                    // The value for the MDC we expect.
+                    let mut expect_mdc = [0; 22];
+                    expect_mdc[0] = 0xD3; // MDC tag
+                    expect_mdc[1] = 0x14; // MDC length
+                    expect_mdc[2..].copy_from_slice(&hash);
+
+                    // constant-time comparison of expected MDC and MDC from message
+                    if expect_mdc.ct_ne(&msg_mdc).into() {
+                        return Err(io::Error::other(Error::MdcError));
+                    }
```
lubux: Nit: We could also compare without a copy, but it is correct. 

```
                    let mdc_ok = msg_mdc[0].ct_eq(&0xD3)
                        & msg_mdc[1].ct_eq(&0x14)
                        & msg_mdc[2..].ct_eq(hash.as_slice());
                    if bool::from(!mdc_ok) {
                        return Err(io::Error::other(Error::MdcError));
                    }
```

hko-s: That is so much nicer. Thanks!

src/composed/message/types.rs
```diff
@@ -523,11 +529,46 @@ impl DecryptionOptions {
...
+    /// In streaming mode, MDC errors may only be returned after the caller has already read
+    /// modified plaintext! If the ciphertext has been altered, some of the released plaintext may
+    /// be incorrect (and possibly attacker-controlled to some degree).
+    /// In addition, in this mode, possibly modified plaintext is read by rPGP's message
+    /// parser before authentication. This may trigger any number of (potentially attacker-caused)
+    /// parsing errors.
+    ///
+    /// If the attacker can observe the resulting modified plaintext, error outputs, or timing
+    /// characteristics, they may be able to mount attacks on the encrypted data.
+    #[must_use]
```
dignifiedquire, line 566: either add `muse_use` to all variants or none

src/crypto/sym/decryptor.rs
```diff
@@ -220,20 +235,62 @@ where
...
+        matches!(
+            self,
+            MaybeProtected::Protected {
+                mode: Seipdv1ReadMode::Streaming,
+                ..
+            }
+        )
+    }
+
+    fn is_sed(&self) -> bool {
```
dignifiedquire: this is a very cryptic function name

src/crypto/sym/decryptor.rs
```diff
@@ -220,20 +235,62 @@ where
...
-        // We use regular sha1 for MDC, not sha1_checked. Collisions are not currently a concern with MDC.
+        /// We use regular sha1 for MDC, not sha1_checked.
+        /// Collisions are not currently a concern with MDC.
         hasher: Sha1,
+
+        /// The read mode for seipdv1 encrypted data.
+        ///
+        /// By default, the data packet is read in one go, so that plaintext is only released
+        /// after the MDC has been checked.
+        mode: Seipdv1ReadMode,
```
dignifiedquire: instead of having a mode that switches, it likely would be better to have three variants of the `MaybeProtected` enum

src/types.rs
```diff
@@ -49,6 +49,41 @@ pub enum EskType {
...
+#[derive(derive_more::Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Seipdv1ReadMode {
+    CheckFirst { max_message_size: usize },
+    Streaming,
+}
+
+/// The maximum message length that we're willing to decrypt in non-streaming SEIPDv1 mode
+const MAX_DEFAULT_UNSTREAMED_MSG_SIZE: usize = 1024 * 1024 * 1024;
+
+impl Default for Seipdv1ReadMode {
```
dignifiedquire, line 79: I don't think we should implement a default for this,

tests/seipdv1.rs
```diff
@@ -111,67 +117,263 @@ pub fn mdc_test_fast_2() {
...
-    let encrypted_data: &[u8] = &{
+    let (encrypted_data, _) = {
         let mut rng = ChaCha8Rng::seed_from_u64(1);
-        make_seipdv1_msg(&mut rng)
+        make_seipdv1_msg(&mut rng, 50_000)
     };
 
-    let total_iterations: u64 = 2_500_000;
-    eprintln!("Running MDC fuzz test with {} iterations", total_iterations);
+    for mode in [Seipdv1ReadMode::default(), Seipdv1ReadMode::Streaming] {
```
dignifiedquire: would be nicer if the test is extracted into a function, and the top level tests is split by mode, much easier debugging

src/types.rs
hko-s: This type is used in `DecryptionOptions`, which in turn also implements `Default`.
I'm not convinced that we can reasonably avoid implementing a default for both of these types.

(And from a more opinionated perspective: It seems good to me to make the more defensive behavior for SEIPDv1 the default for unsuspecting users.)

src/composed/message/types.rs
hko-s: Oops, thanks! It was an oversight that `enable_draft_forwarding` doesn't have it. Fixed.

src/crypto/sym/decryptor.rs
hko-s: I like the idea of giving `MaybeProtected` three variants (done).

But I think we independently still need a type that passes the caller's preference into the machinery that eventually creates `MaybeProtected` objects :thinking:

tests/seipdv1.rs
hko-s: Good point. I've refactored two tests in this file as you suggested.

dignifiedquire (approved):

PR #667, opened 2025-11-22
support for draft-ietf-openpgp-persistent-symmetric-keys-03

hko-s:
This is an implementation of [draft-ietf-openpgp-persistent-symmetric-keys-03](https://www.ietf.org/archive/id/draft-ietf-openpgp-persistent-symmetric-keys-03.html).

The implementation is completely feature gated, no aspects of this PR are visible when the feature is disabled.

Trait-based operations on persistent symmetric key packets (e.g. with the message builder/reader abstrations) are mediated via wrapper types:

- `PersistentSymmetricEncryptionKey`,
- `PersistentSymmetricVerifyingKey`,
- `PersistentSymmetricSigningKey`

These facilitate setting the extra parameters that these symmetric operations require in addition to their asymmetric equivalents.
Only decryption does not require special handling - it is implemented directly on `PersistentSymmetricKey`.

Note that `PersistentSymmetricSigningKey` uses interior mutability in order to use an Rng in the `sign` operation.


src/types/params/plain_secret.rs
```diff
@@ -405,6 +413,30 @@ impl PlainSecretParams {
...
+            ) => {
+                let PublicParams::AEAD(pub_param) = pub_params else {
+                    bail!("inconsistent key state");
+                };
+
+                priv_key.decrypt(aead_key::EncryptionFields {
+                    data: encrypted,
+                    aead: *aead,
+                    // PKESK version
+                    version: match typ {
```
dignifiedquire, line 456: maybe use an enum here, instead of numbers?

src/crypto/aead_key.rs
```diff
@@ -0,0 +1,218 @@
...
+    types::{SignatureBytes, Tag},
+};
+
+/// Secret key for AEAD persistent symmetric keys
+#[derive(Clone, PartialEq, Eq, ZeroizeOnDrop, derive_more::Debug)]
+#[cfg_attr(test, derive(proptest_derive::Arbitrary))]
+pub struct SecretKey {
+    #[debug("..")]
+    #[cfg_attr(test, proptest(strategy = "tests::key_gen()"))]
+    pub(crate) key: Vec<u8>, // sized to match the sym_alg
```
dignifiedquire: `Box<[u8]>` if this is fixed size?

src/crypto/aead_key.rs
```diff
@@ -0,0 +1,218 @@
...
+    /// Key and IV derivation
+    /// <https://twisstle.gitlab.io/openpgp-persistent-symmetric-keys/#name-key-and-iv-derivation>
+    ///
+    /// Returns:
+    /// - M bits of key, matching the size of SymmetricKeyAlgorithm,
+    /// - N bit of IV, matching the nonce size of AeadAlgorithm
+    pub(crate) fn derive(
+        persistent_key: &[u8],
+        salt: &[u8; 32],
+        info: (Tag, u8, AeadAlgorithm, SymmetricKeyAlgorithm),
```
dignifiedquire: use a type?

src/crypto/aead_key.rs
```diff
@@ -0,0 +1,218 @@
...
+
+        // The signature consists of this series of values:
+        //
+        // A 1-octet AEAD algorithm (see section 9.6 of [RFC9580]).
+        let aead = AeadAlgorithm::Ocb; // FIXME: should be a parameter
+
+        // 32 octets of salt.
+        // The salt is used to derive the message authentication key and MUST be securely generated
+        // (see section 13.10 of [RFC9580]).
+        let mut rng = thread_rng(); // FIXME: should be a parameter?
```
dignifiedquire: yes, this must be a parameter

src/packet/key/symmetric.rs
```diff
@@ -0,0 +1,554 @@
...
+        let packet_header = PacketHeader::new_fixed(Tag::PersistentSymmetricKey, len.try_into()?);
+
+        Ok(Self {
+            packet_header,
+            details,
+            secret_params,
+        })
+    }
+
+    /// Parses a `SecretKey` packet from the given buffer.
```
dignifiedquire: wrong comment

src/packet/key/symmetric.rs
```diff
@@ -0,0 +1,554 @@
...
+
+    pub fn as_unlockable<'a>(
+        &'a self,
+        key_pw: &'a Password,
+    ) -> UnlockablePersistentSymmetricKey<'a> {
+        UnlockablePersistentSymmetricKey { psk: self, key_pw }
+    }
+}
+
+pub struct UnlockablePersistentSymmetricKey<'a> {
```
dignifiedquire: this concept is new, why do we need it?

src/packet/key/symmetric.rs
```diff
@@ -0,0 +1,554 @@
...
+impl<'a> Debug for UnlockablePersistentSymmetricKey<'a> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        self.psk.fmt(f)
+    }
+}
+
+impl<'a> VerifyingKey for UnlockablePersistentSymmetricKey<'a> {
+    fn verify(
+        &self,
+        _hash: HashAlgorithm,
```
dignifiedquire: should at least check it in some form

src/types/signature.rs
```diff
@@ -22,6 +23,11 @@ pub enum SignatureBytes {
     ///
     /// This format was introduced in RFC 9580 and is currently only used for Ed25519 and Ed448.
     Native(#[debug("{}", hex::encode(_0))] Bytes),
+
+    /// A cryptographic "signature" from draft-ietf-openpgp-persistent-symmetric-keys
+    ///
+    /// See https://twisstle.gitlab.io/openpgp-persistent-symmetric-keys/#name-algorithm-specific-fields-for
+    PersistentSymmetric(AeadAlgorithm, [u8; 32], Vec<u8>),
```
dignifiedquire: lets use named fields here please,

src/packet/key/symmetric.rs
hko-s: My idea for that type was that it allows passing a `Password` into a regular `encrypt` and `verify` call like in https://github.com/rpgp/rpgp/pull/667/commits/d09cd49174bb2b22c799649d939332a01784fd86.

That seems like a reasonably nice way to hide away the need for a password from the rest of the code.

hko-s: In other words, this new type combines a (potentially password-locked) persistent symmetric key with a password.

The combined type can then be used in APIs that expect to operate on public key packets (which can not be locked, and don't need a password).

src/packet/key/symmetric.rs
hko-s: good point, adjusted

src/crypto/aead_key.rs
```diff
@@ -0,0 +1,248 @@
...
+    types::{PkeskVersion, SignatureBytes, Tag},
+};
+
+/// Secret key for AEAD persistent symmetric keys
+#[derive(Clone, PartialEq, Eq, ZeroizeOnDrop, derive_more::Debug)]
+#[cfg_attr(test, derive(proptest_derive::Arbitrary))]
+pub struct SecretKey {
+    #[debug("..")]
+    #[cfg_attr(test, proptest(strategy = "tests::key_gen()"))]
+    pub(crate) key: Box<[u8]>, // must be sized to match the sym_alg
```
lubux, line 26: Zeroizing?

src/crypto/aead_key.rs
```diff
@@ -0,0 +1,248 @@
...
+}
+
+pub struct EncryptionFields<'a> {
+    pub data: Bytes,
+    pub sym_alg: SymmetricKeyAlgorithm,
+    pub aead: AeadAlgorithm,
+    pub version: PkeskVersion,
+    pub salt: &'a [u8; 32],
```
lubux, line 34: Is it worth the complexity to have a reference instead owned type here?

src/crypto/aead_key.rs
```diff
@@ -0,0 +1,248 @@
...
+            value.version,
+            value.sym_alg.into(),
+            value.aead.into(),
+        ]
+    }
+}
+
+impl SecretKey {
+    /// Signing operation for persistent symmetric keys that exposes the full algorithmic flexibility
+    pub fn sign_persistent_symmetric<RNG: Rng + CryptoRng>(
```
lubux: `compute_and_wrap_persistent_mac` is more precise name here?

I realize the standard is somewhat vague in its use of the term signature, but in my opinion what is actually produced is not a signature, but rather a message authentication code (MAC).

src/crypto/aead_key.rs
```diff
@@ -0,0 +1,248 @@
...
+        let mut salt = [0; 32];
+        rng.fill(&mut salt);
+
+        let signature = self.calculate_signature(version, sym_alg, aead, &salt, digest)?;
+        let tag = signature.to_vec().into();
+
+        Ok(SignatureBytes::PersistentSymmetric { aead, salt, tag })
+    }
+
+    pub(crate) fn calculate_signature(
```
lubux: `compute_persistent_mac`?

src/packet/key/symmetric.rs
```diff
@@ -0,0 +1,615 @@
...
+                bail!("Unsupported secret parameters for persistent symmetric key: {sec_params:?}");
+            };
+
+            let version = SignatureVersion::V6; // FIXME: should not be fixed
+
+            // "buf" is the newly calculated authentication tag
+            let buf = secret.calculate_signature(version, public.sym_alg, *aead, salt, data)?;
+
+            // check if the stored and calculated authentication tags match
+            if buf != **tag {
```
lubux: Should be constant time comparison (crypto subtle)

src/packet/key/symmetric.rs
```diff
@@ -0,0 +1,615 @@
...
+        mut rng: R,
+        plain: &[u8],
+        typ: EskType,
+    ) -> crate::errors::Result<PkeskBytes> {
+        ensure!(
+            matches!(typ, EskType::V6),
+            "only v6 ESK supported right now"
+        );
+
+        let aead = AeadAlgorithm::Ocb; // FIXME: parameter
```
lubux: Do you know why a `Persistent Symmetric Key` only encodes the symmetric but not the aead algorithm?

src/packet/key/symmetric.rs
```diff
@@ -0,0 +1,615 @@
...
+
+                _ => bail!("Unsupported key version for persistent symmetric key signing"),
+            };
+
+            // This trait interface doesn't allow exposing the full flexibility of persistent symmetric
+            // key signatures, so we're using fixed values for `rng` and `aead` here.
+
+            // TODO: expose signing with full flexibility as a separate fn on this type?
+
+            let rng = thread_rng();
```
lubux: Hmm seems a bit odd to me that singing does not take an rng? Do we use deterministic ECDSA?

src/crypto/aead_key.rs
```diff
@@ -0,0 +1,248 @@
...
+        // M + N bits are derived using HKDF.
+        // The left-most M bits are used as symmetric algorithm key, the remaining N bits are
+        // used as initialization vector.
+        let mut output = Zeroizing::new(vec![0u8; key_size + nonce_size]);
+
+        let info_parameter: [u8; 4] = info.into();
+        hk.expand(&info_parameter, &mut output)
+            .expect("expand size is < 255 * HashLength");
+
+        let key: Box<[u8]> = output[0..key_size].into();
```
lubux: Could also use the [split_off](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.split_off) function and return vecs

src/packet/key/symmetric.rs
hko-s: My mental model is that the PSK draft handles AEAD algorithm choice analogously to how message encryption works in PGP (where the AEAD mode is not specified by the involved key packets. Instead, it's a degree of freedom that can be decided separately for each message).

src/crypto/aead_key.rs
hko-s: There is a `derive` for `ZeroizeOnDrop`. I think that should be a robust way to handle zeroizing for the key material?

(Generally, rPGP defaults to the `derive` approach, but in some places where it can't be applied, zeroize is used more "manually".)

src/crypto/aead_key.rs
hko-s: I defaulted to borrowing here, because the other `EncryptionFields` also use that pattern. (And at least for the PQC key material, borrowing is probably actually worth it.)

So I'm leaning towards keeping this as is, because it's following an existing pattern. Otherwise I'd tend to agree that it's not worth it.

src/crypto/aead_key.rs
hko-s: yes! i like giving these functions  more "technical" names.

src/packet/key/symmetric.rs
hko-s: After our discussions around this, I started https://github.com/rpgp/rpgp/pull/742/ to explore passing an RNG into signing operations

src/crypto/aead_key.rs
```diff
@@ -0,0 +1,248 @@
...
+        let iv = output[key_size..].into();
```
pcwizz: There is a zeroization gap for the IV here. The key is however properly covered.

src/packet/signature/de.rs
```diff
@@ -305,6 +305,14 @@ fn subpacket<B: BufRead>(
 
 fn actual_signature<B: BufRead>(typ: &PublicKeyAlgorithm, mut i: B) -> Result<SignatureBytes> {
     match typ {
+        #[cfg(feature = "draft-ietf-openpgp-persistent-symmetric-keys")]
+        PublicKeyAlgorithm::AEAD => {
+            let aead = i.read_u8()?.into();
+            let salt = i.read_arr()?;
+            let tag = i.rest()?.to_vec().into();
+            Ok(SignatureBytes::PersistentSymmetric { aead, salt, tag })
```
pcwizz, line 340: Why not validate the tag length here already?

src/packet/key/symmetric.rs
```diff
@@ -0,0 +1,803 @@
...
+
+// --- signing key
+
+#[derive(derive_more::Debug)]
+pub struct PersistentSymmetricSigningKey<R: CryptoRng + Rng> {
+    psk: PersistentSymmetricKey,
+
+    // NOTE: Interior mutability to work around read-only access via SigningKey trait
+    #[debug("Rng")]
+    rng: RefCell<R>,
```
lubux, line 420: RefCell makes PersistentSymmetricSigningKey lose the Sync property. Are we fine with that?

src/composed/signed_key/shared.rs
```diff
@@ -356,17 +370,29 @@ impl PublicOrSecret {
...
+
+            #[cfg(feature = "draft-ietf-openpgp-persistent-symmetric-keys-03")]
+            PublicOrSecret::PersistentSymmetric(_) => false,
         }
     }
 
+    /// true for Secret, false for Public
+    ///
+    /// (true for PersistentSymmetric, if feature draft-ietf-openpgp-persistent-symmetric-keys-03 is enabled)
     pub fn is_secret(&self) -> bool {
```
lubux, line 389: maybe an additional feature gated `is_persistent_symmetric()` would be great? As it cannot be converted into a SignedSecretKey?

tests/persistent_symmetric.rs
```diff
@@ -0,0 +1,124 @@
...
+
+        let sk = psk
+            .decrypt(&Password::empty(), pkesk.values().unwrap(), EskType::V3_4)
+            .unwrap()
+            .unwrap();
+
+        ring.session_keys.push(sk);
+    }
+
+    let (mut msg, _res) = msg.decrypt_the_ring(ring, false).unwrap();
```
lubux, line 61: Should "the ring" not offer the option to decrypt with a PSK?

src/packet/key/public.rs
lubux, line 372: There should be checks here to not allow algorithm AEAD for ordinary public keys?

src/packet/key/symmetric.rs
```diff
@@ -0,0 +1,803 @@
...
+            // securely generated (see section 13.10 of [RFC9580]).
+            let mut salt: [u8; 32] = [0; 32];
+            rng.fill(&mut salt);
+
+            // A symmetric key encryption of the plaintext value described in section 5.1 of [RFC9580],
+            // performed with the key-encryption key and IV computed as described in Section 7.4,
+            // using the symmetric-key cipher of the key and the indicated AEAD mode, with as
+            // additional data the empty string; including the authentication tag.
+
+            let version = version.into();
```
lubux: This should be the PKESK version and not the key version?

src/packet/key/symmetric.rs
```diff
@@ -0,0 +1,803 @@
...
+        &self.details
+    }
+
+    /// Remove the password protection of the private key material in this key packet.
+    /// This permanently "unlocks" the secret key material.
+    ///
+    /// If the Secret Key material in the packet is not locked, it is left unchanged.
+    ///
+    /// The current locking password for this key must be provided in `password`.
+    pub fn remove_password(&mut self, password: &Password) -> Result<()> {
```
lubux, line 131: Should this not check:

> "AEAD encryption (S2K usage octet 253) MUST be used... Implementations MUST NOT
> decrypt symmetric key material in a Persistent Symmetric Key Packet that was encrypted using a
> different method."

src/crypto/aead_key.rs
```diff
@@ -0,0 +1,230 @@
...
+        };
+
+        let (key, iv) = Self::derive_key_iv(&self.key, data.salt, info);
+
+        let mut buf = data.data.into();
+
+        data.aead
+            .decrypt_in_place(&data.sym_alg, &key, &iv, &[], &mut buf)?;
+
+        Ok(buf.to_vec().into())
```
lubux:
```suggestion
       Ok(Zeroizing::new(buf.into()))
```
?

src/types/params/public.rs
```diff
@@ -9,6 +9,7 @@ use crate::{
     ser::Serialize,
 };
 
+mod aead;
```
lubux, line 13: No feature gate?

src/types/pkesk.rs
```diff
@@ -16,6 +16,11 @@ use crate::{
 /// Values comprising a Public Key Encrypted Session Key
 #[derive(Clone, derive_more::Debug, Eq, PartialEq)]
 pub enum PkeskBytes {
+    Aead {
```
lubux, line 20: No feature gate?

src/types/signature.rs
```diff
@@ -22,6 +23,15 @@ pub enum SignatureBytes {
     ///
     /// This format was introduced in RFC 9580 and is currently only used for Ed25519 and Ed448.
     Native(#[debug("{}", hex::encode(_0))] Bytes),
+
+    /// A cryptographic "signature" from draft-ietf-openpgp-persistent-symmetric-keys-03
+    ///
+    /// See <https://www.ietf.org/archive/id/draft-ietf-openpgp-persistent-symmetric-keys-03.html#name-algorithm-specific-fields-for>
+    PersistentSymmetric {
```
lubux, line 30: No feature gate?

src/crypto/public_key.rs
```diff
@@ -5,6 +5,14 @@ use num_enum::{FromPrimitive, IntoPrimitive};
 #[repr(u8)]
 #[non_exhaustive]
 pub enum PublicKeyAlgorithm {
+    /// AEAD (only for use in a "Persistent Symmetric Key Packet"!)
+    /// Ref <https://www.ietf.org/archive/id/draft-ietf-openpgp-persistent-symmetric-keys-03.html#name-persistent-symmetric-key-al>
+    ///
+    /// CAUTION: This is not a public key algorithm, it may not be used with regular key packets.
+    /// Instead, this algorithm may be used with [`PersistentSymmetricKey`] packets.
+    #[cfg(feature = "draft-ietf-openpgp-persistent-symmetric-keys-03")]
+    AEAD = 0,
```
lubux, line 14: I'm still uneasy that the standard picked 0, as it is the implicit default.

src/packet/key/symmetric.rs
hko-s: We discussed this some more, and considered the tradeoff between `RefCell` (not sync) and `Mutex` (sync, but higher overhead). We leaned towards keeping this as `RefCell` while no one has a specific requirement that would justify the additional overhead of `Mutex`.

src/composed/signed_key/shared.rs
hko-s: Good point. I changed `is_secret` to false for PSK, added `is_persistent_symmetric` and an additional `TryFrom` implementation.

src/packet/key/symmetric.rs
hko-s: Ouch, very good catch! Fixing that makes the encryption API slightly more unwieldy, but I suppose that's unavoidable.

src/packet/key/public.rs
hko-s: Given the current implementation, that is not easily possible, because `pgp::packet::key::symmetric::PersistentSymmetricKey` uses a `PublicKey` as a member.
This  is analogous to how `pgp::packet::key::secret::SecretKey` is constructed.

Now I wonder if it's worth attempting to change that. But then, the risk of accidentally constructing a PublicKey object with the AEAD algorithm seems very small.

Such an object would refuse all cryptographic operations. So it might be confusing/annoying to make such an object, but I think not dangerous.

tests/persistent_symmetric.rs
hko-s: Good point, I added code for that

src/crypto/public_key.rs
hko-s: I'm also vaguely uneasy about that code point, but I can't think of a serious argument why it's dangerous.

Putting a symmetric mechanism into the asymmetric registry is weird, of course - but I know of no obviously preferable alternatives to the current design. Given that, using an "unusual" code point (like zero) does seem nicer than using a regular-looking code point (like "45", which would be the next unassigned one, if the current set of PQC algorithms settle as planned).

(But then, I don't think that using 45 would be prohibitive either. Nothing particularly terrible should happen, if implementations encounter that, e.g. in a PKESK, and don't know how to deal with it.)

PR #668, opened 2025-11-23
refactor!: introduce EncryptionKey trait

hko-s:


src/composed/message/builder.rs
```diff
@@ -433,10 +434,13 @@ impl<R: Read> Builder<'_, R, EncryptionSeipdV1> {
     /// [PK ESK]: https://www.rfc-editor.org/rfc/rfc9580#name-public-key-encrypted-sessio
     /// # References
     /// [RFC 9580 &sect; 5.1 - Public Key Encrypted Session Key Packet](https://www.rfc-editor.org/rfc/rfc9580#section-5.1)
-    pub fn encrypt_to_key<RAND, K>(&mut self, mut rng: RAND, pkey: &K) -> Result<&mut Self>
+    pub fn encrypt_to_key<RAND>(
+        &mut self,
+        mut rng: RAND,
+        pkey: &(impl PublicKeyTrait + Encryptor),
```
dignifiedquire: please stick with generics instead of `impl`

src/composed/signed_key/public.rs
```diff
@@ -162,14 +163,16 @@ impl SignedPublicKey {
...
-    pub fn encrypt<R: Rng + CryptoRng>(
+impl Encryptor for SignedPublicKey {
+    fn encrypt<R: Rng + CryptoRng>(
         &self,
         rng: R,
         plain: &[u8],
         typ: EskType,
     ) -> Result<PkeskBytes> {
-        self.primary_key.encrypt(rng, plain, typ)
+        (&self.primary_key).encrypt(rng, plain, typ)
```
dignifiedquire: this seems odd

dignifiedquire:
we will need to rethink naming, it is very confusing now, to have `Encryptor` and `Public` and `Secret` keys

hko-s:
> we will need to rethink naming, it is very confusing now, to have `Encryptor` and `Public` and `Secret` keys

I'm open to renaming, but to my mind, with #667, the naming in this PR makes some sense.
(With persistent symmetric keys, encryption becomes partly independent of public-key-ness, see [here](https://github.com/hko-s/rpgp/blob/PSK/src/packet/key/symmetric.rs#L163).)

hko-s:
> > we will need to rethink naming, it is very confusing now, to have `Encryptor` and `Public` and `Secret` keys

Ah, and we already have [`Decryptor`](https://docs.rs/pgp/latest/pgp/crypto/trait.Decryptor.html) and [`Signer`](https://docs.rs/pgp/latest/pgp/crypto/trait.Signer.html), so I was trying to extend the pattern analogously.

src/composed/signed_key/public.rs
hko-s: indeed ... at some point the compiler wanted me to do that. now it's happy without. yay!

src/packet/key/secret.rs
```diff
@@ -494,7 +495,7 @@ impl SecretSubkey {
         plain: &[u8],
         typ: EskType,
     ) -> Result<PkeskBytes> {
-        encrypt(&self.details, rng, plain, typ)
+        self.details.encrypt(rng, plain, typ)
```
dignifiedquire: why is this not using the trait?

src/packet/key/secret.rs
```diff
@@ -402,7 +403,7 @@ impl SecretKey {
...
```
dignifiedquire: same as below, couldn't we implement `Encryptor` for `SecretKey`?

src/crypto.rs
```diff
@@ -56,3 +56,13 @@ pub trait Decryptor {
 pub trait Signer {
     fn sign(&self, hash: HashAlgorithm, digest: &[u8]) -> crate::errors::Result<SignatureBytes>;
 }
+
+/// Describes keys that can encrypt data.
+pub trait Encryptor {
```
dignifiedquire: can `Encryptor` ever be used without `PublicKeyTrait`? should it require `PublicKeyTrait`?

hko-s: I suspect it can never be used without `PublicKeyTrait` ... if we don't run into dynnable trouble by requiring that it must be a superset of `PublicKeyTrait`, it could be good to require it :thinking:

src/packet/key/secret.rs
hko-s: Oops. I missed those two, and a third place. Pushed one more commit.

src/crypto.rs
hko-s: Naively changing the trait to `pub trait Encryptor: PublicKeyTrait` causes complaints such as

```
error[E0277]: the trait bound `SecretKey: PublicKeyTrait` is not satisfied
   --> src/packet/key/secret.rs:401:20
    |
401 | impl Encryptor for SecretKey {
    |                    ^^^^^^^^^ unsatisfied trait bound
    |
help: the trait `types::key_traits::PublicKeyTrait` is not implemented for `packet::key::secret::SecretKey`
   --> src/packet/key/secret.rs:29:1
```

hko-s: And (mostly as an aside, in this context) rather surprisingly:

```
the trait bound `composed::key::public::PublicKey: key_traits::PublicKeyTrait` is not satisfied
```

... it seems that one of the public key types doesn't implement `PublicKeyTrait` :dizzy_face:

hko-s: Maybe this means that the secret key types shouldn't implement encryption at all, and users should be required to obtain a public view of them in order to be able to encrypt to them?

src/composed/signed_key/secret.rs
```diff
@@ -218,6 +209,17 @@ impl SignedSecretKey {
...
 }
 
+impl Encryptor for SignedSecretKey {
+    fn encrypt<R: Rng + CryptoRng>(
+        &self,
+        rng: R,
+        plain: &[u8],
+        typ: EskType,
+    ) -> Result<PkeskBytes> {
+        self.primary_key.encrypt(rng, plain, typ)
```
hko-s: Oh no ... this function is hardcoded to encrypt to the primary key? That doesn't seem right :confused:

hko-s: Hm. I guess if it's documented that this is what the function does, then it's reasonable. For a second I was surprised, because decryption is very asymmetrical: it works on picking the correct primary or subkey.

dignifiedquire:
> Maybe this means that the secret key types shouldn't implement encryption at all, and users should be required to obtain a public view of them in order to be able to encrypt to them?

Honestly I would love to remove all encryption capabilities from secret keys

benches/benchmarks/message.rs
```diff
@@ -171,7 +171,7 @@ fn bench_message(c: &mut Criterion) {
                         let mut builder = MessageBuilder::from_reader("", &bytes[..])
                             .seipd_v1(&mut rng, SymmetricKeyAlgorithm::AES128);
                         builder
-                            .encrypt_to_key(&mut rng, &signed_key.secret_subkeys[0].public_key())
+                            .encrypt_to_key(&mut rng, signed_key.secret_subkeys[0].public_key())
```
dignifiedquire: why does this change?

src/packet/key/public.rs
```diff
@@ -419,168 +423,171 @@ impl PubKeyInner {
...
-                    // (See https://www.rfoc-editor.org/rfc/rfc9580.html#section-11.5.1-2)
-                    let curve = params.curve();
-                    match params {
-                        EcdhPublicParams::Curve25519 { hash, alg_sym, .. }
-                        | EcdhPublicParams::P256 { hash, alg_sym, .. }
-                        | EcdhPublicParams::P521 { hash, alg_sym, .. }
-                        | EcdhPublicParams::P384 { hash, alg_sym, .. } => {
-                            if curve.hash_algo()? != *hash || curve.sym_algo()? != *alg_sym {
-                                bail!("Unsupported KDF/KEK parameters for {:?} and KeyVersion::V6: {:?}, {:?}", curve, hash, alg_sym);
+impl EncryptionKey for PubKeyInner {
```
dignifiedquire: I wonder if we want to do this, as this technically changes the visibillity of the function

src/types/key_traits.rs
```diff
@@ -165,3 +167,14 @@ pub trait SecretKeyTrait: KeyDetails + std::fmt::Debug {
     /// when using this as a signer
     fn hash_alg(&self) -> HashAlgorithm;
 }
+
+/// Describes keys that can encrypt plain data (i.e. a session key) into data for a
+/// [PKESK](https://www.rfc-editor.org/rfc/rfc9580#name-public-key-encrypted-sessio).
+pub trait EncryptionKey: PublicKeyTrait {
```
dignifiedquire, line 175: please also add `impl<T: EncryptionKey> EncryptionKey for &T` the same we have for the other traits

benches/benchmarks/message.rs
hko-s: Honestly, I was also confused that the compiler didn't like the previous syntax. Without changing this, it complains:

```
error[E0277]: the trait bound `&...: EncryptionKey` is not satisfied
   --> benches/benchmarks/message.rs:174:55
    |
174 | ...  .encrypt_to_key(&mut rng, &signed_key.secret_subkeys[0].public_key())
    |       --------------           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ unsatisfied trait bound
    |       |
    |       required by a bound introduced by this call
    |
    = help: the trait `pgp::types::EncryptionKey` is not implemented for `&pgp::packet::PublicSubkey`
note: required by a bound in `pgp::composed::MessageBuilder::<'_, R, pgp::composed::EncryptionSeipdV1>::encrypt_to_key`
   --> /home/heiko/src/rpgp/src/composed/message/builder.rs:439:12
    |
436 |     pub fn encrypt_to_key<RAND, E>(&mut se...
    |            -------------- required by a bound in this associated function
...
439 |         E: EncryptionKey,
    |            ^^^^^^^^^^^^^ required by this bound in `MessageBuilder::<'_, R, EncryptionSeipdV1>::encrypt_to_key`
    = note: the full name for the type has been written to '/home/heiko/src/rpgp/target/debug/deps/benches_main-c00e5d0dfa7e80f5.long-type-17560952904533768340.txt'
    = note: consider using `--verbose` to print the full type name to the console
help: consider removing the leading `&`-reference
    |
174 -                             .encrypt_to_key(&mut rng, &signed_key.secret_subkeys[0].public_key())
174 +                             .encrypt_to_key(&mut rng, signed_key.secret_subkeys[0].public_key())
    |

For more information about this error, try `rustc --explain E0277`.
```

I don't understand what's going on there, but it seems that the function doesn't take ownership of the key now (which would be bad).

src/types/key_traits.rs
hko-s: Oh. That will probably resolve the comment above :sweat_smile:

src/packet/key/public.rs
hko-s: Right, there's no strong need to implement `EncryptionKey` for `PubKeyInner`. I've removed it.

dignifiedquire (approved):

PR #671, opened 2025-11-26
refactor!: simplify key generation

dignifiedquire:
Tries out to remove `composed::key::{Public|Secret}Key` from the public API

### DIFF
--- a/tests/key_test.rs
+++ b/tests/key_test.rs
@@ -1294,8 +1294,7 @@ fn test_locked_key() {
         HashAlgorithm::Sha256,
         b"payload".as_slice(),
     )
-    .err()
-    .expect("can't sign with wrong password");
+    .expect_err("can't sign with wrong password");
 
     assert!(matches!(res, Error::InvalidInput { .. }));
 


tests/key_test.rs
```diff
@@ -1275,27 +1273,27 @@ fn test_invalid() {
...
-        .err()
-        .unwrap();
-
-    assert!(matches!(res, pgp::errors::Error::InvalidInput { .. }));
-    let _signed_key = unsigned_pubkey
-        .sign(&mut rng, &*key, &*key.public_key(), &"123".into())
-        .unwrap();
+    let _signed_pubkey = key.signed_public_key();
+
+    // // Incorrect password results in InvalidInput error.
```
dignifiedquire: @hko-s any ideas on how to replicate this test?

src/composed/key/secret.rs
```diff
@@ -2,51 +2,38 @@ use aes_gcm::aead::rand_core::CryptoRng;
...
+    composed::{KeyDetails, SignedPublicSubKey, SignedSecretKey, SignedSecretSubKey},
     errors::Result,
     packet::{self, KeyFlags, Signature},
-    ser::Serialize,
-    types::{Password, PublicKeyTrait, SecretKeyTrait},
+    types::Password,
 };
 
-/// User facing interface to work with the components of a "Transferable Secret Key (TSK)"
+/// Interanal building block to represent the components of a "Transferable Secret Key (TSK)"
```
hko-s:
```suggestion
/// Internal building block to represent the components of a "Transferable Secret Key (TSK)"
```

tests/key_test.rs
hko-s: It seems that this test just deals with password unlocking. So I think we can just make detached data signatures instead, that should produce the same error?

hko-s: I pushed a commit that does that

PR #672, opened 2025-11-29
refactor!: New structure for key traits

dignifiedquire:
- `PublicKeyTrait` -> `VerifyingKey`
- `SecretKeyTrait` -> `SigningKey`
- moved some more things onto `KeyDetails`


hko-s:
I like this set of renames

PR #673, opened 2025-11-29
refactor!: replace chrono::DateTime with custom Timestamp

dignifiedquire:
- replaces usage of `chrono` with a custom `Timestamp` type.
- replaces usage of `std::time::Duration` with a custom `Duration` type.


hko-s (approved):
I like this change!

PR #675, opened 2025-11-30
refactor!: verify -> verify_bindings

dignifiedquire:
For SignedPublicKey, SignedSecretKey, SignedUser, SignedUserAttribute verify is now called verify_bindings to clarify the usage of the methods

### DIFF
--- a/src/composed/signed_key/secret.rs
+++ b/src/composed/signed_key/secret.rs
@@ -151,7 +151,7 @@ impl SignedSecretKey {
 
     /// Drops the secret key material in both the primary key and all secret subkeys.
     /// All other components of the key remain as they are.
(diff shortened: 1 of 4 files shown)


src/composed/signed_key/secret.rs
```diff
@@ -160,7 +151,7 @@ impl SignedSecretKey {
 
     /// Drops the secret key material in both the primary key and all secret subkeys.
     /// All other components of the key remain as they are.
-    pub fn signed_public_key(&self) -> SignedPublicKey {
+    pub fn as_public_key(&self) -> SignedPublicKey {
```
hko-s: Should this be named `to_public_key` to signal that it's an expensive cloney operation?

dignifiedquire: fair

PR #676, opened 2025-12-01
feat: add helper to use multiple secret keys for decryption

ntrinquier:
`Message::decrypt` currently supports passing only one secret key (and password).

This change makes it possible to pass multiple secret keys to decrypt, which is a convenient way to automatically select the appropriate secret key when we do not know which public key was used to encrypt (which is my case: I can receive messages encrypted with different keys, but I do not know which one exactly).

Exposing this capability should also come at no maintenance cost, since the internal data structure (`TheRing`) already supports selecting the correct key and password.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1772,9 +1772,9 @@ dependencies = [
 
 [[package]]
 name = "rsa"
-version = "0.9.9"
+version = "0.9.10"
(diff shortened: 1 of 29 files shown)


dignifiedquire (approved):
thanks

PR #678, opened 2025-12-03
Add back capability to build a SignedPublicKey from a key held externally

baloo:
This bring back an API to create a SignedPublicKey from a key held in an HSM.
This capability got removed in #671.

### DIFF
--- a/tests/hsm-test.rs
+++ b/tests/hsm-test.rs
@@ -4,17 +4,19 @@ use std::{fmt::Debug, fs::File};
 use p256::pkcs8::DecodePrivateKey;
 use pgp::{
     adapter::{EcdsaSigner, RsaSigner},
-    composed::{self, Esk, Message, SignedPublicKey, SignedSecretKey},
+    composed::{self, Esk, Message, SignedPublicKey, SignedPublicSubKey, SignedSecretKey},
     crypto::{
         checksum, ecc_curve::ECCCurve, hash::HashAlgorithm, public_key::PublicKeyAlgorithm,
         sym::SymmetricKeyAlgorithm,
     },
-    packet::{self, PubKeyInner, PublicKey, SignatureConfig},
+    packet::{self, PubKeyInner, PublicKey, SecretSubkey, SignatureConfig},
     types::{
         EcdhPublicParams, Fingerprint, KeyDetails, KeyId, KeyVersion, Mpi, PacketHeaderVersion,
         Password, PkeskBytes, PublicParams, SignatureBytes, SigningKey, Timestamp, VerifyingKey,
     },
 };
+use rand::SeedableRng;
+use rand_chacha::ChaCha8Rng;
 
 #[derive(Debug, Clone)]
 pub struct FakeHsm {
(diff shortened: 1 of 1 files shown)


baloo:
This isn't exactly the best API possible, but it was the best I could come up with.

hko-s:
Thanks for bearing with our trait restructuring activities! :smile: ... I'm going to do some more trait-shuffling over the next days, and will a look at this PR once `main` is settled, in terms of traits

baloo:
Nothing extremely urgent! This is probably something I should have put in the tests along with the rustcrypto adapters when we added that.

If you have an alternative, I'd be happy to test it, please ping me whenever :)

hko-s:
So, mostly as a note to self (this thought could be entirely nonsensical) I kind of suspect that we'll just want to implement the new `pgp::types::key_traits::SigningKey` here, instead.

baloo:
Yeah, that is implemented in the adapter:
https://github.com/rpgp/rpgp/blob/9221ea1361c9ef8e899a7f613a294b9da0d8f3ec/src/adapter/ecdsa.rs#L127-L152

The problem was more to assemble and sign a PGP public key with a private key held externally.
Disclaimer: Our use-case doesn't use subkeys because ... we have a very narrow use-case (PGP signature for RPM packages) which doesn't really require that (just as easy to rotate the whole keyring). All that to say, I didn't focus on that.

hko-s:
> ssemble and sign a PGP public key with a private key held externally.
> Disclaimer: Our use-case doesn't use subkeys because ... we have a very narrow use-case (PGP signature for RPM packages) which doesn't really require that

Ah. I see! I'll look at this in a while, and will report back here.

I guess we should resolve this before the next release. So I even have a deadline :sweat_smile:

baloo:
hopefully we'll get a stable release for rustcrypto by 0.19!

hko-s:
> hopefully we'll get a stable release for rustcrypto by 0.19!

you mean of the rustcrypto++ versions? :scream: (... i was hoping we'd release rpgp 0.19 in the next few weeks)

hko-s:
This seems like a nice and clean solution, thanks!

I think one more capability that we lost in #671 is a convenient way to attach an "embedded back signature" (i.e. a ["primary key binding signature"](ttps://www.rfc-editor.org/rfc/rfc9580.html#sigtype-primary-binding) when binding a signing-capable subkey to a primary).

This is mostly a note-to-self, to tell myself to think about re-adding handling of embedded back-sigs in this PR.

hko-s:
I've finally sat down to see if the available APIs are sufficient to bind a signing subkey in an "HSM primary" context, and it is. It's not super pretty, but maybe it's good enough.

My suggested additions to this PR are in https://github.com/hko-s/rpgp/commits/heiko/hsm-signedpubkey/ (docs/naming, one more test, and a guardrail to avoid accidentally using an unrelated signing key).

With my additions, I'd say this PR is good to merge. The API can apparently do the job, and we can always iterate and improve it some other time.

hko-s:
My proposed additions for this PR are in: https://github.com/baloo/rpgp/pull/2

baloo:
@hko-s merged your changes

hko-s:
> @hko-s merged your changes

Great, thanks!

PR #679, opened 2025-12-03
refactor!: introduce DecryptionKey trait

hko-s:


src/packet/key/secret.rs
```diff
@@ -490,6 +492,32 @@ impl SecretSubkey {
     }
 }
 
+impl DecryptionKey for SecretKey {
+    fn decrypt(
+        &self,
+        key_pw: &Password,
+        values: &PkeskBytes,
+        typ: EskType,
+    ) -> Result<Result<PlainSessionKey>> {
```
hko-s, line 501: Do we really want the double-`Result` in this return type?

dignifiedquire: it is useful to distinguish the two errors, but we don't have to

hko-s: I was wondering about the asymmetry between the return types of `SigningKey` and `DecryptionKey`.
But let's definitely not change return types in this PR (I have no strong impulse to change any of the return types).

PR #680, opened 2025-12-03
fix: Loosen trait requirements to KeyDetails, where appropriate

hko-s:
This PR also renames the key-type related generics to align with their new trait bound types.

### DIFF
--- a/src/packet/key/public.rs
+++ b/src/packet/key/public.rs
@@ -1112,6 +1112,7 @@ mod tests {
         fn public_key_packet_roundtrip(packet: PublicKey) {
             // dyn compat
             let _: Box<&dyn KeyDetails> = Box::new(&packet);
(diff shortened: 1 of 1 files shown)


src/packet/key/public.rs
```diff
@@ -1111,7 +1111,7 @@ mod tests {
         #[ignore]
         fn public_key_packet_roundtrip(packet: PublicKey) {
             // dyn compat
-            let _: Box<&dyn VerifyingKey> = Box::new(&packet);
+            let _: Box<&dyn KeyDetails> = Box::new(&packet);
```
hko-s, line 1114: I wonder if this is now actually checking what it should?

dignifiedquire: should do both I think

PR #682, opened 2025-12-18
refactor!: don't export symbols in duplicate

hko-s:
This looks like an analogous case to #639, where some symbols are (very likely accidentally) exported in two different locations.

### DIFF
--- a/src/types/params.rs
+++ b/src/types/params.rs
@@ -1,4 +1,11 @@
-pub mod encrypted_secret;
-pub mod plain_secret;
-pub mod public;
-pub mod secret;
+mod encrypted_secret;
+mod plain_secret;
+mod public;
+mod secret;
+
+pub use self::{
+    encrypted_secret::*,
+    plain_secret::*,
+    public::{ecdh::EcdhKdfType, *},
+    secret::*,
+};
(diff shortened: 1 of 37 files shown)


src/types/params.rs
```diff
@@ -1,7 +1,7 @@
-pub mod encrypted_secret;
-pub mod plain_secret;
-pub mod public;
-pub mod secret;
+mod encrypted_secret;
```
dignifiedquire: I would like to move to the inverse export, where we keep this, and remove the glob reexports in general

hko-s: Sounds good. In some cases the individual re-exports are shaping visibilities (I think?). But (if so) that seems like a bad pattern, that should be fixed anyway.

hko-s: I've dropped the re-exports in src/types/params.rs, and adjusted the rpgp-internal import accordingly, to see what that would look like. Is this what you had in mind?

If so, it's probably sensible to do it consistently in the entire public API?

dignifiedquire: I am sorry, but this is not great now that I see it, lets use the glob reexports for now

hko-s: Sure, I'm happy with that :)

PR #685, opened 2025-12-20
refactor!: distinguish can_encrypt_comms and can_encrypt_storage in key builder

### LINKED ISSUE #683: Key builder: Expose encrypt_comms and encrypt_storage separately in the API
Currently, the two encryption-related key flags `encrypt_comms` and `encrypt_storage` are handled as just one setting in the key builder, via the `can_encrypt` setting. To de-opinionate the API, I think we should split those two up.

hko-s:
fixes #683

### DIFF
--- a/examples/generate_key.rs
+++ b/examples/generate_key.rs
@@ -61,7 +61,7 @@ fn keygen(
     signkey
         .key_type(signing_key_type)
         .can_sign(true)
-        .can_encrypt(EncryptionCaps::All)
+        .can_encrypt(EncryptionCaps::None)
         .can_authenticate(false);
     let mut encryptkey = SubkeyParamsBuilder::default();
     encryptkey
@@ -73,7 +73,7 @@ fn keygen(
     authkey
         .key_type(auth_key_type)
         .can_sign(false)
-        .can_encrypt(EncryptionCaps::All)
+        .can_encrypt(EncryptionCaps::None)
         .can_authenticate(true);
 
     // Set up parameter builder for the full private key
@@ -82,7 +82,7 @@ fn keygen(
         .key_type(primary_key_type)
         .can_certify(true)
         .can_sign(false)
-        .can_encrypt(EncryptionCaps::All)
+        .can_encrypt(EncryptionCaps::None)
         .primary_user_id(uid.into())
         .subkeys(vec![
             signkey.build()?,


dignifiedquire:
how about using an enum, that includes an all variant, the new api is just unnecessary boilerplate for the default use case

hko-s:
> how about using an enum, that includes an all variant, the new api is just unnecessary boilerplate for the default use case

Good idea.

I sketched that approach. Not sure about the best naming for all the things.

src/composed/key/builder.rs
```diff
@@ -22,6 +24,29 @@ use crate::{
     },
 };
 
+/// A type to specify a configuration for the two key flags
+/// "encrypt communications" and "encrypt storage".
+///
+/// <https://www.rfc-editor.org/rfc/rfc9580#name-key-flags>
+#[derive(Default, Debug, PartialEq, Eq, Clone, Copy)]
+pub enum EncryptionFlags {
```
dignifiedquire: how about `Caps` for `Capabilities` which is what this is called usually in other systems

```rust
pub enum EncryptionCaps {
  None,
  Communication,
  Storage,
  All,
}
```

src/composed/key/builder.rs
```diff
@@ -22,6 +24,29 @@ use crate::{
...
+    #[default]
+    None,
+    Comms,
+    Storage,
+    CommsStorage,
+}
+
+impl EncryptionFlags {
+    fn comms(&self) -> bool {
```
dignifiedquire: should be called something like `can_` or `is_`

src/composed/key/builder.rs
```diff
@@ -126,7 +151,7 @@ impl SecretKeyParamsBuilder {
     fn validate_keytype(
         key_type: Option<&KeyType>,
         can_sign: Option<bool>,
-        can_encrypt: Option<bool>,
+        can_encrypt: Option<EncryptionFlags>,
```
dignifiedquire: no need for `Option` anymore? can just include the default of `None`

hko-s:
I've made changes to address all of your points.

src/composed/key/builder.rs
```diff
@@ -22,6 +24,29 @@ use crate::{
...
+    #[default]
+    None,
+    Communication,
+    Storage,
+    All,
+}
+
+impl EncryptionCaps {
+    fn is_communication(&self) -> bool {
+        *self == Self::Communication || *self == Self::All
```
dignifiedquire: nit: I have a slight preference to write these as `matches!(self, Self::Comunication | Self::All)`

hko-s: Oh yes, I like that!

PR #686, opened 2025-12-20
fix: use SHA-2 default hash algorithms

### LINKED ISSUE #684: Adjust default hash algorithms to SHA2
In https://github.com/rpgp/rpgp/blob/main/src/types/params/public.rs#L244-L255, we're picking Sha3 as the default hash algorithm for Ed448 and the PQC signing algorithms.

For Ed448, which is from RFC 9580, I think SHA3 is not the right default, because RFC 9580 specifies SHA2-256 as "MUST", and SHA2-384/SHA2-512 as "SHOULD", while SHA3 are only at the "MAY" level.

For the PQC signing algorithms, I tend to think the same reasoning applies (draft-ietf-openpgp-pqc does not add any new guidance about hash algorithms. However, it uses SHA3 as part of its key combiner, so there is some weak argument that implementations of draft-ietf-openpgp-pqc can be expected to have SHA3 support).

hko-s:
Fixes #684 

SHA-2 hash algorithms are in the MUST and SHOULD range, in RFC 9580.
As long as there is no OpenPGP guidance to prefer SHA-3, it seems more defensive to stick to SHA-2.

### DIFF
--- a/src/types/params/public.rs
+++ b/src/types/params/public.rs
@@ -241,18 +241,18 @@ impl PublicParams {
             PublicParams::ECDSA(EcdsaPublicParams::P384 { .. }) => HashAlgorithm::Sha384,
             PublicParams::ECDSA(EcdsaPublicParams::P521 { .. }) => HashAlgorithm::Sha512,
 
-            PublicParams::Ed448(_) => HashAlgorithm::Sha3_512,
+            PublicParams::Ed448(_) => HashAlgorithm::Sha512,
 
             #[cfg(feature = "draft-pqc")]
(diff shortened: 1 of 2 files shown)


hko-s:
To be clear, I think this PR is the most defensive thing we can do. But I don't think it's a super crucial change.

(Alternatively, we could also wait until the OpenPGP WG produces explicit guidance, and then adjust our `pubkey algo -> hash algo` mapping to that guidance. But I would not hold my breath for this guidance. The backlog of the WG is big.)

hko-s:
Closing this for now. It's not a crucial change, we can revisit the question later.

PR #693, opened 2025-12-30
docs: mark some decryption code as unused within rpgp

hko-s:


hko-s:
Not sure if we should actually drop all of this. But it seemed noteworthy that the code is not (as far as I can tell) used from within rPGP at all.

dignifiedquire:
I believe we are still using the encryption side, so I would like to keep it around at least as test code

hko-s:
> I believe we are still using the encryption side, so I would like to keep it around at least as test code

Sure, sounds good. I did get confused by the code's existence, for a bit.
Not sure how such confusion could be best avoided. I'll try adding some doc comments.

hko-s:
I added comments instead. Not sure if this PR is worth doing, though :shrug:

src/crypto/sym.rs
```diff
@@ -196,6 +196,8 @@ impl SymmetricKeyAlgorithm {
     /// Decrypt the data using CFB mode, without padding. Overwrites the input.
     /// Uses an IV of all zeroes, as specified in the openpgp cfb mode. Does
     /// resynchronization.
+    ///
+    /// NOTE: This code is not used in rPGP anymore
     pub fn decrypt(self, key: &[u8], prefix: &mut [u8], ciphertext: &mut [u8]) -> Result<()> {
```
dignifiedquire, line 204: actually we could mark these as `cfg(test)`

hko-s: I like that. Pushed a commit that does it.

PR #694, opened 2025-12-30
feat: expand zeroizing of secret values

### LINKED ISSUE #669: Find instances of missing Zeroize and add zeroizing
In some places (e.g. in https://github.com/rpgp/rpgp/blob/main/src/crypto/x25519.rs#L146-L149), sensitive data is stored in the heap, and not cleaned up (that's a block I wrote). I want to take a look around for more instances of that pattern (the same thing is definitely happening in the `x448` hkdf step), and improve those places.

hko-s:
Fixes #669

### DIFF
--- a/src/crypto/aes_kw.rs
+++ b/src/crypto/aes_kw.rs
@@ -5,6 +5,8 @@ use generic_array::{
 use snafu::{ResultExt, Snafu};
 use zeroize::Zeroizing;
 
+const IV_LEN: usize = 8;
+
 /// AES key wrap possible errors.
 #[derive(Debug, Snafu)]
 pub enum Error {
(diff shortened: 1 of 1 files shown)


hko-s:
I wonder if `pgp::crypto::Decryptor::decrypt` should return `Result<Zeroizing<Vec<u8>>>`?

hko-s:
> I wonder if `pgp::crypto::Decryptor::decrypt` should return `Result<Zeroizing<Vec<u8>>>`?

I've changed the return type, and that seems like the right approach, to me.

src/crypto/aes_kw.rs
```diff
@@ -65,7 +66,8 @@ pub fn unwrap(key: &[u8], data: &[u8]) -> Result<Vec<u8>, Error> {
         _ => {
             return Err(InvalidKeySizeSnafu { size: aes_size }.build());
         }
-    };
+    }
+    .map(Into::into);
```
dignifiedquire: we should use [`unwrap`](https://docs.rs/aes-kw/latest/aes_kw/struct.Kek.html#tymethod.unwrap)  instead, and not write into a "plain" vec in the first place

```rust
let key = GenericArray::<u8, U32>::from_slice(key);
let kek = aes_kw::KekAes256::new(key);
let len = data.len() - IV_LEN;
let mut out = Zeroizing::new(vec![0u8; len]);
kek.unwrap(data, &mut out)?;
Ok(out)
```

src/crypto/x25519.rs
```diff
@@ -135,23 +135,24 @@ pub fn hkdf(
     ephemeral: &[u8; 32],
     recipient_public: &[u8; 32],
     shared_secret: &[u8; 32],
-) -> Result<[u8; 16]> {
+) -> Result<Zeroizing<[u8; 16]>> {
```
dignifiedquire, line 138: this moves it from the stack to the heap, making it potentially easier to leak and slower, so not sure we should do this

src/crypto/aes_kw.rs
hko-s: Cool. I've pushed a commit with (I hope) this change

src/crypto/x25519.rs
hko-s: The returned value is a "secret", though, right? Is there some other way we can attempt to clear its content after use?

dignifiedquire (approved):
have you tested how this impacts performance?

hko-s:
> have you tested how this impacts performance?

I have made an attempt to measure performance changes, but on my notebook I think I'm measuring mostly thermal throttling effects :confused: 

I tried to narrowly run one short test that involves ECDH operations, to see if there's any noticable difference:

on branch `main`

```
message/x25519_encrypt_key_aes128_seipdv1/10000
                        time:   [80.359 µs 81.006 µs 81.721 µs]
                        thrpt:  [122.37 MB/s 123.45 MB/s 124.44 MB/s]
```

```
message/x25519_encrypt_key_aes128_seipdv1/10000
                        time:   [82.943 µs 83.516 µs 84.148 µs]
                        thrpt:  [118.84 MB/s 119.74 MB/s 120.57 MB/s]
```                          
---

on branch `zeroize`

```                          
message/x25519_encrypt_key_aes128_seipdv1/10000
                        time:   [83.750 µs 84.300 µs 84.805 µs]
                        thrpt:  [117.92 MB/s 118.62 MB/s 119.40 MB/s]
```                          

```                          
message/x25519_encrypt_key_aes128_seipdv1/10000
                        time:   [80.718 µs 81.528 µs 82.243 µs]
                        thrpt:  [121.59 MB/s 122.66 MB/s 123.89 MB/s]
```                          

... but even then, I seem to measure mostly noise, and can't see a difference.

PR #695, opened 2026-01-03
refactor!: clearly mark v3 key expiration fields as legacy

### LINKED ISSUE #681: V3 key packet expiration time handled incorrectly
In v2/3 key material (which is *very* legacy), expiration is modeled as a two byte value, counted in days (see https://www.rfc-editor.org/rfc/rfc2440.html#section-5.5.2)

However, some code erroneously handles the value as seconds:

- https://github.com/rpgp/rpgp/blob/24b8cc906d8358a5591510655e1081e2b034d5c6/src/packet/key/public.rs#L223
- https://github.com/rpgp/rpgp/blob/24b8cc906d8358a5591510655e1081e2b034d5c6/src/composed/key/builder.rs#L320

We should fix that (and check if there are other places that misinterpret this [very legacy] value).

hko-s:
Fixes #681


hko-s:
As noted in #681, the key builder also interacts with the legacy v3 key expiration field.

See e.g.: https://github.com/rpgp/rpgp/blob/24b8cc906d8358a5591510655e1081e2b034d5c6/src/composed/key/builder.rs#L320

We need to decide between two approaches in the key builder:

- Drop expiration time handling (for the time being). This would not be a regression in terms of functionality, since the current expiration handling is not functional for v4 (or newer) keys.
- Add expiration time handling for modern keys (this would mean carrying the expiration time along until the binding signature for primary and subkeys is made, and then to add the expiration as a signature subpacket).

hko-s:
Now that I look back at this, maybe the name should even be `legacy_v3_expiration_days` instead of `v3_expiration_days`

hko-s:
> Now that I look back at this, maybe the name should even be `legacy_v3_expiration_days` instead of `v3_expiration_days`

I've done this rename, to make the legacy-ness of that field extremely explicit.

hko-s:
And I've dropped expiration time handling from the key builder - it currently only populated the "legacy v3" expiration field, which has no effect for modern (v4+) keys.

So the change (even though it's formally removing functionality) should not be a regression for any real world use case.

dignifiedquire (approved):

PR #696, opened 2026-01-03
refactor: move support for large rsa keys into a separate feature

### LINKED ISSUE #666: clean up large RSA support
Currently, the default maximum size of RSA keys that rPGP will handle is 8192 bit (key generation is capped to 4096 bit).

Optionally, support for larger RSA keys (currently up to 16384 bit) can be enabled with the feature `malformed-artifact-compat`.
It would be good to move this functionality into a separate feature, with an rsa-specific name, before the next release.

In the longer run, we might consider dropping the feature entirely (however, it's unclear how to determine if any users rely on the feature).

hko-s:
Fixes #666

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,8 @@ fn main() -> pgp::errors::Result<()> {
 - `bzip2`: Enables bzip2 support
 - `asm`: Enables assembly based optimizations
 - `wasm`: Allows building for wasm
-- `malformed-artifact-compat`: Be lenient towards some types of malformed artifacts (erroneously formed ECDH PKESK; invalidly short first partial body segments), and allow use of very large RSA keys (>8192 bit). Most users will NOT need this feature, should be disabled by default!
+- `large-rsa`: Allow use of very large RSA keys (raises the limit from 8192 to 16384 bit)
+- `malformed-artifact-compat`: Be lenient towards some types of malformed artifacts (erroneously formed ECDH PKESK; invalidly short first partial body segments). Most users will NOT need this feature, should be disabled by default!
 - `draft-pqc`: Enables implementation of draft-ietf-openpgp-pqc-12 (This is unstable and can have breaking changes in patch releases. DO NOT USE IN PRODUCTION!)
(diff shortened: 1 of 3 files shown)


hko-s:
Todo:

- This feature is not currently running in CI
- Do we want to keep the cap at 16k size? (Once we allow that size, we could just as well also allow 32k?)

README.md
```diff
@@ -120,7 +120,8 @@ fn main() -> pgp::errors::Result<()> {
 - `bzip2`: Enables bzip2 support
 - `asm`: Enables assembly based optimizations
 - `wasm`: Allows building for wasm
-- `malformed-artifact-compat`: Be lenient towards some types of malformed artifacts (erroneously formed ECDH PKESK; invalidly short first partial body segments), and allow use of very large RSA keys (>8192 bit). Most users will NOT need this feature, should be disabled by default!
+- `large-rsa`: Allow use of very large RSA keys (>8192 bit)
```
dignifiedquire: add the max size as well please

dignifiedquire:
> * Do we want to keep the cap at 16k size? (Once we allow that size, we could just as well also allow 32k?)

lets stick with 16K or do we have any users that actually use more?

hko-s:
> lets stick with 16K or do we have any users that actually use more?

Alright. Happy to stick with 16k.

I don't know of any reason to allow bigger sizes. Everything >8k seems extremely exotic, and kind of silly.
So I'm struggling to decide where the upper hard cutoff limit should be :sweat_smile:

README.md
hko-s: good point. done.

hko-s:
Note to self - TODO: add a CI run that does only `cargo check` to make sure the codebase compiles with the feature enabled

hko-s:
> Note to self - TODO: add a CI run that does only `cargo check` to make sure the codebase compiles with the feature enabled

I looked into this and realized two things:
- The [`clippy_check`](https://github.com/rpgp/rpgp/blob/main/.github/workflows/ci.yml#L225-L238) CI job already covers this. It ensures we can compile with `--all-features`
- For some reason, [many of our feature CI jobs](https://github.com/rpgp/rpgp/blob/main/.github/workflows/ci.yml#L54-L72) are run with the nightly toolchain (only). This seems unintentional, and mostly benign - but weird.

So I'd say that this PR is ok as is: It is trivial. And we're checking that it compiles, in CI.

PR #697, opened 2026-01-06
[WIP] fix: ensure readers always consume the full data

dignifiedquire:


hko-s:
Just took a peek into this, and looked at why `test_aead_skesk_msg_eax` is failing. Looks like the reason is that there's a padding packet after the literal data packet. We need to be able to consume that.

dignifiedquire:
Closing in favor of #699

PR #707, opened 2026-01-24
Fix CVE-2026-22705

loganaden:
https://github.com/RustCrypto/signatures/security/advisories/GHSA-hcp2-x6j4-29j7

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -107,7 +107,7 @@ x25519-dalek = { version = "2.0.1", default-features = false, features = [
 
 # PQC
 ml-kem = { version = "0.2.1", features = ["zeroize", "deterministic"], optional = true }
-ml-dsa = { version = "0.0.4", features = ["zeroize"], optional = true }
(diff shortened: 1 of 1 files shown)


dignifiedquire:
Unfortunately this fix is unusable, as this requires newer versions of dependencies that we can't use as rust crypto has not yet released new version

loganaden:
got it. will wait, then update.

hko-s:
> Unfortunately this fix is unusable, as this requires newer versions of dependencies that we can't use as rust crypto has not yet released new version

Just to link it from here: PR #573 tracks the upcoming Rust Crypto releases, and ports rPGP to their current pre-release versions.

So the fix for ml-dsa will most naturally be merged via that PR.

hko-s:
Closing this in favor of #573.

rPGP needs to replace all Rust Crypto crates in one feel swoop, and it will get the ml-dsa update in that context.

PR #721, opened 2026-02-02
Fix key/value parser for armored files

aguinetsb:
There can be cases where the input string is

```
--- BEGIN [...] ---
(no header)

(base64)
--- END [...] ---

this is noise: more noise
```

The current key/value parser for headers won't stop at new lines and will consume the input until the ":" in the noise. This PR fixes this and explicitly search for lines before parsing headers.

### DIFF
--- a/src/armor/reader.rs
+++ b/src/armor/reader.rs
@@ -5,11 +5,11 @@ use buffer_redux::BufReader;
 use byteorder::{BigEndian, ByteOrder};
 use nom::{
     branch::alt,
-    bytes::streaming::{tag, take, take_until, take_until1},
+    bytes::streaming::{tag, take, take_until},
     character::streaming::{digit1, line_ending, not_line_ending, space0},
-    combinator::{complete, map, map_res, opt, success, value},
+    combinator::{all_consuming, complete, map, map_res, opt, rest, success, value, verify},
     multi::many0,
-    sequence::{delimited, pair, preceded, terminated},
+    sequence::{delimited, pair, preceded, separated_pair, terminated},
     AsChar, IResult, Input, Parser,
 };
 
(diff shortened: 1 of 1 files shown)


dignifiedquire:
Is this actually a spec compliant version?  If so can you make sure to add a test that fails without this change please

aguinetsb:
@dignifiedquire good question, let me check.

In the meantime I have a related one: is rpgp designed to parse armored keys that are potentially followed by random noise, or is the input reader supposed to stop at anything that wouldn't fit the spec?

If this is the former, and the format above is not spec compliant, then indeed maybe this PR is irrelevant.

Our use case is that we basically want to extract PGP keys from "arbitrary files". So we'd look for the armor header, decode the key and stop at the armor footer. At least ideally that's what we'd like to do. That would allow not to have a temporary buffer of the armored version in memory.

Thanks!

hko-s:
Sorry for the slow reply!
> [..] is rpgp designed to parse armored keys that are potentially followed by random noise, or is the input reader supposed to stop at anything that wouldn't fit the spec?

rPGP's dearmoring aims to be restrictive. Over time, we've locked down what's accepted without throwing an error, and I think we might do more of that. (This is particularly relevant for "clearsigned messages", but I think it's also good for a library to limit lenience for regular armored inputs.)

For the use case you describe, pre-processing the input seems like a good approach to me.

hko-s:
FYI, there are a lot of tests that deal with weird ASCII Armor in the Interop Test Suite at https://sequoia-pgp.gitlab.io/openpgp-interoperability-test-suite/results.html#Mangled_ASCII_Armored_Key

Those also give some measurement of what the various implementations currently do in a number of edge cases.

PR #723, opened 2026-02-14
feat: implement more functionality from draft-wussler-openpgp-forwarding

hko-s:
This PR extends rPGP to cover all functionality from https://www.ietf.org/archive/id/draft-wussler-openpgp-forwarding-00.html.

(All code that handles the formats from this draft are feature-gated and not available in rPGP by default)

### DIFF
--- a/src/packet/key/secret.rs
+++ b/src/packet/key/secret.rs
@@ -844,9 +844,9 @@ mod tests {
         .expect("compute_proxy_parameter");
 
         assert_eq!(
-            k.as_ref(),
             hex::decode("e89786987c3a3ec761a679bc372cd11a425eda72bd5265d78ad0f5f32ee64f02")
-                .unwrap()
+                .unwrap(),
+            k.as_ref(),
         );
     }
 }
--- a/src/packet/public_key_encrypted_session_key.rs
+++ b/src/packet/public_key_encrypted_session_key.rs
@@ -418,7 +418,7 @@ impl PublicKeyEncryptedSessionKey {
             bail!("Ephemeral public key belongs to a small subgroup");
         }
 
-        let k = Zeroizing::new(Scalar::from_bytes_mod_order(k.into()));
+        let k = Zeroizing::new(Scalar::from_bytes_mod_order(k.into_array()));
 
         // eC = k * eB
         let ec = (*k) * ephemeral;
--- a/src/types.rs
+++ b/src/types.rs
@@ -63,15 +63,15 @@ impl From<[u8; 32]> for ForwardingProxyParameter {
 }
 
 #[cfg(feature = "draft-wussler-openpgp-forwarding")]
-impl From<ForwardingProxyParameter> for [u8; 32] {
-    fn from(value: ForwardingProxyParameter) -> Self {
-        value.0
+impl ForwardingProxyParameter {
+    pub(crate) fn into_array(self) -> [u8; 32] {
+        self.0
     }
 }
 
 #[cfg(feature = "draft-wussler-openpgp-forwarding")]
-impl AsRef<[u8]> for ForwardingProxyParameter {
-    fn as_ref(&self) -> &[u8] {
+impl AsRef<[u8; 32]> for ForwardingProxyParameter {
+    fn as_ref(&self) -> &[u8; 32] {
         &self.0
     }
 }
--- a/tests/forwarding.rs
+++ b/tests/forwarding.rs
@@ -100,8 +100,8 @@ fn forward_a_3_calculate_proxy_param() {
         .expect("generate_proxy_parameter");
 
     assert_eq!(
+        hex::decode("04b657045fc9c0759c5fd11d8ca75a2b1aa101c9c896490bcec100f941e97e0e").unwrap(),
         k.as_ref(),
-        hex::decode("04b657045fc9c0759c5fd11d8ca75a2b1aa101c9c896490bcec100f941e97e0e").unwrap()
     );
 }
 


src/packet/key/secret.rs
```diff
@@ -214,6 +214,58 @@ impl SecretSubkey {
...
         &self.details
     }
+
+    /// Given recipient and forwardee encryption subkeys, compute proxy transformation parameter.
+    #[cfg(feature = "draft-wussler-openpgp-forwarding")]
+    pub fn generate_proxy_parameter(
+        &self,
+        forwardee: &SecretSubkey,
+        password_recipient: &Password,
+        password_fowardee: &Password,
```
lubux: TYPO

tests/forwarding.rs
```diff
@@ -63,7 +81,78 @@ yGZuVVMAK/ypFfebDf4D/rlEw3cysv213m8aoK8nAUO8xQX3XQq3Sg+EGm0BNV8E
...
+        &k[..],
+        hex::decode("04b657045fc9c0759c5fd11d8ca75a2b1aa101c9c896490bcec100f941e97e0e").unwrap()
+    );
+}
+
+#[test]
+fn forward_a_3_transform_pkesk() {
+    let _ = pretty_env_logger::try_init();
+
+    // Get the PKESK from ENCYPTED_MESSAGE
```
lubux:
```suggestion
    // Get the PKESK from ENCRYPTED_MESSAGE
```

src/packet/key/secret.rs
```diff
@@ -214,6 +214,58 @@ impl SecretSubkey {
...
+    ///
+    /// k = dB/dC mod n
+    /// return k
+    #[cfg(feature = "draft-wussler-openpgp-forwarding")]
+    fn compute_proxy_parameter(mut db: [u8; 32], mut dc: [u8; 32]) -> Result<[u8; 32]> {
+        db.reverse();
+        dc.reverse();
+
+        let recipient = curve25519_dalek::Scalar::from_bytes_mod_order(db);
+        let forwardee = curve25519_dalek::Scalar::from_bytes_mod_order(dc);
```
lubux: Maybe we should doublecheck that the `forwardee` Scalar is not 0`

```
invert(): 
Warning
self MUST be nonzero. If you cannot prove that this is the case, you SHOULD NOT USE THIS FUNCTION.
```

```rust
        if forwardee.ct_eq(&Scalar::ZERO).into() {
            bail!("Forwardee private key is zero");
        }
```

src/packet/key/secret.rs
```diff
@@ -214,6 +214,58 @@ impl SecretSubkey {
...
+    ///   Implements ComputeProxyParameter( dB, dC );
+    ///    Input:
+    ///     dB - the recipient's private key integer
+    ///     dC - the forwardee's private key integer
+    ///     n - the size of the field of Curve25519
+    ///
+    /// k = dB/dC mod n
+    /// return k
+    #[cfg(feature = "draft-wussler-openpgp-forwarding")]
+    fn compute_proxy_parameter(mut db: [u8; 32], mut dc: [u8; 32]) -> Result<[u8; 32]> {
```
lubux: There is no zeroizing happening on private params, do we intend to have this? 

- db
- dc
- proxy parameter

src/packet/key/secret.rs
```diff
@@ -214,6 +214,58 @@ impl SecretSubkey {
...
+        db.reverse();
```
lubux, line 266: A comment indicating that we reverse the bytes for little-endian encoding, would help parsing the code.

src/packet/key/secret.rs
lubux: We must also ensure that the secrets are clamped
```
        let recipient = curve25519_dalek::Scalar::from_bytes_mod_order(
            curve25519_dalek::scalar::clamp_integer(db),
        );
        let forwardee = curve25519_dalek::Scalar::from_bytes_mod_order(
            curve25519_dalek::scalar::clamp_integer(dc),
        );
```

src/packet/public_key_encrypted_session_key.rs
```diff
@@ -289,6 +289,134 @@ impl PublicKeyEncryptedSessionKey {
...
+        forwardee: &K,
+        proxy_parameter: [u8; 32],
+    ) -> Result<Self> {
+        use crate::types::{EcdhKdfType, EcdhPublicParams, Mpi};
+
+        let PublicKeyEncryptedSessionKey::V3 {
+            packet_header,
+            id,
+            values,
+            pk_algo,
```
lubux, line 310: Should we add a check that pk_algo is ECDH with x22519 in addition to the values check?
```
let PkeskBytes::Ecdh {
            ref mut public_point,
            ..
        } = values
```

src/packet/public_key_encrypted_session_key.rs
```diff
@@ -289,6 +289,134 @@ impl PublicKeyEncryptedSessionKey {
...
+        // Strip 0x40 prefix for proxy transformation
+        let public_point_data = &public_point.as_ref()[1..33];
+
+        let mut forwarded_point = Self::transform_ecdh_ephemeral(
+            public_point_data.try_into().expect("32 bytes"),
+            proxy_parameter,
+        )?
+        .to_vec();
+        // re-add 0x40 prefix for the OpenPGP "MPI" representation
+        forwarded_point.insert(0, 0x40);
```
lubux: Nit:  We Could avoid the vec allocation with something like:

```
        let mut forwarded_point = [0u8; 33];
        forwarded_point[0] = 0x40;
        forwarded_point[1..].copy_from_slice(
            &Self::transform_ecdh_ephemeral(
                public_point_data.try_into().expect("32 bytes"),
                proxy_parameter,
            )?
        );
```

src/packet/public_key_encrypted_session_key.rs
```diff
@@ -289,6 +289,134 @@ impl PublicKeyEncryptedSessionKey {
...
+    /// Transform PKESK to forward a message to a 'forwardee'.
+    ///
+    /// The underlying forwarding mechanism only supports ECDH with Curve 25519.
+    ///
+    /// <https://www.ietf.org/archive/id/draft-wussler-openpgp-forwarding-00.html#name-forwarding-messages>
+    #[cfg(feature = "draft-wussler-openpgp-forwarding")]
+    pub fn forwarding_transform<K: KeyDetails>(
+        &self,
+        forwardee: &K,
+        proxy_parameter: [u8; 32],
```
lubux: Private params with no zeroizing

src/packet/public_key_encrypted_session_key.rs
```diff
@@ -514,4 +642,48 @@ mod tests {
...
+
+        let small = hex::decode("ecffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff7f")
+            .unwrap();
+
+        let res = PublicKeyEncryptedSessionKey::transform_ecdh_ephemeral(
+            small.try_into().unwrap(),
+            [0; 32],
+        );
+
+        matches!(res, Err(crate::errors::Error::Message { .. }));
```
lubux:
```suggestion
        assert!(matches!(res, Err(crate::errors::Error::Message { .. })));
```

src/packet/public_key_encrypted_session_key.rs
```diff
@@ -289,6 +289,134 @@ impl PublicKeyEncryptedSessionKey {
...
+        // if 0x08 * eB == 0 then abort
+        if (Scalar::from(8u8) * ephemeral).to_bytes() == [0; 32] {
+            bail!("Ephemeral public key belongs to a small subgroup");
+        }
+
+        let k = Scalar::from_bytes_mod_order(k);
+
+        // eC = k * eB
+        // return eC
+        let ec = k * ephemeral;
```
lubux: We also might want to check this for warning the user early on. e.g., if k = 0

```
        if ec.is_identity() {
            bail!("Transformed ephemeral is the identity point");
        }
```

src/packet/public_key_encrypted_session_key.rs
```diff
@@ -289,6 +289,134 @@ impl PublicKeyEncryptedSessionKey {
...
+    ///   eB - the ECDH ephemeral public key decoded from the PKESK
+    ///   k - the proxy transformation parameter retrieved from storage
+    #[cfg(feature = "draft-wussler-openpgp-forwarding")]
+    fn transform_ecdh_ephemeral(eb: [u8; 32], k: [u8; 32]) -> Result<[u8; 32]> {
+        use curve25519_dalek::{MontgomeryPoint, Scalar};
+
+        let ephemeral = MontgomeryPoint(eb);
+
+        // if 0x08 * eB == 0 then abort
+        if (Scalar::from(8u8) * ephemeral).to_bytes() == [0; 32] {
```
lubux: we could us is_identity directly

src/packet/public_key_encrypted_session_key.rs
```diff
@@ -514,4 +642,48 @@ mod tests {
...
+    fn forward_transform_small_subgroup_a_2() {
+        // Small subgroup detection test from
+        // <https://www.ietf.org/archive/id/draft-wussler-openpgp-forwarding-00.html#name-message-transformation>
+
+        let small = hex::decode("ecffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff7f")
+            .unwrap();
+
+        let res = PublicKeyEncryptedSessionKey::transform_ecdh_ephemeral(
+            small.try_into().unwrap(),
+            [0; 32],
```
lubux: For the small subgroup test we should use a non-zero k.

src/packet/public_key_encrypted_session_key.rs
hko-s: Good point to ensure `pk_algo` is ECDH. I've added that check.

I don't think the PKESK knows which curve is used with ECDH though? (https://www.rfc-editor.org/rfc/rfc9580#name-version-3-public-key-encryp)

src/packet/public_key_encrypted_session_key.rs
hko-s: Should be fixed with `ProxyParameter` newtype

src/packet/key/secret.rs
hko-s: Good point :thinking:.

I introduced a `ProxyParameter` newtype (that derives `ZeroizeOnDrop`) to take care of a subset of this problem.

The parameters `db` and `dc` are a bit awkward here (I've mostly exposed this low-level function in order to run a unit test with the test vectors from the draft against it (`forward_proxy_param_a_1`)).
The function takes ownership of these arrays, and it should now zeroize their data internally. This approach feels a bit brittle, but I'm not sure how to improve on it.

src/composed/message/types.rs
```diff
@@ -490,6 +490,9 @@ pub enum Edata<'a> {
 pub struct DecryptionOptions {
     legacy: bool,
     gnupg_aead: bool,
+
+    #[cfg(feature = "draft-wussler-openpgp-forwarding")]
+    draft_forwarding: bool,
```
dignifiedquire, line 498: docs?

src/composed/message/types.rs
```diff
@@ -521,6 +524,13 @@ impl DecryptionOptions {
         self.gnupg_aead = true;
         self
     }
+
+    /// Enable decryption of forwarded messages as specified in draft-wussler-openpgp-forwarding
+    #[cfg(feature = "draft-wussler-openpgp-forwarding")]
+    pub fn enable_draft_forwarding(mut self) -> Self {
```
dignifiedquire, line 533: what about disabeling?

src/packet/key/secret.rs
```diff
@@ -214,6 +214,76 @@ impl SecretSubkey {
...
+                _ => bail!("Only ECDH/Curve 25519 recipient keys are supported"),
+            },
+        )?
+    }
+
+    /// Given the recipient and forwardee private key integers, compute the proxy transformation parameter.
+    ///
+    /// See <https://www.ietf.org/archive/id/draft-wussler-openpgp-forwarding-00.html#name-computing-the-proxy-paramet>
+    ///
+    ///   Implements ComputeProxyParameter( dB, dC );
```
dignifiedquire: strange indent?

src/packet/public_key_encrypted_session_key.rs
```diff
@@ -289,6 +289,146 @@ impl PublicKeyEncryptedSessionKey {
...
+            values,
+        })
+    }
+
+    /// The message forwarding transformation from draft-wussler-openpgp-forwarding
+    ///
+    /// <https://www.ietf.org/archive/id/draft-wussler-openpgp-forwarding-00.html#name-forwarding-messages>
+    ///
+    /// Implements TransformMessage( eB, k );
+    ///   Input:
```
dignifiedquire: strange indent

src/packet/public_key_encrypted_session_key.rs
```diff
@@ -289,6 +289,146 @@ impl PublicKeyEncryptedSessionKey {
...
+    fn transform_ecdh_ephemeral(
+        eb: [u8; 32],
+        k: crate::types::ForwardingProxyParameter,
+    ) -> Result<[u8; 32]> {
+        use curve25519_dalek::{traits::IsIdentity, MontgomeryPoint, Scalar};
+
+        let ephemeral = MontgomeryPoint(eb);
+
+        // if 0x08 * eB == 0 then abort
+        if (Scalar::from(8u8) * ephemeral).is_identity() {
```
dignifiedquire, line 417: can this be a constant?

src/types.rs
```diff
@@ -48,3 +48,30 @@ pub enum EskType {
     /// V6 PKESK or SKESK (introduced in RFC 9580)
     V6,
 }
+
+#[cfg(feature = "draft-wussler-openpgp-forwarding")]
+#[derive(zeroize::ZeroizeOnDrop)]
```
dignifiedquire: derives after comments

src/types.rs
```diff
@@ -48,3 +48,30 @@ pub enum EskType {
...
+
+#[cfg(feature = "draft-wussler-openpgp-forwarding")]
+impl From<[u8; 32]> for ForwardingProxyParameter {
+    fn from(value: [u8; 32]) -> Self {
+        Self(value)
+    }
+}
+
+#[cfg(feature = "draft-wussler-openpgp-forwarding")]
+impl From<ForwardingProxyParameter> for [u8; 32] {
```
dignifiedquire: is this secret material, if so, should this be Zeroizing wrapped, and not this easily extractable?

hko-s: To my mind, the proxy parameter is a weird kind of not-extremely secret (AIUI, it's only useful/dangerous in combination with the forwardee's private key material).

That said, I didn't intend to expose this particular function in the public API, so I've locked it down to a crate local implementation.

(But there's still an AsRef implementation that gives out the parameter. The parameter is ultimately needed outside of rPGP: its natural habitat is in an MTA.)

src/composed/message/types.rs
hko-s: We don't have disabling for the analogous `enable_legacy` and `enable_gnupg_aead` either.

The default for all of them is "not allowed", and these functions are part of a "builder"-like interface.
I don't think we need to allow callers to switch these bools back and forth.

src/packet/public_key_encrypted_session_key.rs
hko-s: As far as I can tell, `curve25519-dalek` doesn't expose a const way to make a Scalar, except for the deprecated https://docs.rs/curve25519-dalek/4.1.3/curve25519_dalek/scalar/struct.Scalar.html#method.from_bits

PR #740, opened 2026-03-11
api!: mark legacy ECC curves as such

link2xt:
Currently Ed25519Legacy and Curve25519Legacy
curves are named just Ed25519 and Curve25519
in the code and this is not what
<https://www.rfc-editor.org/rfc/rfc9580.html#name-ecc-curves-for-openpgp> does.

It is not clear that KeyType::ECDH(ECCCurve::Curve25519) is using Curve25519Legacy and you want to use KeyType::X25519 instead to avoid legacy formats.
KeyType::Ed25519Legacy is named correctly,
but uses ECCCurve::Ed25519 inside
which should be named ECCCurve::Ed25519Legacy.

### DIFF
--- a/benches/benchmarks/key.rs
+++ b/benches/benchmarks/key.rs
@@ -34,18 +34,21 @@ fn bench_key(c: &mut Criterion) {
         b.iter(|| black_box(SignedSecretKey::from_armor_single(&bytes[..]).unwrap()));
     });
 
-    g.bench_function("x25519_parse_armored", |b| {
-        let key = build_key(KeyType::Ed25519Legacy, KeyType::ECDH(ECCCurve::Curve25519));
+    g.bench_function("ed25519legacy_ecdh_curve25519legacy_parse_armored", |b| {
+        let key = build_key(
+            KeyType::Ed25519Legacy,
+            KeyType::ECDH(ECCCurve::Curve25519Legacy),
+        );
         let bytes = key.to_armored_bytes(None.into()).unwrap();
 
         b.iter(|| black_box(SignedSecretKey::from_armor_single(&bytes[..]).unwrap()));
     });
 
-    g.bench_function("x25519_generate", |b| {
+    g.bench_function("ed25519legacy_ecdh_curve25519legacy_generate", |b| {
         b.iter(|| {
             black_box(build_key(
                 KeyType::Ed25519Legacy,
-                KeyType::ECDH(ECCCurve::Curve25519),
+                KeyType::ECDH(ECCCurve::Curve25519Legacy),
             ))
         })
     });
--- a/src/crypto/ecc_curve.rs
+++ b/src/crypto/ecc_curve.rs
@@ -5,16 +5,31 @@ use crate::{
     errors::unsupported_err,
 };
 
+/// ECC Curves.
+///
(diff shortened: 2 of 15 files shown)


src/crypto/ecc_curve.rs
```diff
@@ -5,17 +5,32 @@ use crate::{
...
+/// <https://www.rfc-editor.org/rfc/rfc9580.html#name-ecc-curves-for-openpgp>
 #[derive(Clone, Debug, Eq, PartialEq)]
 #[cfg_attr(test, derive(proptest_derive::Arbitrary))]
 pub enum ECCCurve {
-    Curve25519,
-    Ed25519,
+    /// Curve25519Legacy ("1.3.6.1.4.1.3029.1.5.1") curve.
+    ///
+    /// For modern applications you probably want to use X25519 instead of ECDH with this curve.
+    Curve25519Legacy,
```
link2xt, line 17: This renaming is a breaking change. If we absolutely want to avoid breaking changes, we can keep the name as is, then I will only add documentation.

link2xt:
cc @hko-s @dignifiedquire This is basically ready, if you don't have objections to breaking API, I will finish the TODOs/FIXMEs. I think it is much more clear when the names match the RFC 9580, the second best option is to put RFC names into the documentation comments at least.

benches/benchmarks/key.rs
```diff
@@ -35,7 +35,11 @@ fn bench_key(c: &mut Criterion) {
     });
 
     g.bench_function("x25519_parse_armored", |b| {
-        let key = build_key(KeyType::Ed25519Legacy, KeyType::ECDH(ECCCurve::Curve25519));
+        // TODO: benchmark is named x25519_parse_armored, should use X25519?
```
hko-s: Before RFC 9580 came out, "x25519" was often (and reasonably, I guess) used as shorthand naming for ECDH over Curve25519 ("legacy").

I think we should just adjust the name of the benchmark, to straighten out the terminology. Maybe to `ecdh_curve25519_legacy`, or something similar?

src/crypto/ecc_curve.rs
hko-s: I think a semver break is fine, as long as we otherwise like the change.

src/types/params/public/ecdh.rs
```diff
@@ -60,6 +60,7 @@ impl EcdhKdfType {
 #[cfg_attr(test, derive(proptest_derive::Arbitrary))]
 pub enum EcdhPublicParams {
     /// ECDH public parameters for a curve that we know uses Mpi representation
+    // TODO: should be renamed to Curve25519Legacy?
```
hko-s: Renaming this as well seems appropriate to me

hko-s:
> cc @hko-s @dignifiedquire This is basically ready, if you don't have objections to breaking API, I will finish the TODOs/FIXMEs. I think it is much more clear when the names match the RFC 9580, the second best option is to put RFC names into the documentation comments at least.

Thanks! I agree that aligning these terms with RFC 9580 is better in the long run. When I implemented the new parts of RFC 9580 in rPGP, it seemed too intrusive to rename these types, so I kept the old names. But at this point, the renames seem less confusing to me as well.

Breaking semver for this doesn't worry me (we'll probably do some more of that before the next release again, anyway :sweat_smile:)

PR #741, opened 2026-03-11
fix: abort on all-zero x25519 shared secret

link2xt:
We already do this for x448.

### DIFF
--- a/src/crypto/ecdh.rs
+++ b/src/crypto/ecdh.rs
@@ -9,7 +9,7 @@ use crate::{
         aes_kw, ecc_curve::ECCCurve, public_key::PublicKeyAlgorithm, sym::SymmetricKeyAlgorithm,
         Decryptor,
     },
-    errors::{ensure, ensure_eq, unsupported_err, Error, Result},
+    errors::{bail, ensure, ensure_eq, unsupported_err, Error, Result},
     ser::Serialize,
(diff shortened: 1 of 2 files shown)


link2xt:
Here x448 has a similar check:
https://github.com/rpgp/rpgp/blob/12e61b359cf184bc657e9a464ba628cae6dc93f8/src/crypto/x448.rs#L191-L194

So this is for consistency.

https://www.rfc-editor.org/rfc/rfc7748 says you MAY abort on all-zero shared secret for x25519 and x448.

link2xt:
https://docs.rs/x25519-dalek/latest/x25519_dalek/struct.SharedSecret.html#method.was_contributory says it is constant-time.

hko-s:
It seems to me there's consensus that checking for an all-zero shared secret is a sensible thing to do.
I'm not a qualified reviewer for the ECC maths this PR deals with, but I still took a look.

After some reading, I agree that the already existing check in the x448 case (in [`x448::Secret::as_diffie_hellman`](https://docs.rs/x448/latest/x448/struct.Secret.html#method.as_diffie_hellman)) deals with the same class of problem.
But now I wonder if that check covers all relevant cases that could lead to an all-zero shared secret. An all zero secret key would also produce an all-zero shared secret, wouldn't it? (But I don't know if should we care about that case)

Either way, as far as I can tell, the change in this PR is an improvement.

PR #743, opened 2026-03-17
Add tests with unusual keys

wiktor-k:
Hi folks :wave: 

In Signstar we had a couple of tests with keys of non-standard formats which caused issues previously. I'm not quite sure if they were rpgp issues but in the end I think these are better off in this repository. Everything works well but these may be considered regression tests if ever anyone changes the parser code.

See ya!


hko-s:
Great, thank you!

I think it might be nice to add operations that use those keys - and/or fixed test vectors that use them. To exercise the key material in actual use.

I'll try to come up with something soon.

wiktor-k:
> Great, thank you!
> 
> I think it might be nice to add operations that use those keys - and/or fixed test vectors that use them. To exercise the key material in actual use.
> 
> I'll try to come up with something soon.

Yep, good idea. We can sign something and then compare that with static data... or sign and verify :thinking:

wiktor-k:
FYI @hko-s added sign/verify as instructed. :wave:

hko-s:
Thanks for adding more test code, even though I said I would :sweat_smile:

... I'll try one more round of note-to-self-ing here:

As a final bit of defensive pinning down the semantics, I think it would be good to check against external artifacts made with these keys (e.g. detached signatures that GnuPG produced with them)

hko-s:
I've added GnuPG signatures and tests for them in https://github.com/hko-s/rpgp/commit/c42788feebe2f399efbf1b9358158e107d85e173

(And moved the artifacts into a separate directory. The structure of the test data is not very disciplined, my intuition is to move these odd keys into a separate place, and hopefully to remember to add other weird keys into that directory later :sweat_smile:)

wiktor-k:
> I've added GnuPG signatures and tests for them in [hko-s@c42788f](https://github.com/hko-s/rpgp/commit/c42788feebe2f399efbf1b9358158e107d85e173)
> 
> (And moved the artifacts into a separate directory. The structure of the test data is not very disciplined, my intuition is to move these odd keys into a separate place, and hopefully to remember to add other weird keys into that directory later 😅)

Thanks! I've cherry-picked that particular commit and pushed here. :wave:

PR #744, opened 2026-03-20
refactor: replace `regex` with `memchr` for newline normalization

CathalMullan:
Avoids pulling in a heavy dep like `regex` for this singular usecase.

Locally this cuts the release build time from 14s -> 7s.

Timings:
- [Before.html](https://github.com/user-attachments/files/26130784/BEFORE.html)
- [After.html](https://github.com/user-attachments/files/26130786/AFTER.html)

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ fn main() -> pgp::errors::Result<()> {
 - `large-rsa`: Allow use of very large RSA keys (raises the limit from 8192 to 16384 bit)
 - `malformed-artifact-compat`: Be lenient towards some types of malformed artifacts (erroneously formed ECDH PKESK; invalidly short first partial body segments). Most users will NOT need this feature, should be disabled by default!
 - `draft-pqc`: Enables implementation of draft-ietf-openpgp-pqc-12 (This is unstable and can have breaking changes in patch releases. DO NOT USE IN PRODUCTION!)
-- `draft-wussler-openpgp-forwarding`: Enables support for the formats from [draft-wussler-openpgp-forwarding](https://datatracker.ietf.org/doc/draft-wussler-openpgp-forwarding/), and decryption of forwarded ECDH Curve25519 OpenPGP messages
+- `draft-wussler-openpgp-forwarding`: Enables support for the formats and functionality from [draft-wussler-openpgp-forwarding](https://datatracker.ietf.org/doc/draft-wussler-openpgp-forwarding/)
 
 ## Current Status
 
--- a/src/composed/message/types.rs
+++ b/src/composed/message/types.rs
@@ -490,6 +490,12 @@ pub enum Edata<'a> {
 pub struct DecryptionOptions {
     legacy: bool,
     gnupg_aead: bool,
+
+    /// Allow processing of forwarded messages.
+    ///
+    /// Also see [`DecryptionOptions::enable_draft_forwarding`].
+    #[cfg(feature = "draft-wussler-openpgp-forwarding")]
+    draft_forwarding: bool,
 }
 
 impl DecryptionOptions {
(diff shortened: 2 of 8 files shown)


hko-s:
Wow! Those build time graphs look amazing, thanks a lot!

I'll need a minute to take a closer look, but hope to get to it sometime next week.

dignifiedquire:
looks good to me

cosmicexplorer:
1. awesome change!!! I could not tell you how delighted I was to see this work.
2. regarding @hko-s:
> After staring at the diff for a bit, I can't come up with any ideas for edge cases that this change could break

The function of `memchr` is imo simpler than `regex` and edge cases tend to be generally reduced (this sounds kind of rude — I am mostly trying to motivate the next comment).

3. I created a path library and it has a particular construction for splitting by a single byte (or two bytes), expressed as an iterator, covering edge cases for normalization of path separators: https://codeberg.org/cosmicexplorer/deep-link/src/commit/1ea3eba5d599d8c48ea56816b7103b12ee49d505/trait-utils/src/bounds.rs#L28
This approach to state can be useful for splitting:
```rust
#[derive(Debug, Copy, Clone, PartialEq, Eq)]
pub enum SeparatorReadHead {
  BeforeStart,
  /// May be any valid index into the slice, from 0..len.
  MatchedInternalPosition(usize),
  AfterEnd,
}
```

That generic code is used for path separator splitting with: https://codeberg.org/cosmicexplorer/deep-link/src/commit/1ea3eba5d599d8c48ea56816b7103b12ee49d505/d-major/pathing/src/splitting/split_impl/components.rs#L33

4. In the `zip` crate I was able to introduce the first instance of safe SIMD zip parsing—it's been improved a lot since then but you can also see this generic abstraction being formed around the `memchr` primitive (hence why I found this change so exciting): https://github.com/zip-rs/zip2/blob/6244796c7f84719f7427d07a0e74f378ca7b5c7c/src/read/magic_finder.rs#L248
```rust
pub trait FinderDirection<'a> {
    fn new(needle: &'a [u8]) -> Self;
    fn reset_cursor(bounds: (u64, u64), window_size: usize) -> u64;
    fn scope_window(window: &[u8], mid_window_offset: usize) -> (&[u8], usize);

    fn needle(&self) -> &[u8];
    fn find(&self, haystack: &[u8]) -> Option<usize>;
    fn move_cursor(&self, cursor: u64, bounds: (u64, u64), window_size: usize) -> Option<u64>;
    fn move_scope(&self, offset: usize) -> usize;
}
```

We have been unable to make that into a DoubleEndedIterator yet, but that's not always a bad thing.

5. Finally if you're not familiar, I strongly recommend reading the rebar benchmarks by the regex/memchr author, who is also great to work with https://github.com/BurntSushi/rebar/blob/master/README.md some the only good benchmarks I've ever seen, and he identifies a point relevant to this:
> This group is mainly meant to demonstrate two things. Firstly is whether the regex engine does some of the most basic forms of optimization by recognizing that a pattern is just a literal, and that a full blown regex engine is probably not needed.

Much (not all) of the discussion here regards ways that the regex engine can "cheat" and especially whether it can use a literal search technique. I have wanted to do a phd around this topic, but one critical point I believe is that automata cannot be decomposed or easily used as components of other search mechanisms (which can be formalized), while literal matchers can be invoked more easily as a subroutine.

---------


That was just to say I'm very glad to see this crate adopt this approach for parsing, and while this PR seems expertly well-scoped, I believe it would be intriguing to consider the approach for other parsing problems.

However, from the code I can also see your concerns are much more about ensuring you have the right bytes in the right places than my experience. So I can see this is potentially misplaced.

I do however think path strings are the worst part of the rust stdlib but it seems your use cases for them are extremely straightforward which is also good.

Please feel free to ignore this. Good change, great project!

PR #745, opened 2026-03-25
Improvements for draft-forwarding (implementation and docs)

hko-s:
This PR improves zeroization and documentation for the draft-forwarding implementation (as introduced in https://github.com/rpgp/rpgp/pull/723 and https://github.com/rpgp/rpgp/pull/637)

### DIFF
--- a/src/composed/message/types.rs
+++ b/src/composed/message/types.rs
@@ -18,8 +18,8 @@ use crate::{
     parsing_reader::BufReadParsing,
     ser::Serialize,
     types::{
-        DecryptionKey, EskType, KeyDetails, Password, PkeskVersion, SecretParams, SkeskVersion,
-        Tag, VerifyingKey,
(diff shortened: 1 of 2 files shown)


hko-s:
Note to self: as pointed out by ROS, it would be good to fully feature-gate the draft-forwarding related business logic. 

Some (relatively trivial, but still draft-forwarding-specific) logic is not currently gated in:

https://github.com/rpgp/rpgp/blob/47a0ce2b548a429e43ad19fb263365bdd58c5c22/src/composed/message/types.rs#L1153

hko-s:
> Note to self: as pointed out by ROS, it would be good to fully feature-gate the draft-forwarding related business logic.
> 
> Some (relatively trivial, but still draft-forwarding-specific) logic is not currently gated in:
> 
> https://github.com/rpgp/rpgp/blob/47a0ce2b548a429e43ad19fb263365bdd58c5c22/src/composed/message/types.rs#L1153

This PR now replaces the try_key! macro with plain functions (and cleans up the feature gating, as part of that change)

PR #746, opened 2026-03-25
signer: make the parameter explicitely a prehash

baloo:
This is an attempt at making the usage of the slice explicit.

In a followup commit, I'd love to try to adjust the signer APIs to allow for the raw (unhashed) slice to be passed instead. When a private key is held externally (HSM, TPM, ...) the interface may not allow to sign prehashed byte strings.

This is an API breaking change.

### DIFF
--- a/src/adapter/ecdsa.rs
+++ b/src/adapter/ecdsa.rs
@@ -131,7 +131,7 @@ where
     T: PrehashSigner<ecdsa::Signature<C>>,
     C::Digest: KnownDigest,
 {
-    fn sign(
+    fn sign_prehash(
         &self,
         _key_pw: &Password,
         hash: HashAlgorithm,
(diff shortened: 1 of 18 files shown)


baloo:
I'm specifically aiming at removing the [`PrehashSigner` requirement from the `RsaSigner`](https://github.com/rpgp/rpgp/blob/47a0ce2b548a429e43ad19fb263365bdd58c5c22/src/adapter/rsa.rs#L75) which may not be implemented on every HSMs.

This is especially problematic for signing RPM which, up to very recently, only supported RSA keys (and will continue to be required if you're targeting any rpm-based distribution that may lag behind).

dignifiedquire:
why rename the existing method, instead of adding something like `sign_hashed` and work with that?

PR #747, opened 2026-03-31
WIP: add some convenience functionality to Timestamp and Duration

hko-s:
rPGP's `Timestamp` and `Duration` APIs are currently very minimal.

This PR collects some ergonomics improvements, based on concrete needs in various software that uses rPGP.

### DIFF
--- a/src/types/duration.rs
+++ b/src/types/duration.rs
@@ -10,6 +10,8 @@ use crate::ser::Serialize;
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Default)]
 pub struct Duration(u32);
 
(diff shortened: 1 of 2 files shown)


hko-s:
TODO: I tend to think we should make all potential u32 overflows into explicit `Error`s.

hko-s:
Notes to self:

- time math could be saturating (avoids overflowing, but doesn't need fallible APIs)
- some functions could take `t: impl Into<SystemTime>` for ergonomic interaction with chrono

hko-s:
The fact that `Add` can be called for a `Duration` with the value "0" (which is defined as an "infinite" duration) is a bit of a footgun.

Maybe the result should be an `Option<Timestamp>` that is `None` when adding a `Duration` that is 0?

PR #748, opened 2026-04-01
test: don't run a test by default that requires "tests/tests"

hko-s:


ros-cr:
Context: I originally ran into this while testing, and reported it to @hko-s. Thanks for fixing.

PR #760, opened 2026-04-04
fix: don't panic in legacy_key_id() for v2/3 keys with very short RSA modulus

hko-s:
When dealing with v2/v3 keys with unreasonably short RSA modulus (shorter than 8 bytes), the `legacy_key_id` function did not correctly check for an integer underflow. In release builds this caused a panic when trying to take a subslice.

Issue report and reproducer test by [Christian Reitter](https://github.com/invd) from [Radically Open Security](https://www.radicallyopensecurity.com/)

### DIFF
--- a/src/packet/key/public.rs
+++ b/src/packet/key/public.rs
@@ -807,17 +807,15 @@ impl KeyDetails for PubKeyInner {
             KeyVersion::V2 | KeyVersion::V3 => match &self.public_params {
                 PublicParams::RSA(params) => {
                     let n: Mpi = params.key.n().into();
-                    let raw: [u8; 8] = if n.len() >= 8 {
+                    let mut raw = [0u8; 8];
+                    if n.len() >= 8 {
                         let offset = n.len() - 8;
-                        n.as_ref()[offset..].try_into().expect("fixed size")
+                        raw.copy_from_slice(&n.as_ref()[offset..]);
                     } else {
-                        let padding = 8 - n.len();
-
-                        let mut padded = vec![0; padding];
-                        padded.extend_from_slice(n.as_ref());
-
-                        padded.try_into().expect("fixed size")
-                    };
+                        // unreasonably short modulus, left-pad with zeros
+                        let offset = 8 - n.len();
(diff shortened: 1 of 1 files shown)


src/packet/key/public.rs
```diff
@@ -807,8 +807,20 @@ impl KeyDetails for PubKeyInner {
             KeyVersion::V2 | KeyVersion::V3 => match &self.public_params {
                 PublicParams::RSA(params) => {
                     let n: Mpi = params.key.n().into();
-                    let offset = n.len() - 8;
-                    let raw: [u8; 8] = n.as_ref()[offset..].try_into().expect("fixed size");
+                    let raw: [u8; 8] = match n.len() {
```
dignifiedquire: a match is a bit odd here, I think `if n.len() < 8` would read a lot easier

hko-s: Right, I changed match to if.

(But left the "happy path" first, so it's now `if n.len() >= 8 {` ...)

src/packet/key/public.rs
```diff
@@ -807,8 +807,17 @@ impl KeyDetails for PubKeyInner {
...
+                        let offset = n.len() - 8;
+                        n.as_ref()[offset..].try_into().expect("fixed size")
+                    } else {
+                        let padding = 8 - n.len();
+
+                        let mut padded = vec![0; padding];
+                        padded.extend_from_slice(n.as_ref());
+
+                        padded.try_into().expect("fixed size")
+                    };
```
dignifiedquire: something like this is a bit cleaner, and avoids panics and the allocation
```rust
let raw =  [0u8; 8];
    if n.len() >= 8 {
        let offset = n.len() - 8;
        raw.copy_from_slice(n.as_ref()[offset..]);
    } else {
        let offset = 8 - n.len();

        raw[offset..].copy_from_slice(n.as_ref());
    }
```

hko-s: Good point. Changed it in this way.

dignifiedquire (approved):

PR #764, opened 2026-04-17
fix: subpacket length handling

hko-s:
This fixes two related issues in rPGP's subpacket length handling:

- When parsing a Signature packet, the length in the header of a subpacket is not rejected if it is too long. As long as the subpacket content can be parsed correctly, and doesn't exceed the length from the header, inconsistencies are ignored
- When serializing a Signature packet, the length field is not checked for consistency with the actual payload (this allowed roundtripping of erroneous Subpackets) 

Issue report and reproducer test by [Christian Reitter](https://github.com/invd) from [Radically Open Security](https://www.radicallyopensecurity.com/).


tests/security.rs
```diff
@@ -369,3 +369,54 @@ fn signed_public_key_legacy_key_id_crash() {
...
+            assert!(e.to_string().contains("Inconsistent subpacket length"));
+        }
+    }
+}
+
+#[test]
+fn signed_public_key_subkey_write_len_expect() {
+    use pgp::ser::Serialize;
+
+    // TODO
```
dignifiedquire: ?

tests/signature_test.rs
```diff
@@ -1,6 +1,3 @@
-#[macro_use]
```
dignifiedquire, line 1: ?

hko-s: The asserts are not there anymore, because the signature now doesn't parse anymore

tests/security.rs
hko-s: good question. that's definitely not serving any useful purpose - i'm removing the comment.

tests/signature_test.rs
hko-s: This test has changed (as a side effect of the business logic change in this PR) from

"parsing a weird signature (and checking the content of the parsed object)"

to

"the parser rejects this signature"

so there's no more object to (pretty)assert on.

dignifiedquire (approved):

PR #767, opened 2026-04-21
Explicitly handle edge case in dearmoring code

hko-s:
This PR adds a (fuzzer-generated) unit test that caused the dearmoring code to panic in debug builds.
It also adds a check for this situation, and returns an orderly error instead of a panic.

(The underlying logic error should never be a problem in release builds, so this PR serves mostly as cleanup.)

Issue report and reproducer test by [Christian Reitter](https://github.com/invd) from [Radically Open Security](https://www.radicallyopensecurity.com/).


dignifiedquire (approved):

ros-cr:
During a brief retest of the patch, the integer overflow didn't show up again, which is a positive signal.

A notable detail of how I found this originally: the problematic input was greater than 8192 bytes, even after minimization. Depending on the existing contents of the fuzzer corpus directory, overriding the maximum allowed input length for the fuzzer via a setting like `-max_len=10000` was therefore necessary to make this issue reachable + observable.
Discovering the issue also requires integer overflow checks to be enabled, which is not the default in all cargo build modes, but that is currently taken care of via `overflow-checks = true` in `cargo.toml` and less specific to this issue.

PR #768, opened 2026-04-22
chore: add rustfmt config that matches CI

hko-s:
It seems CI is enforcing this style of formatting, but we don't currently have a config in the project that helps to automatically produce it?


dignifiedquire:
it's stored in here: https://github.com/rpgp/rpgp/blob/main/Makefile.toml

hko-s:
Ah, I see. I was not aware of that mechanism.

PR #770, opened 2026-04-30
ci: add dependency cooldown for dependabot and setup zizmor

link2xt:
[Zizmor](https://github.com/zizmorcore/zizmor) is a security linter for github actions. Probably most important change is `permissions: {}` which makes sure CI cannot do many things they don't need to like creating releases. If we need to do it, permissions should be explicitly set for the jobs that are meant to create releases.

Also added dependency cooldown for dependabot just because zizmor recommends this. Probably does not hurt, in case some bad dependency update is released and revoked two days later because it is bad.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,5 @@ updates:
   schedule:
     interval: monthly
   open-pull-requests-limit: 10
+  cooldown:
+    default-days: 7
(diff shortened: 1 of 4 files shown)


link2xt:
As for "archived repository" warnings, we can ignore them, they will show up in "security" tab. Then can be dismissed, or we do something about them. It will not show up on PRs that don't change related lines.

PR #784, opened 2026-05-24
fix: Timestamp::now() panics in WASM

### LINKED ISSUE #783: Timestamp::now() panics in WASM
Hi everyone,

I think I found a bug when creating a custom subkey and secret key in WASM. Since the `chrono` was replaced with a custom type in 2435971, the `std::SystemTime` in `now` code panics when trying to create a new key-pair:

```
panicked at /rustc/4a4ef493e3a1488c6e321570238084b38948f6db/library/std/src/sys/pal/wasm/../unsupported/time.rs:35:9:
time not implemented on this platform
```

Even providing a custom Timestamp via `created_at` in the builder does not prevent the panic.

I'm not 100% familiar with the code and still new to Rust, but I would guess that in the "new" Timestamp type the now function could be replaced with something like this:

```rs
pub fn now() -> Self {
    #[cfg(feature = "wasm")]
    Self((js_sys::Date::now() / 1000.0) as u32)

    #[cfg(not(feature = "wasm"))]
    SystemTime::now()
            .try_into()
            .expect("now is too far into the future")
}
```

Either `web-time` or `js-sys` could be used as a replacement when the `wasm` feature is enabled.

I look forward to your feedback and would be happy to provide a PR if you want.

Tim-Otte:
I added `web-time` as a dependency and used it as a drop-in replacement for `std::time` which is incompatible with WASM.

Fixes #783

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1468,6 +1468,7 @@ dependencies = [
  "tempfile",
  "testresult",
  "twofish",
+ "web-time",
  "x25519-dalek",
  "zeroize",
 ]
@@ -2318,6 +2319,16 @@ name = "web-sys"
 version = "0.3.77"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "33b6dd2ef9186f1f2072e409e99cd22a975331a6b3591b12c764e0e55c60d5d2"
+dependencies = [
+ "js-sys",
+ "wasm-bindgen",
+]
+
+[[package]]
+name = "web-time"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5a6580f308b1fad9207618087a65c04e7a10bc77e02c8e84e9b00dd4b12fa0bb"
 dependencies = [
  "js-sys",
  "wasm-bindgen",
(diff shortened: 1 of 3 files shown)


hko-s:
This looks good to me, thank you!

CI is showing one complaint by rustfmt about ordering of the imports, so that will need adjusting (even if the change is probably not a clear improvement for human readers).

Tim-Otte:
Hi @hko-s, I'm happy to help.

I did not really understand, what the action was complaining about, but finally figured how to run the check manually/locally. The commit above should fix the formatting, even if I don't really like the proposed formatting :laughing:

hko-s:
> I did not really understand, what the action was complaining about, but finally figured how to run the check manually/locally. The commit above should fix the formatting, even if I don't really like the proposed formatting 😆

Thanks! It also took me a minute to understand what the diff in CI was saying. And then I rolled my eyes a bit :smile:

rPGP prioritizes consistency and automatic formatting over the last iota of niceness that manual formatting might in some places be able to bring. (This PR is a case where rustfmt is making things a little less nice - I also preferred your manual formatting.)

dignifiedquire (approved):
thanks

PR #797, opened 2026-06-06
docs: document what ADSK in key flags means

link2xt:


link2xt:
Generally more documentation comments is good to have, but for this one finding the reference is more difficult because it is not standard.

hko-s:
The text should probably clarify (even more) that the flag is non-standard.

Something like this, maybe?

> Sets the non-standard "Additional Decryption SubKey" flag.
>
> Proprietary feature introduced in GnuPG 2.4.1: <https://www.gnupg.org/blog/20230321-adsk.html>

hko-s (approved):
Looks great, thanks!

PR #800, opened 2026-06-08
Check signature types in verify_* methods

### LINKED ISSUE #799: `verify_bindings` does not check signature types
`verify_bindings` e.g. here checks that there is at least one signature and checks all of them:
https://github.com/rpgp/rpgp/blob/2df698c8c1d3f0f9714000cd503998dadc3aa7ab/src/composed/signed_key/secret.rs#L245-L256

However, the signature type is not checked. I tried to add signature type checking, e.g. check in `verify_subkey_binding` that signature is indeed a subkey binding signature as the documentation says:
https://github.com/rpgp/rpgp/blob/2df698c8c1d3f0f9714000cd503998dadc3aa7ab/src/packet/signature/types.rs#L631-L632

One of the tests failed because the signature checked turned out to be a revocation signature.

Will open a PR adding the checks and we can decide what to do with it.

link2xt:
Closes #799

### DIFF
--- a/src/packet/signature/types.rs
+++ b/src/packet/signature/types.rs
@@ -527,7 +527,14 @@ impl Signature {
         key.verify(config.hash_alg, hash, signature)
     }
 
-    /// Verifies a certification signature type (for self-signatures).
+    /// Verifies a certification signature or certification revocation signature (for self-signatures).
+    ///
+    /// Allowed signature types are:
+    /// - Generic Certification Signature (Type ID 0x10)
+    /// - Persona Certification Signature (Type ID 0x11)
+    /// - Casual Certification Signature (Type ID 0x12)
+    /// - Positive Certification Signature (Type ID 0x13)
+    /// - Certification Revocation Signature (Type ID 0x30)
     pub fn verify_certification<V>(&self, key: &V, tag: Tag, id: &impl Serialize) -> Result<()>
     where
         V: VerifyingKey + Serialize,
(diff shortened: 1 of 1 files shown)


src/packet/signature/types.rs
```diff
@@ -558,6 +558,18 @@ impl Signature {
...
         debug!("verifying certification {key_id:?} {self:#?}");
 
+        ensure!(
+            matches!(
+                config.typ,
+                SignatureType::CertGeneric
+                    | SignatureType::CertPersona
+                    | SignatureType::CertCasual
+                    | SignatureType::CertPositive
+                    | SignatureType::ThirdParty
```
link2xt: The function here is called `verify_third_party_certification`, but it is called from `verify_certification` as well which is for checking self-signatures. A lot of tests fail if I don't allow 0x10..0x13 signature types as well.

src/packet/signature/types.rs
```diff
@@ -645,6 +657,12 @@ impl Signature {
             unsupported_err!("signature version {:?}", self.version());
         };
 
+        ensure_eq!(
+            config.typ,
+            SignatureType::SubkeyBinding,
```
link2xt: This check broke `test_parse_gnupg_v1` where one of the signatures is a revocation signature.

link2xt: Test key `tests/openpgp-interop/testcases/keys/gnupg-v1-004.asc` has a `SubkeyRevocation` signature.

link2xt: I have allowed revocation signatures here and updated the documentation.

src/packet/signature/types.rs
hko-s: I think we should allow all of 0x10..0x13 and certification revocation signatures here.

This validation method should be usable to check both "positive" and "negative" (revocation-)  statements about certifications.

hko-s: `ThirdParty` should not be allowed in this function.

(The signature type is generally not used, as far as I know, but either way, it needs a different hash digest than the one this function calculates)

hko-s: (Fwiw: Third party certifications on identities use the exact same set of signature types as first party certifications. There are some conventions for which of the 0x10..0x13 ids are "typically" used for which case, but those are fuzzy and have not been uniformly applied.
0x11 is apparently widely considered non-meaningful [and as far as I know it's just ignored by GnuPG]. But I think we should be unopinionated about that, in this function.)

hko-s (approved):
One nit (`SignatureType::ThirdParty` should be removed), otherwise looks good to me

src/packet/signature/types.rs
link2xt: Removed ThirdParty and updated the documentation.

hko-s:
Looks good to me, I think this is strictly better than before.

PR #805, opened 2026-06-19
chore: disable default `num_enum` features

CathalMullan:
Was slightly surprised when looking at my dep graph to see a TOML parser being pulled in by `pgp`. 

Turns out `num_enum` (via `proc-macro-crate`) has an option to read your `Cargo.toml` to track package renames. So since `num_enum` isn't renamed here, it's not needed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -789,12 +789,6 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


hko-s (approved):
Good catch, thanks!

PR #808, opened 2026-06-25
docs: More documentation for Message

hko-s:
Includes guidance to avoid implicit large allocations.

Thanks to @lintowe for pointing out that the previous documentation for `Message::as_data_vec` and `Message::as_data_string` did not make it sufficiently clear that they involve a risk of resource exhaustion.

### DIFF
--- a/src/composed/message/types.rs
+++ b/src/composed/message/types.rs
@@ -153,8 +153,7 @@ impl BufRead for MessageReader<'_> {
 ///
 /// The core of any OpenPGP message is a plaintext payload, encoded as a "Literal Message".
 ///
-/// A Literal Message may be encrypted, cryptographically
-/// signed, and/or compressed.
+/// A Literal Message may be encrypted, cryptographically signed, and/or compressed.
 ///
 /// The OpenPGP message grammar allows arbitrary (and recursive) wrapping of the central
 /// Literal Message in a series of:
--- a/src/composed/message/types.rs
+++ b/src/composed/message/types.rs
@@ -183,7 +183,14 @@ impl BufRead for MessageReader<'_> {
 /// Callers should consume `Message` as a streaming reader, wherever possible. Depending on the
 /// use case, wrapping `Message` with a limiting reader
 /// (e.g. via [`Read::take`](https://doc.rust-lang.org/std/io/trait.Read.html#method.take)
-/// is a good strategy.
+/// is a good strategy.)
+///
+/// Note: If the message still contains unread data after limited reading (as tested e.g. via
+/// `Message::has_buffer_available`), the caller must consider the operation unsuccessful, and
+/// proceed accordingly - e.g. by throwing an error.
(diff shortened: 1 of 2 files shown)


src/composed/message/types.rs
```diff
@@ -983,13 +1018,33 @@ impl<'a> Message<'a> {
...
+    ///
+    /// NOTE: The output of this call may be arbitrarily large, and allocate a potentially large
+    /// amount of memory!
+    /// This convenience function is not appropriate for use in environments where "out of memory"
+    /// crashes are unacceptable!
+    ///
+    /// Callers should consider using a limited reader (e.g. via `Read::take`) to safely consume
+    /// message payloads via the `Read` implementation of this type.
+    ///
+    /// See the top level [`Message`] documentation for more information!
```
link2xt: This kind of references easily gets outdated, someone will later decide to shorten `Message` documentation or put something else there and this will become outdated and hard to figure out what this is referring to.

src/composed/message/types.rs
```diff
@@ -983,13 +1018,33 @@ impl<'a> Message<'a> {
...
```
link2xt: Same as my other comment above.

src/composed/message/types.rs
hko-s: That's a fair concern. I've duplicated the top level text to both functions. Is that what you had in mind?

I wanted to keep duplication limited, but agree that that is brittle in the longer run.

link2xt:
Running `cargo doc` warns that `Message::has_buffer_available` referenced in the docs is private.

src/composed/message/types.rs
```diff
@@ -151,16 +151,51 @@ impl BufRead for MessageReader<'_> {
...
+///
+/// In particular, consumers who need to defend against resource exhaustion must avoid allocating
+/// more memory than is available in their context!
 ///
-/// OpenPGP message may consist of multiple layers wrapped around each other (such as for example:
-/// an encrypted message, containing a compressed message, containing a one-pass signed message,
-/// containing a literal message).
+/// Defensive users should consume `Message` as a streaming reader, wherever possible.
+/// Depending on the use case, wrapping `Message` with a limiting reader
+/// (e.g. via [`Read::take`](https://doc.rust-lang.org/std/io/trait.Read.html#method.take)
```
link2xt: This can be just
```
(e.g. via [`std::io::Read::take`] is a good strategy.)
```

This way it points to the documentation corresponding to the locally used rust automatically and not hardcoded URL.

hko-s: Good point, thanks

hko-s:
> Running `cargo doc` warns that `Message::has_buffer_available` referenced in the docs is private.

Ouch, thanks for catching that! I adjusted the text to no rely on private functions.

link2xt (approved):

PR #809, opened 2026-06-25
chore!: drop the unused MAX_BUFFER_SIZE const

hko-s:


hko-s:
I think we should merge this, now that main contains another semver break already

link2xt (approved):

link2xt:
There is no "cleanup" tag:
https://github.com/rpgp/rpgp/blob/82aae1f2eb84b19f5f76edc024ef2ca62cb12283/cliff.toml#L47-L64
Should probably be `chore` or `refactor`.

PR #812, opened 2026-06-29
try crypto::dsa::tests::test_dsa_2048 500 times

link2xt:
attempt to reproduce flaky test on macos arm64

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,13 +43,16 @@ jobs:
     - name: Install cargo-nextest
       uses: taiki-e/install-action@v2
       with:
-        tool: nextest@0.9.80
+        tool: nextest@0.9.137
 
     - name: tests
       run: cargo nextest run --lib --bins --tests --all
 
     - name: doc tests
       run: cargo test --doc
+
+    - name: test crypto::dsa::tests::test_dsa_2048 500 times
+      run: cargo nextest run --stress-count 500 crypto::dsa::tests::test_dsa_2048
 
     - name: asm
       if: ${{ matrix.rust == env.RUST_NIGHTLY }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
       run: cargo test --doc
 
     - name: test crypto::dsa::tests::test_dsa_2048 500 times
-      run: cargo nextest run --stress-count 500 crypto::dsa::tests::test_dsa_2048
+      run: cargo nextest run --lib --bins --tests --all --features asm --stress-count 500 crypto::dsa::tests::test_dsa_2048
 
     - name: asm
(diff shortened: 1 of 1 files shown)


baloo:
I think I have an isolated reproducer on macos/arm64.
This appears to be a bug in sha1 0.10 asm on macos.

baloo:
https://github.com/baloo/sha1-reproducer
https://github.com/RustCrypto/hashes/issues/891

hko-s:
> I think I have an isolated reproducer on macos/arm64. This appears to be a bug in sha1 0.10 asm on macos.

Awesome! Does that reproducer have any non-deterministic behavior?

(I'm frankly a bit confused about what different effects our CI runs have exhibited, and wonder if this find explains *all* of the recent CI outcomes.)

baloo:
~~I'm not sure. I'm stressing the code path with a bug, I haven't checked the value of the iterator. I don't know if that occurs on the first hit or not.~~

~~I have to borrow my partner's laptop to try.~~

EDIT: No, that was a previous version of the reproducer. This is fully reproducable as far as I can tell. Or there is another issue somewhere else.

hko-s:
> EDIT: No, that was a previous version of the reproducer. This is fully reproducable as far as I can tell. Or there is another issue somewhere else.

Alright, thanks! I'm starting to think that some of the seeming non-determinism in our CI runs might just have been caused by the (random) execution order of tests, which would surface problems in different test cases first.

We also saw failures in `types::s2k::tests::argon2_skesk_msg` which decrypts "SEIPDv1" messages (after an Argon2 step, which I assume works correctly). But SEIPDv1 encryption internally uses SHA1-based checksums. And indeed, the error thrown by the test is `MdcError` - which would be explained perfectly by a broken SHA1 calculation.

baloo:
I think [newpavlov's comment](https://github.com/RustCrypto/hashes/issues/891#issuecomment-4836611198) and this [comment](https://github.com/RustCrypto/hashes/blob/sha1-v0.10.6/sha1/Cargo.toml#L30) suggests rpgp should never enable `sha1/asm` feature.

Leave that to the downstream consumer of rpgp to decide whether they want the broken optimization or not.

hko-s:
> I think [newpavlov's comment](https://github.com/RustCrypto/hashes/issues/891#issuecomment-4836611198) and this [comment](https://github.com/RustCrypto/hashes/blob/sha1-v0.10.6/sha1/Cargo.toml#L30) suggests rpgp should never enable `sha1/asm` feature.
> 
> Leave that to the downstream consumer of rpgp to decide whether they want the broken optimization or not.

I made https://github.com/rpgp/rpgp/pull/814 to consider dropping the `asm` feature entirely from rPGP.

It doesn't feel entirely ideal to rely on downstream users to know if/which optimizations they want to enable, though.

(This change makes rPGP slower for SHA1 on Intel, I assume? And probably slower for SHA2 as well?!)

hko-s:
Resolved by #814

PR #814, opened 2026-06-30
chore: drop the asm feature for sha1

hko-s:
Rustcrypto `asm` should be opted into by downstream users.

See https://github.com/rpgp/rpgp/pull/812 for context

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -77,7 +77,6 @@ rand = "0.8.6"
 ripemd = { version = "0.1.3", features = ["oid"] }
 rsa = { version = "0.9.10" }
 sha1 = { version = "0.10.6", features = ["oid"] }
-sha1-asm = { version = "0.5.3", optional = true }
 sha1-checked = { version = "0.10", features = ["zeroize"] }
 sha2 = { version = "0.10.6", features = ["oid"] }
 sha3 = { version = "0.10.8", features = ["oid"] }
@@ -139,7 +138,6 @@ default = ["bzip2"]
 
 # See README.md for features documentation
 bzip2 = ["dep:bzip2"]
-asm = ["dep:sha1-asm", "sha1/asm", "sha2/asm", "md-5/asm"]
 wasm = ["getrandom", "getrandom/js", "web-time"]
 
 large-rsa = []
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -139,7 +138,6 @@ default = ["bzip2"]
 
 # See README.md for features documentation
 bzip2 = ["dep:bzip2"]
-asm = ["dep:sha1-asm", "sha1/asm", "sha2/asm", "md-5/asm"]
```
baloo, line 142: My suggestion was only to remove “sha1/asm” from that list.

The direct dependency on sha1-asm can definitely be dropped though.

In any case, this is just dropping an optimization, nothing a downstream can’t re-enable.

Maybe replace that with an __internal_asm flags? To keep exercising those code-paths in CI and root out bugs?

baloo: Rpgp exercices a ton of the api surface exposed by rustcrypto. I don’t want to speak for the rustcrypto project, but I believe this provides a ton of value for rustcrypto as a “ci”/validation, especially because it ties multiple crates together in different combinations.

hko-s: Fair points, thank you!

I've reinstated the sha2/md5 use of asm (though I wonder if there is any serious justification for enabling asm for md5)

And it sounds reasonable to me to enable asm for sha1 as an extra "internal" bit of CI, as an extra bit of Rust Crypto testing. I wonder if we'd want to re-enable it for x86 only, in this context, or what would be a good approach here?

baloo: In any case, the `asm` backend is getting dropped in the next release.

baloo (approved):

Cargo.toml
baloo: Ideally, we'd get a sha1 0.10.7 that disables asm on aarch64. But I don't know if there any will to do that upstream.

hko-s: > Ideally, we'd get a sha1 0.10.7 that disables asm on aarch64. But I don't know if there any will to do that upstream.

If that's only a fix on an effectively "EOL" branch of the crate, then it's probably not worth complicating rPGP's CI over it, though, right?

(Assuming anyone makes such a release at all)

dignifiedquire (approved):

PR #825, opened 2026-07-02
feat!: update pqc references to the now published RFC 9980

hko-s:
and rename feature from `draft-pqc` to `pqc`

PR #827, opened 2026-07-16
fix!: don't panic on short x25519 session key

hko-s:
This fixes a regression introduced in 4db0b6d8a15171de9b5cd8df36aa66cb09f3c908

Found and reported by @hibrian827 - thanks!


link2xt:
In https://github.com/chatmail/core/ we have `#![cfg_attr(not(test), warn(clippy::arithmetic_side_effects))]` to catch this kind of bugs. See https://rust-lang.github.io/rust-clippy/master/index.html?search=arithm#arithmetic_side_effects for the description. We still have some functions marked with `#[expect(clippy::arithmetic_side_effects)]`, but this prevents introducing more bugs.

hko-s:
> In https://github.com/chatmail/core/ we have `#![cfg_attr(not(test), warn(clippy::arithmetic_side_effects))]` to catch this kind of bugs. See https://rust-lang.github.io/rust-clippy/master/index.html?search=arithm#arithmetic_side_effects for the description. We still have some functions marked with `#[expect(clippy::arithmetic_side_effects)]`, but this prevents introducing more bugs.

It's going to require some work to introduce that in rpgp, but I agree it sounds worth doing.

dignifiedquire (approved):

src/crypto/aes_kw.rs
```diff
@@ -12,6 +12,8 @@ const IV_LEN: usize = 8;
 pub enum Error {
     #[snafu(display("invalid key size: {}", size))]
     InvalidKeySize { size: usize },
+    #[snafu(display("invalid data size: {}", size))]
+    InvalidDataSize { size: usize },
```
dignifiedquire, line 16: this is a breaking change, the enum is non exhaustive

hko-s: Right, so if we want to make a point release, that branch should not add this new error type.
I could split this commit in two, with the first part reusing an existing error type, so that we can cherry pick that into a point release?

hko-s:
We already have another semver breaking change in main at this point (renaming the PQC feature, via #825).

I think we should just merge this as is?

PR #854, opened 2026-09-18
Optionally parse certs with non-canonical MPIs in embedded signatures

lubux:
Behind the `malformed-artifact-compat` feature flag keep an embedded signature that does not re-serialize to its declared length as an opaque subpacket instead of rejecting the whole certificate. 
The affected subkey binding won't verify.

This allows to import a key that contain a subkey with a malformed embedded signature.
We observed several such keys in the wild as presented in the test vector.


hko-s (approved):
This looks like a sensible - and now nicely contained - change to me (even if it is to handle unreasonable data).

